Issue #36, opened 2015-01-22
Regression: clang 3.5 causes struct members to be defined again


joakim-noah:
I recently ran dstep from Arch and found that the output is different.  Since the dstep package hasn't been updated, I'm guessing the update from clang 3.4 to 3.5 caused this regression.  Here's a test header file I came up with to demonstrate the problem:

```C
struct foo { int x; int y;};

struct bar { struct foo a; struct foo b;};
```

And the D output dstep produces:

```D
extern (C):

struct foo
{
        int x;
        int y;
}

struct bar
{
        struct foo
        {
                int x;
                int y;
        }
        foo a;
        struct foo
        {
                int x;
                int y;
        }
        foo b;
}
```

There's no reason to define the struct again twice, not sure why it's doing that.

mihails-strasuns-sociomantic:
Don't have old clang to see how it worked before but right now the problem is how libclang reports cursors:

```D
struct foo { }; // CXCursor_StructDecl

struct foo var; // CXCursor_VarDecl

struct bar // CXCursor_StructDecl
{
    struct foo var; // CXCursor_StructDecl
}
```

I am looking for a way to distinguish that latter `CXCursor_StructDecl` is not really a struct declaration based on cursor metadata but nothing obvious seems to match.

mihails-strasuns-sociomantic:
One option cold be to store list of all definition cursors (http://clang.llvm.org/doxygen/group__CINDEX__CURSOR__XREF.html#gafcfbec461e561bf13f1e8540bbbd655b) and check it before outputing a new one.

mihails-strasuns-sociomantic:
This may be related, global struct variables are also generated wrong:
```C
struct zzz { };

extern struct zzz x;
```
into
```D
extern (C):

extern __gshared zzz x;

struct zzz
{
}
```

mihails-strasuns-sociomantic:
#37

jacob-carlborg:
@mihails-strasuns-sociomantic `struct foo var` is not reported as a `CXCursor_FieldDecl` when inside a struct?

jacob-carlborg:
> Don't have old clang to see how it worked before but right now

If you want to, you can run `dub --config=test` which will download a couple of versions of libclang and run the whole test suite. The actual test suite requires Ruby and two gems.

mihails-strasuns-sociomantic:
> struct foo var is not reported as a CXCursor_FieldDecl when inside a struct?

[...]

PR #37, opened 2015-02-13
Proof of concept fix for #36

mihails-strasuns-sociomantic:
Not checked with a test suite and results in extra empty
new lines placed instead but showcases core idea. @jacob-carlborg would you suggest any better place where to put it in dstep sources?

### DIFF
--- a/clang/Cursor.d
+++ b/clang/Cursor.d
@@ -104,6 +104,11 @@ struct Cursor
     {
         return clang_isCursorDefinition(cast(CXCursor) cx) != 0;
     }
+
+    @property Cursor definition () const
+    {
+        return Cursor(clang_getCursorDefinition(cast(CXCursor) cx));
+    }
 }
 
 struct ObjcCursor
--- a/dstep/translator/Record.d
+++ b/dstep/translator/Record.d
@@ -20,6 +20,8 @@ import dstep.translator.Type;
 
 class Record (Data) : Declaration
 {
+    static bool[Cursor] recordDefinitions;
+
     this (Cursor cursor, Cursor parent, Translator translator)
     {
         super(cursor, parent, translator);
@@ -36,10 +38,16 @@ class Record (Data) : Declaration
                         case CXCursor_FieldDecl:
                             if (!cursor.type.isExposed && cursor.type.declaration.isValid)
                             {
-                                output.newContext();
-                                output.currentContext.indent in {
-                                    context.instanceVariables ~= translator.translate(cursor.type.declaration);
-                                };
+                                auto def = cursor.type.declaration.definition;
+                                auto known = def in this.recordDefinitions;
+
+                                if (!known)
+                                {
+                                    output.newContext();
+                                    output.currentContext.indent in {
+                                        context.instanceVariables ~= translator.translate(cursor.type.declaration);
+                                    };
+                                }
 
                                 if (cursor.type.declaration.type.isEnum || !cursor.type.isAnonymous)
                                     translateVariable(cursor, context);
@@ -60,8 +68,12 @@ private:
     string writeRecord (string name, void delegate (Data context) dg)
     {
         auto context = new Data;
-        static if (is(typeof(context.isFwdDeclaration) == bool))
-            context.isFwdDeclaration = !cursor.isDefinition;
+        
+        if (cursor.isDefinition)
+            this.recordDefinitions[cursor] = true;
+        else
+            context.isFwdDeclaration = true;
+
         context.name = translateIdentifier(name);
 
         dg(context);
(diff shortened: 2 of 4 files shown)


jacob-carlborg:
I would like to take a look myself how the Clang AST looks like before I can determine if this is a good solution.

mihails-strasuns-sociomantic:
On related topic : is there any pretty printer of raw C AST representation in dstep sources? I didn't spot anything similar, could be useful for debugging.

jacob-carlborg:
No, there isn't.

jacob-carlborg:
But Clang itself has an AST dumper: `$ clang -Xclang -ast-dump <some file>`. Some kind of documentation [here](http://clang.llvm.org/docs/IntroductionToTheClangAST.html).

dstep/translator/Record.d
```diff
@@ -59,9 +61,17 @@ private:
...
+
         auto context = new Data;
-        static if (is(typeof(context.isFwdDeclaration) == bool))
-            context.isFwdDeclaration = !cursor.isDefinition;
+        
+        if (cursor.isDefinition)
+            this.record_definitions[cursor] = true;
+        else
+            context.isFwdDeclaration = true;
+
```
jacob-carlborg: I think it would be better to handle all this in `Record.translate`. It's [these lines](https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Record.d#L39-L42) that will write a struct declaration inline. When the definition is already know, don't execute those lines.

clang/Cursor.d
```diff
@@ -104,6 +104,11 @@ struct Cursor
     {
         return clang_isCursorDefinition(cast(CXCursor) cx) != 0;
     }
+
+    Cursor definition () const @property
```
jacob-carlborg: `@property` should be first to follow style in the file.

jacob-carlborg:
Could you please add a testcase.

dstep/translator/Record.d
mihails-strasuns-sociomantic: I am not sure I understand. Lines you have linked get called _from_ `writeRecord` and are responsible for writing the fields (`CXCursor_FieldDecl`). Struct declarations get written from the return value of `writeRecord` (`return context.data`).

jacob-carlborg: The lines I've linked above are the lines responsible for translating the actual type declaration. The [lines below](https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Record.d#L44-L45) are for the field declaration.

This is the flow of for translating a struct:

0. All top level nodes are iterated and translated, [here](https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Translator.d#L62-L68)
1. When a struct is seen, it will start translating that, [here](https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Translator.d#L140)
3. `writeRecord` is called which will translate the identifier of the struct, [here](https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Record.d#L60)
4. `writeRecord` then calls the delegate which will start iterating all the children nodes of the struct, [here](https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Record.d#L31)
5. When a field is seen it will check if the type is not exposed and has a declaration, [here](https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Record.d#L37)
6. If the type is not exposed and has a declaration it will translate the declaration of the type [here](https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Record.d#L41)
7. If the declaration is a struct it will start from the beginning at step 1

mihails-strasuns-sociomantic:
Updated.

dstep/translator/Record.d
mihails-strasuns-sociomantic: Ok, thanks, I have missed step 6 where it went in recursion and was confused by seemingly broken iteration over struct members.

mihails-strasuns-sociomantic:
Hm, I have just spotted that it reorders last two struct declaration in generated binding but can't spot obvious reason to do so, any ideas?

jacob-carlborg:
`F` is deferred: https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Translator.d#L93 https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Translator.d#L141-L151

dstep/translator/Record.d
```diff
@@ -20,6 +20,8 @@ import dstep.translator.Type;
 
 class Record (Data) : Declaration
 {
+    static bool[Cursor] record_definitions;
```
jacob-carlborg: Please use camel casing.

mihails-strasuns-sociomantic: Updated

jacob-carlborg:
Looking good.

jacob-carlborg:
Thanks.

PR #43, opened 2015-11-01
Create support for auto converting includes to imports.

Mitten-O:
Added command line options
--import-filter
A regular expression. All includes matching this will be converted to imports.
Only the basename part of the file name is used to generate the import.
--import-prefix
A prefix to give the imports. The import will have the form "<prefix><basename>"

This could be extended in the future to support filter-prefix pairs to give different prefixes to
includes from different paths, but this was all I needed for now.

### DIFF
--- a/dstep/driver/Application.d
+++ b/dstep/driver/Application.d
@@ -25,6 +25,7 @@ import clang.Util;
 
 import dstep.core.Exceptions;
 import dstep.translator.Translator;
+import dstep.translator.IncludeHandler;
 
 class Application : DStack.Application
 {
@@ -77,6 +78,14 @@ class Application : DStack.Application
             .on(&handleLanguage);
 
         arguments("objective-c", "Treat source input file as Objective-C input.");
+
+        arguments('f',"import-filter", "A regex to filter includes that will be auto converted.")
+            .params(1)
+            .defaults(".*");
+
+        arguments('p',"import-prefix", "A prefix to add to any custom generated import")
+            .params(1)
+            .defaults("");
     }
 
 private:
@@ -131,6 +140,12 @@ private:
         // FIXME: Cannot use type inference here, probably a bug. Results in segfault.
         if (arguments.rawArgs.any!((string e) => e == "-ObjC"))
             handleObjectiveC();
+
+        if (arguments["import-prefix"].hasValue)
+            handleAutoImportPrefix(arguments["import-prefix"].value);
+
+        if (arguments["import-filter"].hasValue)
+            handleAutoImportFilter(arguments["import-filter"].value);
     }
 
     void handleObjectiveC ()
@@ -166,6 +181,16 @@ private:
 
         argsToRestore ~= "-x";
         argsToRestore ~= language;
+    }
+
+    void handleAutoImportPrefix (string prefix)
+    {
+        includeHandler.autoImportPrefix = prefix;
+    }
+
+    void handleAutoImportFilter (string filter)
+    {
+        includeHandler.autoImportFilter = filter;
     }
 
     @property string[] remainingArgs ()
@@ -213,6 +238,8 @@ private:
         println("    -ObjC, --objective-c         Treat source input file as Objective-C input.");
         println("    -x, --language <language>    Treat subsequent input files as having type <language>.");
         println("    -h, --help                   Show this message and exit.");
+        println("    -f, --import-filter          A regex to filter includes that will be auto converted to imports.");
+        println("    -p, --import-prefix          A prefix to add to any import generated from an include.");
         println();
         println("All options that Clang accepts can be used as well.");
         println();
--- a/dstep/translator/IncludeHandler.d
+++ b/dstep/translator/IncludeHandler.d
@@ -7,8 +7,10 @@
 module dstep.translator.IncludeHandler;
 
 import Path = std.path;
+import std.regex;
+import std.conv;
 
-import mambo.core._;
+import mambo.core.Array;
 
 private IncludeHandler includeHandler_;
 
@@ -26,6 +28,16 @@ class IncludeHandler
 {
     private string[] rawIncludes;
     private string[] imports;
+
+    // True if includes should be converted to imports.
+    private bool convertIncludes = false;
+
+    // Includes matching this will be converted to imports.
+    private Regex!char convertableIncludePattern = regex(".*");
+
+    // Prefix for auto generated imports.
+    private string importPrefix = "";
+
     static string[string] knownIncludes;
 
     static this ()
@@ -109,12 +121,33 @@ class IncludeHandler
         imports ~= "core.stdc.config";
     }
 
+
+    /// Makes includes that match regex filter be converted to import with prefix.
+    @property void autoImportPrefix (string prefix)
+    {
+        convertIncludes = true;
+        importPrefix = prefix;
+    }
+
+    @property string autoImportPrefix ()
+    {
+        return this.importPrefix;
+    }
+
+    /// Makes includes that match regex filter be converted to import with prefix.
+    @property void autoImportFilter (string filter)
+    {
+        convertIncludes = true;
+        convertableIncludePattern = regex(filter);
+    }
+
     string[] toImports ()
     {
-        auto r =  rawIncludes.map!((e) {
+        auto r = rawIncludes.map!((e) {
             if (auto i = isKnownInclude(e))
                 return toImport(i);
-
+            else if (convertIncludes && isConvertableInclude(e) )
+                return toImport(autoConvertInclude(e));
             else
                 return "";
         });
@@ -122,6 +155,12 @@ class IncludeHandler
         auto imps = imports.map!(e => toImport(e));
 
         return r.append(imps).filter!(e => e.any).unique.toArray;
+    }
+
+    /// Returns the base name (last component without extension) of a file path.
+    static string bareName (string path)
+    {
+        return Path.stripExtension(Path.baseName(path));
     }
 
 private:
@@ -139,5 +178,27 @@ private:
             return r.value;
 
         return null;
+    }
+
+    /// Checks if the given include file name should be converted to an import declaration.
+    bool isConvertableInclude (string include)
+    {
+        // Do not try to convert empty strings, no matter what the pattern says.
+        return include != "" && matchFirst(include, convertableIncludePattern);
+    }
+
+
+    /// Generates an importable module name from an include file name.
+    string autoConvertInclude (string include)
+    {
+        string prefix = importPrefix;
+
+        // If we have a prefix, append a dot to it if it doesn't already have one at the end.
+        if (prefix != "" && !prefix.endsWith("."))
+        {
+            prefix ~= ".";
+        }
+
+        return prefix ~ bareName(include);
     }
 }
--- a/features/include_conversion.feature
+++ b/features/include_conversion.feature
@@ -1,0 +1,5 @@
+Feature: Include Conversion
+
+  Scenario: Convert includes
+    Then I test the file "includes" with filter "dummy_includes/transform.*" and prefix "transformed"
+
--- a/test_files/dummy_includes/README.md
+++ b/test_files/dummy_includes/README.md
@@ -1,0 +1,8 @@
+## Directory dummy_includes
+
+The support for import generation works based on the use of types
+haling from specific includes. That is why to test it
+we need a few dummy includes that define types.
+
+(Other test files cannot be used because they need
+to be free to contain name conflicts.)
--- a/README.markdown
+++ b/README.markdown
@@ -49,7 +49,7 @@ Use `-h` for usage information. Any flags recognized by Clang can be used.
 ## Limitations/Known issues
 
 * Doesn't translate preprocessor macros of any kind
-* Doesn't translate `#include` to `import`. A few standard C headers are translated
+* Only very simplistic translation of `#include` to `import`. A few standard C headers are translated
 * Doesn't translate C++ at all
 * Umbrella headers. Some headers just serve to include other headers. If these other headers contain some form of protection, like `#error`, to be included directly this can cause problems for DStep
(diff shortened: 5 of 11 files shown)


jacob-carlborg:
* The existing tests break
* There are no tests for this feature
* The code doesn't follow the existing style. The new code isn't even consistent in it self

Mitten-O:
Sorry, I had trouble noticing the tests (I even tried to look for some, though not very hard), being unfamiliar with cucumber.

I created one now for the change I made and fixed what broke.  Is one sufficient or should I maybe test the defaults or something like that?

I also tried to align my white space with your style. I'm a bit blind to it, though.

dstep/translator/IncludeHandler.d
```diff
@@ -109,21 +123,43 @@ class IncludeHandler
         imports ~= "core.stdc.config";
     }
 
+    /// Makes includes that match regex filter be converted to import with prefix.
+    void setAutoImportPrefix (string prefix)
```
jacob-carlborg: Please use properties instead of Java style setters.

dstep/translator/IncludeHandler.d
```diff
@@ -109,21 +123,43 @@ class IncludeHandler
...
+    }
+
     string[] toImports ()
     {
-        auto r =  rawIncludes.map!((e) {
+        auto r =  mambo.core.Array.map!((e) {
             if (auto i = isKnownInclude(e))
                 return toImport(i);
-
+            else if( this.convertIncludes && isConvertableInclude(e) )
```
jacob-carlborg: Please put a space before the opening parenthesis and no spaces after it.

dstep/translator/IncludeHandler.d
```diff
@@ -109,21 +123,43 @@ class IncludeHandler
...
+    void setAutoImportFilter (string filter)
+    {
+        this.convertIncludes = true;
+        this.convertableIncludePattern = regex(filter);
+    }
+
     string[] toImports ()
     {
-        auto r =  rawIncludes.map!((e) {
+        auto r =  mambo.core.Array.map!((e) {
```
jacob-carlborg: Why was this changed?

dstep/driver/Application.d
```diff
@@ -160,6 +175,16 @@ private:
         argsToRestore ~= language;
     }
 
+    void handleAutoImportPrefix (string prefix)
+    {
+        includeHandler.setAutoImportPrefix(prefix);
+    }
+
+    void handleAutoImportFilter(string filter)
```
jacob-carlborg: Please add a space after the method name.

dstep/translator/IncludeHandler.d
```diff
@@ -109,21 +123,43 @@ class IncludeHandler
...
 
         return r.append(imps).filter!(e => e.any).unique.toArray;
     }
 
+    /// Returns the base name (last component without extension) of a file path.
+    static string baseName (string path)
+    {
+        string last_component = text(retro(Path.pathSplitter(path)).front);
+        return Path.stripExtension( last_component );
+    }
```
jacob-carlborg: Phobos already contains `baseName` [1], would that work?

[1] http://dlang.org/phobos/std_path.html#.baseName

features/include_conversion.feature
```diff
@@ -0,0 +1,5 @@
+Feature: Include Conversion
+
+  Scenario: Convert includes
+    Then I test the file "includes" with filter "dummy_includes/transform.*" and prefix "transformed."
```
jacob-carlborg: Should the prefix include a dot? I would prefer that is handled automatically by dstep.

test_files/dummy_includes/README.md
```diff
@@ -0,0 +1,8 @@
+Directory dummy_includes
+========================
```
jacob-carlborg: Please use the one or multiple `#` instead of the `= `.

jacob-carlborg:
> I created one now for the change I made and fixed what broke. Is one sufficient or should I maybe test the defaults or something like that?

There is already a simple test for includes. Not sure if that's enough.

dub.json
```diff
@@ -19,7 +19,7 @@
             "targetName": "dstep",
             "sourcePaths": ["dstep", "clang"],
             "importPaths": ["dstep", "clang"],
-            "lflags-posix": ["-lclang", "-rpath", ".", "-L."],
+            "lflags-posix": ["-lclang","-lz", "-rpath", ".", "-L."],
```
jacob-carlborg: I don't understand why this is necessary.

dstep/translator/IncludeHandler.d
Mitten-O: I got an ambiguity about the map in mambo and the standard library. I found a cleaner way to resolve it.

dub.json
Mitten-O: I got linker errors from tango without it. Ubuntu 15.04, ldc 0.14.0.

Mitten-O:
> There is already a simple test for includes. Not sure if that's enough.
The added test checks that imports are automatically generated from included files.

The imports are actually generated only for headers whose types are actually used in the file. This is arguably unintuitive, but it works for my needs and was the most fluent way to add it to the existing logic. I added a note about this in the dummy include README.

dstep/translator/IncludeHandler.d
```diff
@@ -109,12 +121,39 @@ class IncludeHandler
         imports ~= "core.stdc.config";
     }
 
+    @property{
```
jacob-carlborg: I would prefer that you put the `@property` attribute on each method.

dstep/translator/IncludeHandler.d
```diff
@@ -109,12 +121,39 @@ class IncludeHandler
...
+        /// Makes includes that match regex filter be converted to import with prefix.
+        void autoImportFilter (string filter)
+        {
+            this.convertIncludes = true;
+            this.convertableIncludePattern = regex(filter);
+        }
+
+        string autoImportFilter()
+        {
+            return this.convertableIncludePattern.stringof;
```
jacob-carlborg: I'm pretty sure this will return `"convertableIncludePattern"` and not the actual regular expression. I haven't been able to find how to extract a string representation of a regular expression, but you could store the string separately.

jacob-carlborg:
> The imports are actually generated only for headers whose types are actually used in the file. This is arguably unintuitive, but it works for my needs and was the most fluent way to add it to the existing logic

One could argue that C includes are not very intuitive :smiley:. The problem with C includes is that they're not reliable. Often includes are missing because they're included implicitly through other includes. In some cases it's not even possible to add all required includes even if you wanted to. I have seen header files that don't include a single header file yet they still use types declared in other files.

Instead DStep looks at type (or similar) that is used in a header file and asks Clang where it's declared. Then it maps that answer from Clang to a corresponding D module.

dub.json
jacob-carlborg: That's weird. Can you see  in the linker error which function is using libz? Could you please try with DMD as well. I've never compiled DStep with LDC.

Mitten-O: Here is the full output:

The determined compiler type "ldc" doesn't match the expected type "dmd". This will probably result in build errors.
Target tango 1.0.1+2.067 is up to date. Use --force to rebuild.
Target mambo 0.0.4 is up to date. Use --force to rebuild.
Target dstack 0.0.3 is up to date. Use --force to rebuild.
Building dstep 0.1.1+commit.34.g582f71b configuration "default", build type debug.
Running ldmd2...
/usr/bin/ld: ../../../.dub/packages/tango-1.0.1_2.067/libtango.a(libtango.o): undefined reference to symbol 'inflateInit2_'
/lib/x86_64-linux-gnu/libz.so.1: error adding symbols: DSO missing from command line
collect2: error: ld returned 1 exit status
Error: /usr/bin/gcc failed with status: 1
FAIL .dub/build/default-debug-linux.posix-x86_64-ldc_2065-16CE78E47BDF9B160EF11DDF27A84EA1/ dstep executable
Error executing command build:
ldmd2 failed with exit code 1.

Seems to be missing inflateInit2.

Works fine without it when using dmd, but isn't this a small price for supporting ldc (or Ubuntu, or whatever has the bug that causes this)?

dstep/translator/IncludeHandler.d
Mitten-O: I just removed the getter. It was only there for the sake of completeness.

Mitten-O:
I moved the attribute to its proper place. Yes, that looks much nicer. Thanks for bearing with me, I'm new to D. Even worse, because of my job, I'm pretty steeped in the dark age logic of c++.

BTW, if and when I finish polishing this to the standards of the project, would you like it to be rebased into a neater story or do you want to keep the full journey?

jacob-carlborg:
> BTW, if and when I finish polishing this to the standards of the project, would you like it to be rebased into a neater story or do you want to keep the full journey?

In this case I think it's best to squash all commits to one.

dstep/translator/IncludeHandler.d
```diff
@@ -133,11 +172,33 @@ private:
 
     string isKnownInclude (string include)
     {
-        include = Path.stripExtension(include);
+        include = Path.stripExtension (include);
```
jacob-carlborg: There should be no space between the function name and the opening parenthesis.

dstep/translator/IncludeHandler.d
```diff
@@ -133,11 +172,33 @@ private:
...
         return null;
     }
-}
\ No newline at end of file
+
+    /// Checks if the given include file name should be converted to an import declaration.
+    bool isConvertableInclude (string include)
+    {
+        // Do not try to convert empty strings, no matter what the pattern says.
+        return include != "" && cast(bool)(matchFirst(include, convertableIncludePattern));
```
jacob-carlborg: I don't think it's necessary to explicitly convert to `bool`. Look at the example for `Captures` [1].

[1] http://dlang.org/phobos/std_regex.html#Captures

dstep/translator/IncludeHandler.d
```diff
@@ -133,11 +172,33 @@ private:
...
+    }
+
+
+    /// Generates an importable module name from an include file name.
+    string autoConvertInclude (string include)
+    {
+        string prefix = this.importPrefix;
+
+        // If we have a valid prefix, append a dot to it if it doesn't have one at the end.
+        if (prefix != "" && prefix != null && prefix[prefix.length-1] != '.')
```
jacob-carlborg: You can use `endsWith` instead of these three conditions.

dstep/translator/IncludeHandler.d
```diff
@@ -133,11 +172,33 @@ private:
...
+        {
+            prefix = prefix ~ ".";
```
jacob-carlborg: You can use the append operator `~=` instead.

dstep/translator/IncludeHandler.d
```diff
@@ -109,12 +121,33 @@ class IncludeHandler
         imports ~= "core.stdc.config";
     }
 
+
+    /// Makes includes that match regex filter be converted to import with prefix.
+    @property void autoImportPrefix (string prefix)
+    {
+        this.convertIncludes = true;
+        this.importPrefix = prefix;
```
jacob-carlborg: Please don't specify `this` unless it's necessary.

Mitten-O:
I'm back! I fixed the latest style deviations and squashed everything together. Also, it seems with the updates you have done, the need for the "lz" flag is gone so I got rid of that addition. So good updates, thanks.

jacob-carlborg:
I have been thinking about this. I'm wondering if it's possible to automatically translate all includes as long as the files are nested within a given root directory. Given a directory layout as:

```
foo
  |- bar
  |   | - a.h
  |
  b.h
```

And `b.h` containing: `#include "bar/a.h"`. When running DStep like `dstep b.h --root foo`, `b.h` would be translated to `import foo.bar.a;`, because `a.h` is located inside the given root, `foo`. Thoughts?

Mitten-O:
You mean to actually look for the includes in the file system and run dstep on them, too? That sounds rather useful. For projects that have one master header that includes the rest, one could translate the whole thing with one command.

My main interest lies in wrapping libgit2 for D (the available bindings are a bit outdated) and their header structure is fairly flat, so nested translation is not entirely necessary for my needs, though.

The other interpretation of what you suggest is to just translate the names, not reading the file system or anything, just converting includes that start with the root name to dotted form. That would be simpler to implement and certainly help already.

Either way a bit beyond the scope of this pull request. It's perfectly fine if you choose to reject this if my changes are in the way of some proper vision for include conversion.

jacob-carlborg:
> You mean to actually look for the includes in the file system and run dstep on them, too? That sounds rather useful. For projects that have one master header that includes the rest, one could translate the whole thing with one command.

No, that's not what I'm thinking.

> The other interpretation of what you suggest is to just translate the names, not reading the file system or anything, just converting includes that start with the root name to dotted form. That would be simpler to implement and certainly help already.

That's kind of what I'm thinking.

DStep would continue to only add imports for what's used. The include directives are not reliable since includes are basically like public imports in D. I've also seen header files without a single include directive, yet they refer to symbols not present in the header file.

The idea of the `--root` flag is to get the top level package in D, and all the nested packages. For any symbol that is referenced in a header file that is declared in some other header file, which path is included in the root directory, specified with the `--root` flag, will be translated to an import statement.

Not sure if that explanation clarifies anything :smiley:. Would that be of any help for you?

> Either way a bit beyond the scope of this pull request. It's perfectly fine if you choose to reject this if my changes are in the way of some proper vision for include conversion.

I don't think what I have in mind is any more complicated than this pull request (I hope :smiley:). I'm not saying that you need to implement my suggestion instead.

Mitten-O:
Ah, I think I get you now. In cases where there are no subfolders in `foo`, your `dstep b.h --root foo` would be equivalent with my `dstep b.h --include-filter "foo/*" --include-prefix "foo"`. It would have nesting, which my solution does not, and be more semantic, but less powerful than regexes. You're right, it does not sound too complicated. But it is a solution to the same problem my changes are trying to solve, so if you do go that way, this pull request will be obsolete.

PR #45, opened 2016-01-18
Extend the linker flags to get builds on Debian Sid and Fedora Rawhide.

russel:
These are the two extra linker flags that allowed builds on Debian Sid and Fedora Rawhide.


jacob-carlborg:
:ok_hand:

Issue #35, opened 2015-01-20
Structs with a terminating flexible array member trigger a segfault


joakim-noah:
For example, this simple header file will get dstep to segfault:

```C
struct foobar {
    int state;
    int info[];
};
```

Removing the brackets on the last member will get dstep to work again, [some info on flexible array members](https://www.securecoding.cert.org/confluence/display/seccode/void+MEMxx-C.+Understand+how+flexible+array+members+are+to+be+used).

joakim-noah:
[As I just noted in the forum](http://forum.dlang.org/thread/mailman.33.1247093746.14071.digitalmars-d@puremagic.com?page=2#post-wbvejavjgatpbrwxirzd:40forum.dlang.org), it appears the right way to translate this would be to `int[0] info;`.

jacob-carlborg:
That should be straightforward to implement.

Issue #26, opened 2014-06-08
dstep dumps core on a simple header


tbrowder:
Given a file t.h, with contents:

extern const char *const sys_errlist[];

Run dstep on it and get a core dump:

$ dstep -x c -o t.d t.c
Segmentation fault (core dumped)

I built dtep from this repo on a Debian Linux 64-bit system (after building and installing the dmd suite from its repos).

rlonstein:
(bump)

Any chance of this being fixed?

jacob-carlborg:
Yes, of course. I'll see if I can get it fixed during the weekend.

jacob-carlborg:
I haven't actually fixed this yet (I have worked improving the test infrastructure) but I can only reproduce this with libclang 3.4.x. Version 3.1-3.3 and 3.5.0 seems to be ok.

rlonstein:
I'll test on Linux and report back. Thanks.

rlonstein:
Works as expected.

PR #48, opened 2016-01-24
Two minor array binding fixes

Dicebot:
First commit fixes root issue behind #26
Second ensures that #35 actually works (it seemed working but
bug is still open)

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -23,7 +23,7 @@ The source code is available under the [Boost Software License 1.0](http://www.b
 
 ### Requirements
 
-* libclang - [http://clang.llvm.org](http://clang.llvm.org) - 3.1 or later that is binary compatible with 3.1
+* libclang - [http://clang.llvm.org](http://clang.llvm.org) - 3.4 or later that is binary compatible with 3.4
 * DMD - [http://dlang.org/download.html](http://dlang.org/download.html) - 2.069.2
 * Dub [http://code.dlang.org/download](http://code.dlang.org/download)
 
--- a/dstep/translator/Type.d
+++ b/dstep/translator/Type.d
@@ -16,6 +16,8 @@ import dstep.translator.IncludeHandler;
 import dstep.translator.Translator;
 import dstep.translator.Output;
 
+import std.conv;
+
 string translateType (Type type, bool rewriteIdToObjcObject = true, bool applyConst = true)
 in
 {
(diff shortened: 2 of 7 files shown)


jacob-carlborg:
Tests are failing.

mihails-strasuns-sociomantic:
Yep, checking travis output.

mihails-strasuns-sociomantic:
Oh looks like it produced a different AST in clang 3.1:

```
Testing with libclang version 3.1
...
    +extern __gshared <unimplemented> sys_errlist;
```

Is there a way to query libclang version from the bindings?

jacob-carlborg:
> Is there a way to query libclang version from the bindings?

There's [clang_getClangVersion](http://clang.llvm.org/doxygen/group__CINDEX__MISC.html#ga307e960e9dccaf721c0032cd4edd8908).

mihails-strasuns-sociomantic:
Thanks. I'll abuse the CI a bit to get necessary debugging info.

mihails-strasuns-sociomantic:
Nasty. It says `CXType_Unexposed` for extern static array types in old clang and doesn't even parse it as valid declaration. I'll try finding a workaround but it may be a dead end. Are there version-based exceptions in your test system?

mihails-strasuns-sociomantic:
Yep, for clang 3.3 and before type of such variable is completely unknown from a cursor - `CXType_Unexposed`, empty `spelling` string, attempts to force `ArrayType` result in invalid element type and -1 size.

@jacob-carlborg waiting for your call on this.

jacob-carlborg:
So you're basically saying that this can' work for Clang 3.3 and before?

mihails-strasuns-sociomantic:
Yes, at least it is beyond my libclang knowledge (have no idea what to do if parser procuded cursor is missing required information). Still, making it work for 3.4+ is better than keeping it broken completely ;)

jacob-carlborg:
Ok, I guess we can drop support for 3.3 and before. Could you please create a separate commit for that explaining the reason in the commit message. Remove the versions from the test and update the readme.

mihails-strasuns-sociomantic:
Ok, green now. NB: some OS/platform combinations now have empty clang version array - you may want to add some new tested versions there.

jacob-carlborg:
:ok_hand:

jacob-carlborg:
Thanks.

mihails-strasuns-sociomantic:
Thanks!

PR #51, opened 2016-02-15
Basic defines translation.

ciechowoj:
I implemented a translation of basic define directives (like #define FOO 0 -> enum FOO = 0;). There is some things that may be improved or may be unacceptable to be merged, but I'm creating this request to spark some discussion about the dstep implementation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,5 @@ foo.d
 clangs
 *.gch
 bin/test
+bin/unittest
 .DS_Store
--- a/clang/Cursor.d
+++ b/clang/Cursor.d
@@ -9,7 +9,11 @@ module clang.Cursor;
 import mambo.core._;
 
 import clang.c.Index;
+import clang.Index;
 import clang.SourceLocation;
+import clang.SourceRange;
+import clang.Token;
+import clang.TranslationUnit;
 import clang.Type;
 import clang.Util;
 import clang.Visitor;
@@ -18,28 +22,54 @@ struct Cursor
 {
     mixin CX;
 
+    private static const CXCursorKind[string] predefined;
+
+    static this()
+    {
+        predefined = queryPredefined();
+    }
+
     @property static Cursor empty ()
     {
         auto r = clang_getNullCursor();
         return Cursor(r);
     }
 
-    @property string spelling ()
+    @property string spelling () const
     {
         return toD(clang_getCursorSpelling(cx));
     }
 
-    @property CXCursorKind kind ()
+    @property CXCursorKind kind () const
     {
         return clang_getCursorKind(cx);
     }
 
-    @property SourceLocation location ()
+    @property SourceLocation location () const
     {
         return SourceLocation(clang_getCursorLocation(cx));
     }
 
-    @property Type type ()
+    @property TokenRange tokens() const
+    {
+        import std.algorithm.mutation : stripRight;
+
+        CXTranslationUnit unit = clang_Cursor_getTranslationUnit(cx);
+        CXToken* tokens = null;
+        uint numTokens = 0;
+        clang_tokenize(unit, extent.cx, &tokens, &numTokens);
+        auto result = TokenRange(unit, tokens, numTokens);
+
+        // For some reason libclang returns some tokens out of cursors extent.cursor
+        return result.stripRight!(token => !intersects(extent, token.extent));
+    }
+
+    @property SourceRange extent() const
+    {
+        return SourceRange(clang_getCursorExtent(cx));
+    }
+
+    @property Type type () const
     {
         auto r = clang_getCursorType(cx);
         return Type(r);
@@ -80,9 +110,34 @@ struct Cursor
         return clang_Cursor_isNull(cx) != 0;
     }
 
-    @property Visitor all ()
+    @property Visitor all () const
     {
         return Visitor(this);
+    }
+
+    @property Cursor[] children(bool ignorePredefined = false) const
+    {
+        import std.array : appender;
+        import std.stdio;
+
+        Cursor[] result = [];
+        auto app = appender(result);
+
+        if (ignorePredefined && isTranslationUnit)
+        {
+            foreach (cursor, _; all)
+            {
+                if (!cursor.isPredefined)
+                    app.put(cursor);
+            }
+        }
+        else
+        {
+            foreach (cursor, _; all)
+                app.put(cursor);
+        }
+
+        return app.data;
     }
 
     @property CXLanguageKind language ()
@@ -105,7 +160,34 @@ struct Cursor
         return clang_isCursorDefinition(cast(CXCursor) cx) != 0;
     }
 
-    @property Cursor definition () const
+    bool isTranslationUnit() const
+    {
+        return clang_isTranslationUnit(kind) != 0;
+    }
+
+    private static CXCursorKind[string] queryPredefined()
+    {
+        CXCursorKind[string] result;
+
+        Index index = Index(false, false);
+        TranslationUnit unit = TranslationUnit.parseString(
+            index,
+            "",
+            []);
+
+        foreach (cursor; unit.cursor.children)
+            result[cursor.spelling] = cursor.kind;
+
+        return result;
+    }
+
+    bool isPredefined() const
+    {
+        auto xkind = spelling in predefined;
+        return xkind !is null && *xkind == kind;
+    }
+
+    @property Cursor definition() const
     {
         return Cursor(clang_getCursorDefinition(cast(CXCursor) cx));
     }
--- a/clang/File.d
+++ b/clang/File.d
@@ -13,8 +13,13 @@ struct File
 {
     mixin CX;
 
-    @property string name ()
+    string name()
     {
         return toD(clang_getFileName(cx));
+    }
+
+    string toString()
+    {
+        return name;
     }
 }
--- a/clang/SourceLocation.d
+++ b/clang/SourceLocation.d
@@ -22,12 +22,29 @@ struct SourceLocation
         uint offset;
     }
 
-    @property Spelling spelling ()
+    @property Spelling spelling() const
     {
         Spelling spell;
 
         clang_getSpellingLocation(cx, &spell.file.cx, &spell.line, &spell.column, &spell.offset);
 
         return spell;
+    }
+
+    @property size_t offset() const
+    {
+        return spelling.offset;
+    }
+
+    @property string path() const
+    {
+        return spelling.file.name;
+    }
+
+    @property string toString() const
+    {
+        import std.format: format;
+        auto s = spelling;
+        return format("SourceLocation(file = %s, line = %d, column = %d, offset = %d)", s.file, s.line, s.column, s.offset);
     }
 }
--- a/clang/SourceRange.d
+++ b/clang/SourceRange.d
@@ -1,0 +1,46 @@
+/**
+ * Copyright: Copyright (c) 2016 Wojciech Szęszoł. All rights reserved.
+ * Authors: Wojciech Szęszoł
+ * Version: Initial created: Feb 14, 2016
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+
+module clang.SourceRange;
+
+import std.conv;
+import clang.c.Index;
+import clang.SourceLocation;
+import clang.Util;
+
+struct SourceRange
+{
+    mixin CX;
+
+    @property SourceLocation start() const
+    {
+        return SourceLocation(clang_getRangeStart(cx));
+    }
+
+    @property SourceLocation end() const
+    {
+        return SourceLocation(clang_getRangeEnd(cx));
+    }
+
+    @property string path() const
+    {
+        return start.path;
+    }
+
+    @property string toString() const
+    {
+        import std.format: format;
+        return format("SourceRange(start = %s, end = %s)", start, end);
+    }
+}
+
+bool intersects(in SourceRange a, in SourceRange b)
+{
+    return a.path == b.path &&
+        (a.start.offset <= b.start.offset && b.start.offset < a.end.offset) ||
+        (a.start.offset < b.end.offset && b.end.offset <= a.end.offset);
+}
--- a/clang/Token.d
+++ b/clang/Token.d
@@ -1,0 +1,148 @@
+/**
+ * Copyright: Copyright (c) 2016 Wojciech Szęszoł. All rights reserved.
+ * Authors: Wojciech Szęszoł
+ * Version: Initial created: Feb 14, 2016
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+ 
+module clang.Token;
+
+import std.typecons;
+import std.conv : to;
+
+import clang.c.Index;
+import clang.SourceLocation;
+import clang.SourceRange;
+import clang.Type;
+import clang.Util;
+import clang.Visitor;
+import clang.Cursor;
+
+struct Token
+{
+    private struct Container
+    {
+        CXTranslationUnit translationUnit;
+        CXToken* tokens;
+        ulong numTokens;
+
+        ~this()
+        {
+            if (tokens != null)
+            {
+                clang_disposeTokens(
+                    translationUnit,
+                    tokens,
+                    to!uint(numTokens));
+            }
+        }
+    }
+
+    private const RefCounted!Container container;
+    size_t index;
+
+    @property private Container* containerPtr() const
+    {
+        return cast(Container*) &(container.refCountedPayload());
+    }
+
+    @property static Cursor empty ()
+    {
+        auto r = clang_getNullCursor();
+        return Cursor(r);
+    }
+
+    @property string spelling() const
+    {
+        return toD(clang_getTokenSpelling(
+            containerPtr.translationUnit,
+            containerPtr.tokens[index]));
+    }
+
+    @property CXTokenKind kind() const
+    {
+        return clang_getTokenKind(containerPtr.tokens[index]);
+    }
+
+    @property SourceRange extent() const
+    {
+        return SourceRange(
+            clang_getTokenExtent(
+                containerPtr.translationUnit,
+                containerPtr.tokens[index]));
+    }
+
+    @property string toString() const
+    {
+        import std.format: format;
+        return format("Token(%s, %s)", kind, spelling);
+    }
+}
+
+struct TokenRange
+{
+    private const RefCounted!(Token.Container) container;
+    private size_t begin;
+    private size_t end;
+
+    private static RefCounted!(Token.Container) makeContainer(
+        CXTranslationUnit translationUnit,
+        CXToken* tokens,
+        ulong numTokens)
+    {
+        RefCounted!(Token.Container) result;
+        result.translationUnit = translationUnit;
+        result.tokens = tokens;
+        result.numTokens = numTokens;
+        return result;
+    }
+
+    this(CXTranslationUnit translationUnit,
+        CXToken* tokens,
+        ulong numTokens)
+    {
+        container = makeContainer(translationUnit, tokens, numTokens);
+        begin = 0;
+        end = numTokens;
+    }
+
+    @property bool empty() const
+    {
+        return begin >= end;
+    }
+
+    @property Token front() const
+    {
+        return Token(container, begin);
+    }
+
+    @property Token back() const
+    {
+        return Token(container, end - 1);
+    }
+
+    @property void popFront()
+    {
+        ++begin;
+    }
+
+    @property void popBack()
+    {
+        --end;
+    }
+
+    @property TokenRange save() const
+    {
+        return this;
+    }
+
+    @property size_t length() const
+    {
+        return end - begin;
+    }
+
+    Token opIndex(size_t index) const
+    {
+        return Token(container, begin + index);
+    }
+}
--- a/clang/TranslationUnit.d
+++ b/clang/TranslationUnit.d
@@ -24,7 +24,7 @@ struct TranslationUnit
 
     static TranslationUnit parse (Index index, string sourceFilename, string[] commandLineArgs,
         CXUnsavedFile[] unsavedFiles = null,
-        uint options = CXTranslationUnit_Flags.CXTranslationUnit_None)
+        uint options = CXTranslationUnit_Flags.CXTranslationUnit_DetailedPreprocessingRecord)
     {
         return TranslationUnit(
             clang_parseTranslationUnit(
@@ -35,6 +35,33 @@ struct TranslationUnit
                 toCArray!(CXUnsavedFile)(unsavedFiles),
                 cast(uint) unsavedFiles.length,
                 options));
+    }
+
+    static TranslationUnit parseString (
+        Index index,
+        string source,
+        string[] commandLineArgs,
+        CXUnsavedFile[] unsavedFiles = null,
+        uint options = CXTranslationUnit_Flags.CXTranslationUnit_DetailedPreprocessingRecord)
+    {
+        import std.file;
+
+        auto file = namedTempFile("dstep", ".h");
+        auto name = file.name();
+        file.write(source);
+        file.flush();
+        file.detach();
+
+        auto translationUnit = TranslationUnit.parse(
+            index,
+            name,
+            commandLineArgs,
+            unsavedFiles,
+            options);
+
+        remove(name);
+
+        return translationUnit;
     }
 
     private this (CXTranslationUnit cx)
--- a/clang/Util.d
+++ b/clang/Util.d
@@ -6,11 +6,12 @@
  */
 module clang.Util;
 
+import std.conv;
+import std.stdio;
+
 import mambo.core._;
 
 import clang.c.Index;
-
-import std.conv;
 
 immutable(char*)* strToCArray (string[] arr)
 {
@@ -76,4 +77,42 @@ mixin template CX ()
     {
         return cx !is CType.init;
     }
+}
+
+extern (C) int mkstemps(char*, int);
+extern (C) int close(int);
+
+class NamedTempFileException : object.Exception
+{
+    this (string message, string file = __FILE__, size_t line = __LINE__)
+    {
+        super(message, file, line);
+    }
+}
+
+File namedTempFile(string prefix, string suffix) 
+{
+    import std.random;
+    import std.file;
+    import std.path;
+    import std.format;
+
+    void randstr (char[] slice)
+    {
+        for (uint i = 0; i < slice.length; ++i)
+            slice[i] = uniform!("[]")('A', 'Z');
+    }
+
+    string name = format("%sXXXXXXXXXXXXXXXX%s\0", prefix, suffix);
+    char[] path = buildPath(tempDir(), name).dup;
+    const size_t termAnd6XSize = 7;
+    randstr(path[$ - name.length + prefix.length .. $ - suffix.length - termAnd6XSize]);
+
+    int fd = mkstemps(path.ptr, cast(int) suffix.length);
+    scope (exit) close(fd);
+
+    if (fd == -1)
+        throw new NamedTempFileException("Cannot create \"%s\" temporary file.".format(path));
+
+    return File(path[0..$-1], "wb+");
 }
--- a/dstep/translator/Translator.d
+++ b/dstep/translator/Translator.d
@@ -57,9 +57,21 @@ class Translator
         inputFile = translationUnit.file(inputFilename);
     }
 
+    this (TranslationUnit translationUnit, const Options options = Options.init)
+    {
+        this.translationUnit = translationUnit;
+        outputFile = options.outputFile;
+        language = options.language;
+    }
+
     void translate ()
     {
-        foreach (cursor, parent ; translationUnit.declarations)
+        write(outputFile, translateToString());
+    }
+
+    string translateToString()
+    {
+        foreach (cursor, parent; translationUnit.cursor.all)
         {
             if (skipDeclaration(cursor))
                 continue;
@@ -68,6 +80,7 @@ class Translator
             auto code = translate(cursor, parent);
 
             with (CXCursorKind)
+            {
                 switch (cursor.kind)
                 {
                     case CXCursor_ObjCInterfaceDecl:
@@ -85,33 +98,37 @@ class Translator
                     case CXCursor_VarDecl: output.variables ~= code; break;
                     case CXCursor_FunctionDecl: output.functions ~= code; break;
                     case CXCursor_TypedefDecl: output.typedefs ~= code; break;
-
+                    case CXCursor_MacroDefinition:
+                        if (code != "") 
+                            output.defines ~= code;
+                        break;
                     default: continue;
                 }
+            }
         }
 
         output.structs ~= deferredDeclarations.values;
         output.externDeclaration = externDeclaration();
 
-        auto data = output.toString;
-        write(outputFile, data);
+        return output.toString;
     }
 
     string translate (Cursor cursor, Cursor parent = Cursor.empty)
     {
         with (CXCursorKind)
+        {
             switch (cursor.kind)
             {
                 case CXCursor_ObjCInterfaceDecl:
-                    return (new ObjcInterface!(ClassData)(cursor, parent, this)).translate;
+                    return (new ObjcInterface!(ClassData)(cursor, parent, this)).translate();
                 break;
 
                 case CXCursor_ObjCProtocolDecl:
-                    return (new ObjcInterface!(InterfaceData)(cursor, parent, this)).translate;
+                    return (new ObjcInterface!(InterfaceData)(cursor, parent, this)).translate();
                 break;
 
                 case CXCursor_ObjCCategoryDecl:
-                    return (new Category(cursor, parent, this)).translate;
+                    return (new Category(cursor, parent, this)).translate();
                 break;
 
                 case CXCursor_VarDecl:
@@ -150,13 +167,19 @@ class Translator
                         return "";
                     }
                     break;
-                case CXCursor_EnumDecl: return (new Enum(cursor, parent, this)).translate; break;
-                case CXCursor_UnionDecl: return (new Record!(UnionData)(cursor, parent, this)).translate; break;
+                case CXCursor_EnumDecl: 
+                    return (new Enum(cursor, parent, this)).translate(); break;
+
+                case CXCursor_UnionDecl: 
+                    return (new Record!(UnionData)(cursor, parent, this)).translate(); break;
+
+                case CXCursor_MacroDefinition:
+                    return translateDefine(cursor);
 
                 default:
                     return "";
-                    //assert(0, `Translator.translate: missing implementation for "` ~ cursor.kind.toString ~ `".`);
             }
+        }
     }
 
     string variable (Cursor cursor, String context = null)
@@ -184,8 +207,9 @@ class Translator
 private:
 
     bool skipDeclaration (Cursor cursor)
-    {
-        return inputFile != cursor.location.spelling.file;
+    {   
+        return (inputFilename != "" && inputFile != cursor.location.spelling.file) 
+            || cursor.isPredefined;  
     }
 
     string externDeclaration ()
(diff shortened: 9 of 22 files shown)


jacob-carlborg:
Interesting, I'll have a look at this as soon as I have some time.

jacob-carlborg:
First some general comments about the style guide:

* Use four space for indentation
* Braces on its own line
* Consistent module copyright notice
* @property on the same line as the function declarations, or skip it completely (it doesn't do anything useful)

jacob-carlborg:
Remove any Sublime related configs.

clang/Cursor.d
```diff
@@ -105,7 +169,184 @@ struct Cursor
         return clang_isCursorDefinition(cast(CXCursor) cx) != 0;
     }
 
-    @property Cursor definition () const
+    bool isTranslationUnit() const
+    {
+        auto kind = clang_getCursorKind(cast(CXCursor) cx);
```
jacob-carlborg: There's already `kind` that can be used.

clang/Cursor.d
```diff
@@ -105,7 +169,184 @@ struct Cursor
...
+        return clang_isTranslationUnit(kind) != 0;
+    }
+
+    private bool _isPredefined() const
+    {
+        immutable ubyte[string] predefined = [
```
jacob-carlborg: The predefined macro symbols are platform dependent, that needs to be handled.

clang/Cursor.d
```diff
@@ -105,7 +169,184 @@ struct Cursor
...
+            "__STDC_UTF_16__" : 0,
+            "__STDC_UTF_32__" : 0,
+            "__int128_t" : 0,
+            "__uint128_t" : 0,
+            "__builtin_va_list" : 0,
+            "unix" : 0,
+            "linux" : 0,
+        ];
+
+        return (spelling in predefined) != null;
```
jacob-carlborg: `null` should always be compared with `is`, or let the implicit cast to bool handle it.

clang/File.d
```diff
@@ -13,8 +13,15 @@ struct File
 {
     mixin CX;
 
-    @property string name ()
+    @property 
+    string name() const
     {
-        return toD(clang_getFileName(cx));
+        return toD(clang_getFileName(cast(void*) cx));
```
jacob-carlborg: Why the cast?

clang/SourceLocation.d
```diff
@@ -30,4 +31,24 @@ struct SourceLocation
...
+    {
+        return spelling.file.name;
+    }
+
+    @property
+    string toString() const
+    {
+        import std.format: format;
+        auto s = spelling;
+        return format("%s(%d,%d,%d)", s.file, s.line, s.column, s.offset);
```
jacob-carlborg: Please include the type (SourceLocation) and the fields in the string.

clang/SourceRange.d
```diff
@@ -0,0 +1,45 @@
...
+ */
+
+module clang.SourceRange;
+import std.conv;
+import clang.c.Index;
+import clang.SourceLocation;
+
+struct SourceRange
+{
+    CXSourceRange _impl;
```
jacob-carlborg: Any reason the `CX` mixin is not used?

clang/SourceRange.d
```diff
@@ -0,0 +1,45 @@
...
+    string path() const 
+    {
+        return start.path;
+    }
+
+    @property
+    string toString() const 
+    {
+        import std.format: format;
+        return format("SourceRange(%s, %s)", to!string(start), to!string(end));
```
jacob-carlborg: * Please include the field names
* `format` will convert all its arguments to string, that's the whole point :wink:. No need to call `to!string`

clang/Token.d
```diff
@@ -0,0 +1,159 @@
...
+        import std.format: format;
+        import std.conv: to;
+        return format("Token(%s, %s)", to!string(kind), spelling);
+    }
+}
+
+struct TokenRange
+{
+    private const RefCounted!(Token.Container) _container;
+    private size_t _begin;
```
jacob-carlborg: Please don't prefix symbols with `_`.

dstep/translator/Translator.d
```diff
@@ -57,17 +58,48 @@ class Translator
...
+        {
+            if (skipDeclaration(cursor) || isInternalIdentifier(cursor.spelling))
+            {
+                continue;
+            }
+
+            output.newContext();
+            auto code = translate(cursor, parent);
+
+            if (!code.isNull && cursor.kind == CXCursorKind.CXCursor_MacroDefinition)
```
jacob-carlborg: Isn't it better to check the cursor kind before translating?

dstep/translator/Translator.d
```diff
@@ -57,17 +58,48 @@ class Translator
...
+    }
+
     void translate ()
     {
+        write(outputFile, translateToString());
+    }
+
+    string translateToString()
+    {
+        foreach (cursor, parent ; translationUnit.notDeclarations)
```
jacob-carlborg: Why is this a separate foreach loop?

dstep/translator/Translator.d
```diff
@@ -85,33 +117,35 @@ class Translator
...
         output.structs ~= deferredDeclarations.values;
         output.externDeclaration = externDeclaration();
 
-        auto data = output.toString;
-        write(outputFile, data);
+        return output.toString;
     }
 
-    string translate (Cursor cursor, Cursor parent = Cursor.empty)
+    Nullable!string translate (Cursor cursor, Cursor parent = Cursor.empty)
```
jacob-carlborg: What does `Nullable` add?

dstep/translator/Translator.d
```diff
@@ -184,8 +226,25 @@ class Translator
...
-    {
-        return inputFile != cursor.location.spelling.file;
+    {   
+        string spelling = cursor.spelling;
+
+        if (inputFilename != "" && inputFile != cursor.location.spelling.file)
+        {
+            return true;
+        }
+        else if (spelling == "__int128_t" || spelling == "__uint128_t")
```
jacob-carlborg: Isn't this already handled by `isInternalIdentifier`?

dstep/translator/Translator.d
```diff
@@ -311,6 +386,22 @@ void handleInclude (Type type)
     includeHandler.addInclude(getInclude(type));
 }
 
+bool isInternalIdentifier(string identifier)
```
jacob-carlborg: Why not use `_isPredefined`?

features/step_definitions/common.rb
```diff
@@ -22,6 +22,7 @@
 
 Then /^the files "([^"]*)" and "([^"]*)" should be equal$/ do |file1, file2|
   check_file_content(file1, IO.read(file2), true)
+  # expect(file1).to have_file_content(IO.read(file2))
```
jacob-carlborg: Remove.

clang/File.d
ciechowoj: To remove const.

clang/SourceLocation.d
ciechowoj: Right.

clang/Token.d
ciechowoj: Eh, I don't like this because '_' simplifies some things - e.g. you do not have name collisions with function arguments (common case in constructor). But, ok...

ciechowoj:
"Remove any Sublime related configs."

May I leave exceptions in .gitignore?

clang/Cursor.d
ciechowoj: Fixed.

clang/Cursor.d
ciechowoj: Fixed.

features/step_definitions/common.rb
ciechowoj: Done.

clang/SourceRange.d
ciechowoj: Fixed.

dstep/translator/Translator.d
ciechowoj: Right, it is.

clang/Cursor.d
ciechowoj: Fixed, now I consider all symbols that are defined in **empty** translation unit internal. Another way would be to check the symbol extent (SourceRange) predefined symbols doesn't seem to have a file path assigned.

dstep/translator/Translator.d
ciechowoj: Yes, it is - fixed.

dstep/translator/Translator.d
ciechowoj: Fixed.

clang/SourceRange.d
ciechowoj: "Please include the field names"
Currently it looks like "SourceRange(SourceLocation(file,116,9,3735), SourceLocation(file,116,38,3764))" - isn't it good enough?

"format will convert all its arguments to string, that's the whole point :wink:. No need to call to!string"
Good point.

ciechowoj:
I fixed most of minor issues there remains: "Why is this a separate foreach loop?" and "What does Nullable add?".

clang/File.d
```diff
@@ -13,8 +13,13 @@ struct File
 {
     mixin CX;
 
-    @property string name ()
+    string name() const
     {
-        return toD(clang_getFileName(cx));
+        return toD(clang_getFileName(cast(void*) cx));
```
jacob-carlborg: I'm not keen on casting away const, is it really necessary to make `name` const?

clang/SourceRange.d
jacob-carlborg: >  isn't it good enough

I guess, but I think it's more clear if the names are included.

clang/Token.d
jacob-carlborg: I usually handle that with qualifying the instance variables with `this`, when necessary.

dstep/translator/Translator.d
jacob-carlborg: I don't really see a difference except for removing the braces.

clang/SourceRange.d
ciechowoj: Something like "SourceRange(start = SourceLocation(file,116,9,3735), end = SourceLocation(file,116,38,3764))"?

jacob-carlborg:
> May I leave exceptions in .gitignore?

Sorry, but no. You can add a [global ignore file](http://stackoverflow.com/questions/7335420/global-git-ignore) or add it to `.git/info/exclude` for private project dependent ignore.

> I'm not D expert. Does `@property` really do nothing

It does a few things: http://dlang.org/spec/function.html#property-functions

clang/SourceRange.d
jacob-carlborg: Yes, and ideally for `SourceLocation` as well.

clang/File.d
ciechowoj: Hard to say if it is absolutely necessary, but very convenient. And the user of the interface expects that getters that do not change the objects can be called on const object.

clang_getFileName do not modify its argument:



    CXString clang_getFileName(CXFile SFile) {
      if (!SFile)
        return cxstring::createNull();

      FileEntry *FEnt = static_cast<FileEntry *>(SFile);
      return cxstring::createRef(FEnt->getName());
    }

    class FileEntry {
      const char *Name;           // Name of the file.
     
      /* ... */

      const char *getName() const { return Name; }


    CXString createRef(StringRef String) {
      // If the string is not nul-terminated, we have to make a copy.

      // FIXME: This is doing a one past end read, and should be removed! For memory
      // we don't manage, the API string can become unterminated at any time outside
      // our control.

      if (!String.empty() && String.data()[String.size()] != 0)
        return createDup(String);

      CXString Result;
      Result.data = String.data();
      Result.private_flags = (unsigned) CXS_Unmanaged;
      return Result;
    }

(at least for now).

dstep/translator/Translator.d
ciechowoj: Oh, my mistake. It should be good now.

clang/SourceRange.d
ciechowoj: It was painful : ) (I prefer compact things), but I added these field names.

ciechowoj:
> Sorry, but no. You can add a global ignore file or add it to .git/info/exclude for private project dependent ignore.

Done.

> It does a few things: http://dlang.org/spec/function.html#property-functions

Ok, anyway I left them (only moved to the same line as function headers).

clang/File.d
jacob-carlborg: What I mean is, does some other code require `name` to be const? If not, then I would remove const.

ciechowoj: Currently `File::toString()` and because I made it const and yes currently it const can be removed from both methods. But it is only a matter of time when someone needs to call that method on const File. For now I can remove it.

jacob-carlborg: I suggest removing it for now. I think it's better to not use const at all than using const and breaking it. Note that const in D has much stronger guarantees that in C++.

ciechowoj: Done.

jacob-carlborg:
> The switch statement is repeated twice (in translate and translateToString) and it shouldn't

Yeah, that's because it doesn't output the declarations in the same order as the C file.

> I duplicated the foreach loop because existing one iterates only through definitions and I believed it would took some effort to change it to iterate over all symbols (I needed it to iterate on preprocessor directives) and as I focused on adding that define translation I didn't want to rewrite too much unrelated code

Shouldn't it be enough to replace `translationUnit.declarations` with something like `translationUnit.cursor.all`?

> Current implementation changes order of statements with respect to original C file, it is in TODO list already and should be fixed

Agree.

> The Output is global state, it's good for one shoot run of the program but bad for example for unit testing

Agree.

> And I introduced Nullable to let handle the case when underlying translate function cannot translate given statement (so it can return "null" to let rest of the code to know that the statement should be ignored)

I'll have to take another look at how the result is used.

> Preferably I would like to rewrite the whole thing in more functional way and so that there is no global state and the statements are translated in the original order

I agree, but one step at the time.

unittest.d
```diff
@@ -0,0 +1,5 @@
+
+int main() 
+{
+    return 0;
+}
```
jacob-carlborg: This file shouldn't be necessary. You can add `-main` when compiling to let the compiler generate a main function.

jacob-carlborg:
I don't like that you're changing unrelated files, i.e. `.gitmodules` and `dub.selections.json`.

jacob-carlborg:
I'll have to take a look again at all this after all your updates.

ciechowoj:
> I don't like that you're changing unrelated files, i.e. `.gitmodules` and `dub.selections.json`.
Hmm, I did it when replacing tabs with spaces, but maybe unnecessarily. I could revert that but it would require to add new commits. I could try to remove that changes at all and squash the (unnecessary) commits but I'm not really sure how that action will interact with current pull request (when I change history of my own repository).

jacob-carlborg:
> I'm not really sure how that action will interact with current pull request (when I change history of my own repository)

Shouldn't be any problem. Just push with `--force` and the pull request will be updated automatically. BTW, these files are generated by tools (dub and git) so it's best to not touch them at all.

ciechowoj:
> Just push with --force and the pull request will be updated automatically. 
Seems like it worked : ).

ciechowoj:
@Dicebot 
Could you elaborate that git tip further?

jacob-carlborg:
@ciechowoj http://stackoverflow.com/questions/804115/when-do-you-use-git-rebase-instead-of-git-merge

ciechowoj:
Well, something happened.

.gitignore
```diff
@@ -18,4 +18,6 @@ foo.d
 clangs
 *.gch
 bin/test
+bin/unittest
 .DS_Store
+*.log
```
jacob-carlborg: What is this?

clang/Cursor.d
```diff
@@ -18,28 +22,61 @@ struct Cursor
...
+
+        // For some reason libclang returns some tokens out of cursors extent.cursor
+
+        while (!result.empty())
+        {
+            if (!intersects(extent, result.back.extent))
+                result.popBack();
+            else
+                break;
+        }
```
jacob-carlborg: Can't this be done with `std.algorithm.find`?

clang/Cursor.d
```diff
@@ -80,11 +117,38 @@ struct Cursor
...
+                if (!cursor.isPredefined)
+                    app.put(cursor);
+            }
+        }
+        else
+        {
+            foreach (cursor, _; all)
+            {
+                app.put(cursor);
+            }
```
jacob-carlborg: Please remove the curly braces for the foreach loop.

clang/SourceRange.d
```diff
@@ -0,0 +1,45 @@
+/**
+ * Copyright: Copyright (c) 2016 Wojciech Szęszoł. All rights reserved.
+ * Authors: Wojciech Szęszoł
+ * Version: Initial created: Feb 14, 2016
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+
+module clang.SourceRange;
+import std.conv;
```
jacob-carlborg: Newline between module and imports.

clang/Token.d
```diff
@@ -0,0 +1,148 @@
...
+ * Authors: Wojciech Szęszoł
+ * Version: Initial created: Feb 14, 2016
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+ 
+module clang.Token;
+
+import std.typecons;
+import std.conv : to;
+import clang.c.Index;
```
jacob-carlborg: Please group the imports. First standard library, then third party libraries, last current project. Separate with newline.

clang/TranslationUnit.d
```diff
@@ -37,6 +37,61 @@ struct TranslationUnit
...
+
+        string tempname ()
+        {
+            return tempdir() ~ dirSeparator ~ randstr(20) ~ ".dstep.h";
+        }
+
+        string temp = tempdir();
+
+        if (!exists(temp)) 
+            mkdir(temp);
```
jacob-carlborg: Not sure how important this is but I don't think this is safe. If I recall correctly there's a Posix function that does this safely. It actually creates the directory and not just returning the name.

clang/TranslationUnit.d
```diff
@@ -37,6 +37,61 @@ struct TranslationUnit
...
+        }
+
+        string tempdir ()
+        {
+            return tempDir() ~ dirSeparator ~ "dstep";
+        }
+
+        string tempname ()
+        {
+            return tempdir() ~ dirSeparator ~ randstr(20) ~ ".dstep.h";
```
jacob-carlborg: I suggest using `std.path.buildPath` instead.

dstep/translator/Translator.d
```diff
@@ -150,13 +167,21 @@ class Translator
...
-                case CXCursor_EnumDecl: return (new Enum(cursor, parent, this)).translate; break;
-                case CXCursor_UnionDecl: return (new Record!(UnionData)(cursor, parent, this)).translate; break;
+                case CXCursor_EnumDecl: 
+                    return (new Enum(cursor, parent, this)).translate(); break;
+
+                case CXCursor_UnionDecl: 
+                    return (new Record!(UnionData)(cursor, parent, this)).translate(); break;
+
+                case CXCursor_MacroDefinition:
+                case CXCursor_PreprocessingDirective:
```
jacob-carlborg: When will cursor be `CXCursor_PreprocessingDirective`?

dstep/translator/Translator.d
```diff
@@ -199,6 +225,22 @@ private:
...
+    auto tokens = cursor.tokens();
+
+    if (tokens.length == 2) 
+    {
+        import std.format: format;
+        string result = format("enum %s = %s;", tokens[0].spelling, tokens[1].spelling);
+        return result;
+    }
+    else 
+    {
```
jacob-carlborg: Please remove the braces for the else statement.

dstep/translator/Translator.d
```diff
@@ -199,6 +225,22 @@ private:
...
 }
 
+string translateDefine(Cursor cursor) {
+
+    auto tokens = cursor.tokens();
+
+    if (tokens.length == 2) 
+    {
+        import std.format: format;
+        string result = format("enum %s = %s;", tokens[0].spelling, tokens[1].spelling);
```
jacob-carlborg: I don't see the point of storing the result in a variable.

dub.json
```diff
@@ -29,6 +29,17 @@
...
+
+        {
+            "name": "unittest",
+            "mainSourceFile": "test.d",
+            "targetName": "unittest",
+            "sourcePaths": ["dstep", "clang", "unit_tests"],
+            "importPaths": ["dstep"],
+            "lflags-posix": ["-lclang", "-rpath", ".", "-L.", "-L/usr/lib64/llvm", "-L/usr/lib/llvm-3.7/lib"],
+            "lflags-windows": ["+\\", "+clang"],
+            "excludedSourceFiles": ["dstep/driver/*"]
```
jacob-carlborg: Can the standard `dub test` be used for the unit tests?

unit_tests/Common.d
```diff
@@ -0,0 +1,64 @@
...
+}
+
+void assert_translates(string C, string D, string file = __FILE__, size_t line = __LINE__) 
+{
+    auto translated = translate(makeTranslationUnit(C));
+
+    string separator = "----------------";
+
+    string message = "\nC code translated to:\n" ~ separator ~ "\n" ~ translated ~ 
+        "\n" ~ separator ~ "\nExpected D code:\n" ~ separator ~ "\n" ~ D ~ "\n" ~ separator ~ "\n";
```
jacob-carlborg: Perhaps `format` would be better here.

jacob-carlborg:
Please include the unit tests in the Travis build.

.gitignore
ciechowoj: It isn't anymore.

clang/Cursor.d
ciechowoj: Nope, but I've provided another library solution.

clang/Cursor.d
ciechowoj: Ok.

clang/SourceRange.d
ciechowoj: Fixed.

clang/Token.d
ciechowoj: Fixed.

clang/TranslationUnit.d
ciechowoj: It is quite important because libclang API doesn't allow to parse just string (it needs a file in every case, I'm depending on it in unit tests and for checking if the symbol is predefined). I believe it is the best it can be done for now. I'll work on better solution later.

clang/TranslationUnit.d
ciechowoj: Right, it is better idea.

dstep/translator/Translator.d
ciechowoj: Removed, for now `CXCursor_PreprocessingDirective` is irrelevant.

dstep/translator/Translator.d
ciechowoj: I think `else` without braces when `if` has braces looks just ugly. I refactored in anyway.

dstep/translator/Translator.d
ciechowoj: Fixed.

unit_tests/Common.d

dub.json
ciechowoj: Yes. `dub test` runs `unittest` configuration. I needed to add the `unittest.d` file back, because I couldn't find a way to pass that `-main` flag in dub configuration (e.g. removing `mainSourceFile` causes dub to show a warning).

clang/TranslationUnit.d
Dicebot: There is http://dlang.org/library/std/stdio/file.tmpfile.html

ciechowoj: Right, but I need the name actually.

ciechowoj: If you consider current solution too risky, I can add some random suffix to `dstep` sub-directory so that there is no chance of collision with another application.

clang/Cursor.d
```diff
@@ -18,28 +22,54 @@ struct Cursor
...
+        import std.algorithm.mutation : stripRight;
+
+        CXTranslationUnit unit = clang_Cursor_getTranslationUnit(cx);
+        CXToken* tokens = null;
+        uint numTokens = 0;
+        clang_tokenize(unit, extent.cx, &tokens, &numTokens);
+        auto result = TokenRange(unit, tokens, numTokens);
+
+        // For some reason libclang returns some tokens out of cursors extent.cursor
+        return stripRight!((Token token) => !intersects(extent, token.extent))(result);
```
jacob-carlborg: I prefer to use UFCS and type inference. `result.stripRight!(token => ...`

clang/TranslationUnit.d
jacob-carlborg: It's the [mkstemp](http://pubs.opengroup.org/onlinepubs/009695399/functions/mkstemp.html) function that I originally had in mind.

unit_tests/Common.d
```diff
@@ -0,0 +1,63 @@
...
+
+class TranslateAssertError : object.Error 
+{
+    this (string message, string file, ulong line) 
+    {
+        super(message, file, line);
+    }
+}
+
+void assert_translates(string C, string D, string file = __FILE__, size_t line = __LINE__)
```
jacob-carlborg: * Please use camel case for the method name
* Lower case identifiers for the parameters

unit_tests/Common.d
```diff
@@ -0,0 +1,63 @@
...
+public import clang.Token;
+
+Index index;
+
+static this() 
+{
+    index = Index(false, false);
+}
+
+TranslationUnit makeTranslationUnit(string C)
```
jacob-carlborg: Lower case identifiers for the parameters

unit_tests/Common.d
```diff
@@ -0,0 +1,63 @@
...
+        super(message, file, line);
+    }
+}
+
+void assert_translates(string C, string D, string file = __FILE__, size_t line = __LINE__) 
+{
+    import std.format : format;
+
+    auto translated = translate(makeTranslationUnit(C));
+    auto fmt = "\nC code translated to:\n%1$s\n%2$s\n%1$s\nExpected D code:\n%1$s\n%3$s\n%1$s\n";
```
jacob-carlborg: I think this would be more readable using a multi line [delimited string](http://dlang.org/spec/lex.html#DelimitedString).

dub.json
jacob-carlborg: Right. It's confusing that the command is called `test` and the configuration is called `unittest`.

test_files/preprocessor.h
```diff
@@ -0,0 +1,4 @@
+
+#define FOO 0
+#define FOOBAR 1
+#define FOOBARBAZ 2
```
jacob-carlborg: What happens if the define is not an integer? Do we need to check the actual token kind/type of the second token to make sure it can be translated to D?

ciechowoj: Current implementation blindly converts all defines to enum. e.g. if it is string it blindly converts it to enum FOO = "string".

clang/Cursor.d
ciechowoj: Fixed.

unit_tests/Common.d

unit_tests/Common.d

test_files/preprocessor.h
jacob-carlborg: So perhaps the implementation should check the token kind and only translate the ones that can be translated to D.

clang/Cursor.d
```diff
@@ -18,28 +22,54 @@ struct Cursor
...
+        import std.algorithm.mutation : stripRight;
+
+        CXTranslationUnit unit = clang_Cursor_getTranslationUnit(cx);
+        CXToken* tokens = null;
+        uint numTokens = 0;
+        clang_tokenize(unit, extent.cx, &tokens, &numTokens);
+        auto result = TokenRange(unit, tokens, numTokens);
+
+        // For some reason libclang returns some tokens out of cursors extent.cursor
+        return stripRight!((token) => !intersects(extent, token.extent))(result);
```
jacob-carlborg: * UFCS is still not used
* It's not necessary have parentheses around the parameter for a lambda if it only has one parameter

ciechowoj: Sorry, now it should be ok.

test_files/preprocessor.h
ciechowoj: Yes, I have in plans to improve the handling of `#define`s, but lets make it step by step.

jacob-carlborg: Fair enough.

ciechowoj:
I have problems with building unit tests on travis. For unknown reason dmd returns 0, despite there is no error only warnings and deprecation warnings.

Performing "unittest" build using dmd for x86_64.
	tango 1.0.3+2.068: building configuration "static"...
	../../../.dub/packages/tango-1.0.3_2.068/tango/math/IEEE.d(614,17): Warning: instead of C-style syntax, use D-style syntax 'real[3][] vals'
	../../../.dub/packages/tango-1.0.3_2.068/tango/math/IEEE.d(646,18): Warning: instead of C-style syntax, use D-style syntax 'real[3][] extendedvals'
	../../../.dub/packages/tango-1.0.3_2.068/tango/math/Math.d(437,21): Warning: instead of C-style syntax, use D-style syntax 'real[2][] vals'
	../../../.dub/packages/tango-1.0.3_2.068/tango/math/Math.d(1788,17): Warning: instead of C-style syntax, use D-style syntax 'real[3][] vals'
	../../../.dub/packages/tango-1.0.3_2.068/tango/math/Math.d(1906,30): Warning: instead of C-style syntax, use D-style syntax 'real[] pp'
	tango.io.FileScan is deprecated. Please use tango.io.FilePath.toList or tango.io.vfs.* instead.
	tango.math.BigInt is deprecated. Please use std.bigint instead.
	../../../.dub/packages/tango-1.0.3_2.068/tango/math/IEEE.d(1403,9): Deprecation: use std.math.isNaN to deal with NaN operands rather than floating point operator '!<'
	dmd failed with exit code 1.

ciechowoj:
I believe the problem is compiling the tango with `-unittest` flag. But how to make the dub to compile dependencies in usual way and only the target with `-unittest` flag.

jacob-carlborg:
> For unknown reason dmd returns 1, despite there is no error only warnings and deprecation warnings

I think that's because of the deprecation messages. I think there is a way to allow deprecations.

> I believe the problem is compiling the tango with -unittest flag. But how to make the dub to compile dependencies in usual way and only the target with -unittest flag.

Sounds like a bug. You can compile with the `--verbose` flag to see if Dub is passing the `-unittest` flag to the dependencies.

ciechowoj:
`dub` compiles the dependency with 
`dmd -lib -of../../../.dub/packages/tango-1.0.3_2.068/.dub/build/static-unittest-linux.posix-x86_64-dmd_2070-E70DE3E272683A60875B03BE9C7523C9/libtango.a -debug -g -unittest -w -version=Have_tango -I../../../.dub/packages/tango-1.0.3_2.068 ../../../.dub/packages/tango-1.0.3_2.068/tango/core/Array.d ../../../.dub/packages/tango-1.0.3_2.068/tango/core/ArrayLiteral.d ../../../.dub/packages/tango-1.0.3_2.068/tango/core/BitArray.d ../../../.dub/packages/tango-1.0.3_2.068/tango/core/BitManip.d ../../../.dub/packages/tango-1.0.3_2.068/tango/core/ByteSwap.d ../../../.dub/packages/tango-1.0.3_2.068/tango/core/Compiler.d ../../../.dub/packages/tango-1.0.3_2.068/tango/core/Exception.d ../../../.dub/packages/tango-1.0.3_2.068/tango/core/Memory.d ../../../.dub/packages/tango-1.0.3_2.068/tango/core/Octal.d ../../../.dub/packages/tango-1.0.3_2.068/tango/core/Runtime.d ../../../.dub/packages/tango-`

Anyway, it should compile fine even with `-unittest` flags, however this is the major suspect as I cannot see any other difference between this and normal build...

ciechowoj:
It seems like bug in `dub`. I've overridden `-w` by customizing `unittest` build type and now dub passes both `-w` and `-wi`.
` +    "buildTypes": {
 +        "unittest": {
 +            "buildOptions": ["unittests", "debugMode", "debugInfo", "warnings"]
 +        },
 +    },`

ciechowoj:
It seems you are right, thanks. But now I have another errors.

ciechowoj:
Btw. do you know what "!<", I see it has some connection with NaNs, but is isn't stated clearly in dspec.

ciechowoj:
Well, it works but I'm nos sure if the unit tests are run in preferred way. Currently they are run by `execute` from `Test.d`. However it seems that `dub` creates and executes the `unittest` binary on its own, so maybe I should run it directly from command line. I believe it needs more tinkering...

jacob-carlborg:
> Btw. do you know what "!<" does? I see it has some connection with NaNs, but is isn't stated clearly in dspec.

@ciechowoj you can have a look at the commit that removed those operators in the regular code https://github.com/SiegeLord/Tango-D2/commit/371a66d61fd5f9f324bab593ffd7c9050dec7649

ciechowoj:
I've already issued pull request to Tango : ), it seems they missed one place it that commit...

ciechowoj:
I fixed temporary file creation (added `mkstemp` (`mkstemps` actually) function @jacob-carlborg suggested). Is there anything more that I should fix?

test_files/preprocessor.h
ciechowoj: Actually it should work for more things than integers. I've just added tests for strings.

clang/Cursor.d
```diff
@@ -105,7 +160,38 @@ struct Cursor
...
+            result[cursor.spelling] = cursor.kind;
+
+        return result;
+    }
+
+    bool isPredefined() const
+    {
+        auto xkind = spelling in predefined;
+
+        if (xkind !is null && *xkind == kind)
```
jacob-carlborg: The if statement is not necessary. Just return the condition directly.

clang/Util.d
```diff
@@ -77,3 +78,35 @@ mixin template CX ()
...
+    string name = format("%sXXXXXXXXXXXXXXXX%s\0", prefix, suffix);
+    char[] path = buildPath(tempDir(), name).dup;
+    const size_t termAnd6XSize = 7;
+    randstr(path[$ - name.length + prefix.length .. $ - suffix.length - termAnd6XSize]);
+
+    int fd = mkstemps(path.ptr, cast(int) suffix.length);
+    scope (exit) close(fd);
+
+    if (fd == -1)
+        throw new ErrnoException("Cannot create temporary file.");
```
jacob-carlborg: Minor detail but I'm not a big fan of `ErrnoException`. It's an implementation detail that the error occurred in C.

unit_tests/Common.d
```diff
@@ -0,0 +1,72 @@
...
+}
+
+string translate(TranslationUnit translationUnit) 
+{
+    resetOutput();
+    auto translator = new Translator(translationUnit);
+    return translator.translateToString();
+}
+
+class TranslateAssertError : object.Error
```
jacob-carlborg: Why not inheriting from `AssertError`?

clang/Util.d

clang/Cursor.d
ciechowoj: That was silly. Fixed.

clang/Util.d
```diff
@@ -77,3 +78,43 @@ mixin template CX ()
...
+    string name = format("%sXXXXXXXXXXXXXXXX%s\0", prefix, suffix);
+    char[] path = buildPath(tempDir(), name).dup;
+    const size_t termAnd6XSize = 7;
+    randstr(path[$ - name.length + prefix.length .. $ - suffix.length - termAnd6XSize]);
+
+    int fd = mkstemps(path.ptr, cast(int) suffix.length);
+    scope (exit) close(fd);
+
+    if (fd == -1)
+        throw new NamedTempFileException("Cannot create temporary file.");
```
jacob-carlborg: Please include the full path of the file.

ciechowoj: Done.

jacob-carlborg:
I think the code looks fine now but commits are a bit of a mess. Ideally one commit should be responsible for one thing. Example, do not do changes to the functionally and style changes in the same commit. There should also be no commits that correct a mistake in a previous commit. Perhaps it's easiest to squash all commits.

ciechowoj:
Done.

jacob-carlborg:
Just one last thing, do you mind adding an entry to the changelog. Add the entry below an new heading "Unreleased". You can look at an older version of the chanelog.

ciechowoj:
Done.

jacob-carlborg:
:ok_hand:

jacob-carlborg:
Thanks.

PR #52, opened 2016-02-19
Translate in order

ciechowoj:
Replaces superfluous switch statement from `Translator.translateString`.
Makes translated statements to appear in C order (at least for existing acceptance tests, I believe it needs more test).

### DIFF
--- a/clang/Cursor.d
+++ b/clang/Cursor.d
@@ -6,10 +6,13 @@
  */
 module clang.Cursor;
 
+import std.array : appender, Appender;
+
 import mambo.core._;
 
 import clang.c.Index;
 import clang.Index;
+import clang.File;
 import clang.SourceLocation;
 import clang.SourceRange;
 import clang.Token;
@@ -48,6 +51,11 @@ struct Cursor
     @property SourceLocation location () const
     {
         return SourceLocation(clang_getCursorLocation(cx));
+    }
+
+    @property File file () const
+    {
+        return location.spelling.file;
     }
 
     @property TokenRange tokens() const
@@ -140,6 +148,16 @@ struct Cursor
         return app.data;
     }
 
+    Cursor semanticParent() const
+    {
+        return Cursor(clang_getCursorSemanticParent(cast(CXCursor) cx));
+    }
+
+    Cursor lexicalParent() const
+    {
+        return Cursor(clang_getCursorLexicalParent(cast(CXCursor) cx));
+    }
+
     @property CXLanguageKind language ()
     {
         return clang_getCursorLanguage(cx);
@@ -190,6 +208,87 @@ struct Cursor
     @property Cursor definition() const
     {
         return Cursor(clang_getCursorDefinition(cast(CXCursor) cx));
+    }
+
+    void dumpAST(ref Appender!string result, size_t indent, File* file)
+    {
+        import std.format;
+        import std.array : replicate;
+        import std.algorithm.comparison : min;
+
+        string stripPrefix(string x)
+        {
+            immutable string prefix = "CXCursor_";
+            immutable size_t prefixSize = prefix.length;
+            return x.startsWith(prefix) ? x[prefixSize..$] : x;
+        }
+
+        string prettyTokens(TokenRange tokens, size_t limit = 5)
+        {
+            string prettyToken(Token token)
+            {
+                immutable string prefix = "CXToken_";
+                immutable size_t prefixSize = prefix.length;
+                auto x = to!string(token.kind);
+                return "%s \"%s\"".format(
+                    x.startsWith(prefix) ? x[prefixSize..$] : x,
+                    token.spelling);
+            }
+
+            auto result = appender!string("[");
+
+            if (tokens.length != 0)
+            {
+                result.put(prettyToken(tokens[0]));
+
+                foreach (Token token; tokens[1..min($, limit)])
+                {
+                    result.put(", ");
+                    result.put(prettyToken(token));
+                }
+            }
+
+            if (tokens.length > limit)
+                result.put(", ..]");
+            else
+                result.put("]");
+
+            return result.data;
+        }
+
+        immutable size_t step = 4;
+
+        auto text = "%s \"%s\" [%d..%d] %s\n".format(
+            stripPrefix(to!string(kind)),
+            spelling,
+            extent.start.offset,
+            extent.end.offset,
+            prettyTokens(tokens));
+
+        result.put(" ".replicate(indent));
+        result.put(text);
+
+        if (file)
+        {
+            foreach (cursor, _; all)
+            {
+                if (!cursor.isPredefined() && cursor.file == *file)
+                    cursor.dumpAST(result, indent + step);
+            }
+        }
+        else
+        {
+            foreach (cursor, _; all)
+            {
+                if (!cursor.isPredefined())
+                    cursor.dumpAST(result, indent + step);
+            }
+        }
+    }
+
+    void dumpAST(ref Appender!string result, size_t indent)
+    {
+        dumpAST(result, indent, null);
     }
 }
 
--- a/clang/TranslationUnit.d
+++ b/clang/TranslationUnit.d
@@ -22,7 +22,10 @@ struct TranslationUnit
 {
     mixin CX;
 
-    static TranslationUnit parse (Index index, string sourceFilename, string[] commandLineArgs,
+    static TranslationUnit parse (
+        Index index,
+        string sourceFilename,
+        string[] commandLineArgs,
         CXUnsavedFile[] unsavedFiles = null,
         uint options = CXTranslationUnit_Flags.CXTranslationUnit_DetailedPreprocessingRecord)
     {
@@ -74,6 +77,16 @@ struct TranslationUnit
         return DiagnosticVisitor(cx);
     }
 
+    @property DiagnosticSet diagnosticSet ()
+    {
+        return DiagnosticSet(clang_getDiagnosticSetFromTU(cx));
+    }
+
+    @property size_t numDiagnostics ()
+    {
+        return clang_getNumDiagnostics(cx);
+    }
+
     @property DeclarationVisitor declarations ()
     {
         return DeclarationVisitor(clang_getTranslationUnitCursor(cx));
@@ -84,10 +97,39 @@ struct TranslationUnit
         return File(clang_getFile(cx, filename.toStringz));
     }
 
+    @property string spelling ()
+    {
+        return toD(clang_getTranslationUnitSpelling(cx));
+    }
+
+    File file ()
+    {
+        return file(spelling);
+    }
+
     @property Cursor cursor ()
     {
         auto r = clang_getTranslationUnitCursor(cx);
         return Cursor(r);
+    }
+
+    string dumpAST (bool skipIncluded = false)
+    {
+        import std.array : appender;
+
+        auto result = appender!string();
+
+        if (skipIncluded)
+        {
+            File file = this.file;
+            cursor.dumpAST(result, 0, &file);
+        }
+        else
+        {
+            cursor.dumpAST(result, 0);
+        }
+
+        return result.data;
     }
 }
 
--- a/dstep/translator/CodeBlock.d
+++ b/dstep/translator/CodeBlock.d
@@ -1,0 +1,160 @@
+/**
+ * Copyright: Copyright (c) 2016 Wojciech Szęszoł. All rights reserved.
+ * Authors: Wojciech Szęszoł
+ * Version: Initial created: Mar 11, 2016
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+
+module dstep.translator.CodeBlock;
+
+import std.array;
+import std.string;
+
+enum EndlHint
+{
+    /// Does nothing, useful for cases when the translator doesn't produce any ouput.
+    empty,
+
+    /// Special node for grouping, puts an extra new line after grouped items, should not have spelling.
+    group,
+
+    /// Single line item, should not have children.
+    singleLine,
+
+    /// Multiline item, the children are indented.
+    multiLine,
+
+    /// Puts braces if children takes more than single line, useful e.g for loops.
+    subscopeWeak,
+
+    /// Always puts braces, useful e.g. for classes.
+    subscopeStrong,
+}
+
+struct CodeBlock
+{
+    string spelling;
+    EndlHint endlHint;
+    CodeBlock[] children;
+
+    this(string spelling, EndlHint endlHint, CodeBlock[] children = [])
+    {
+        this.spelling = spelling;
+        this.endlHint = endlHint;
+        this.children = children;
+    }
+
+    this(EndlHint endlHint, CodeBlock[] children = [])
+    {
+        this.spelling = "";
+        this.endlHint = endlHint;
+        this.children = children;
+    }
+
+    this(string spelling)
+    {
+        this.spelling = spelling;
+        this.endlHint = EndlHint.singleLine;
+        this.children = [];
+    }
+
+    this(CodeBlock[] children)
+    {
+        this.spelling = null;
+        this.endlHint = EndlHint.group;
+        this.children = children;
+    }
+}
+
+size_t countLines(in CodeBlock block)
+{
+    // this desires to be implemented in more efficient way
+    return compose(block).count("\n");
+}
+
+size_t countLines(in CodeBlock[] blocks)
+{
+    // this desires to be implemented in more efficient way
+    auto output = appender!string();
+    compose(output, blocks, 0);
+    return output.data.count("\n");
+}
+
+void compose(ref Appender!string output, in CodeBlock[] blocks, size_t indent)
+{
+    bool isSubscope(EndlHint hint)
+    {
+        return hint == EndlHint.subscopeWeak || hint == EndlHint.subscopeStrong;
+    }
+
+    if (!blocks.empty)
+    {
+        compose(output, blocks[0], indent);
+
+        foreach (i; 1 .. blocks.length)
+        {
+            if (isSubscope(blocks[i - 1].endlHint)
+                || (isSubscope(blocks[i].endlHint) && blocks[i - 1].endlHint != EndlHint.empty))
+                output.put("\n");
+
+            compose(output, blocks[i], indent);
+        }
+    }
+}
+
+void compose(ref Appender!string output, in CodeBlock block, size_t indent)
+{
+    immutable string tab = "    ";
+
+    output.put(tab.replicate(indent));
+    output.put(block.spelling);
+
+    final switch (block.endlHint)
+    {
+        case EndlHint.empty:
+            break;
+
+        case EndlHint.group:
+            compose(output, block.children, indent);
+
+            if (!output.data.endsWith("\n\n"))
+                output.put("\n");
+
+            break;
+
+        case EndlHint.singleLine:
+            output.put("\n");
+            break;
+
+        case EndlHint.multiLine:
+            output.put("\n");
+            compose(output, block.children, indent + 1);
+            break;
+
+        case EndlHint.subscopeWeak:
+            if (countLines(block.children) == 1)
+            {
+                output.put("\n");
+                compose(output, block.children, indent + 1);
+                break;
+            }
+
+        case EndlHint.subscopeStrong:
+            output.put("\n");
+            output.put(tab.replicate(indent));
+            output.put("{\n");
+            compose(output, block.children, indent + 1);
+            output.put(tab.replicate(indent));
+            output.put("}\n");
+            break;
+    }
+}
+
+string compose(in CodeBlock block)
+{
+    auto output = appender!string();
+
+    compose(output, block, 0);
+
+    return output.data;
+}
--- a/dstep/translator/Translator.d
+++ b/dstep/translator/Translator.d
@@ -7,6 +7,7 @@
 module dstep.translator.Translator;
 
 import std.file;
+import std.array;
 
 import mambo.core._;
 
@@ -17,6 +18,7 @@ import clang.TranslationUnit;
 import clang.Type;
 import clang.Util;
 
+import dstep.translator.CodeBlock;
 import dstep.translator.Declaration;
 import dstep.translator.Enum;
 import dstep.translator.IncludeHandler;
@@ -26,7 +28,7 @@ import dstep.translator.Output;
 import dstep.translator.Record;
 import dstep.translator.Type;
 
-private static string[Cursor] anonymousNames;
+static string[Cursor] anonymousNames;
 
 class Translator
 {
@@ -44,7 +46,7 @@ class Translator
         string inputFilename;
         File inputFile;
         Language language;
-        string[string] deferredDeclarations;
+        CodeBlock[string] deferredDeclarations;
     }
 
     this (string inputFilename, TranslationUnit translationUnit, const Options options = Options.init)
@@ -59,9 +61,12 @@ class Translator
 
     this (TranslationUnit translationUnit, const Options options = Options.init)
     {
+        this.inputFilename = translationUnit.spelling;
         this.translationUnit = translationUnit;
         outputFile = options.outputFile;
         language = options.language;
+
+        inputFile = translationUnit.file(inputFilename);
     }
 
     void translate ()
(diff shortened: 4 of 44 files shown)


jacob-carlborg:
I'm going to wait with reviewing this until the #51is merged.

ciechowoj:
Sure : ). Just tell me what needs to be fixed in #51.

ciechowoj:
Btw. It seems that Objective-C tests doesn't run on linux. Is there a way to enable them there?

jacob-carlborg:
You need GNUStep at least. Not sure if anything else is needed.

ciechowoj:
I updated the code.

dstep/translator/Translator.d
```diff
@@ -69,48 +70,62 @@ class Translator
...
+
+        size_t aTrailing = countTrailingEndls(a.data);
+        size_t bLeading = countLeadingEndls(b);
+        size_t bTrailing = countTrailingEndls(b);
+
+        size_t num = min(bLeading + aTrailing, 2);
+        a.put("\n".replicate(num < aTrailing ? 0 : num - aTrailing));
+        a.put(b[bLeading..$-bTrailing]);
+        a.put("\n".replicate(min(bTrailing, 2)));
+    }
```
jacob-carlborg: I'm not sure what the behavior is of these three methods.

dstep/translator/Translator.d
```diff
@@ -120,15 +135,15 @@ class Translator
             switch (cursor.kind)
             {
                 case CXCursor_ObjCInterfaceDecl:
-                    return (new ObjcInterface!(ClassData)(cursor, parent, this)).translate();
+                    return (new ObjcInterface!(ClassData)(cursor, parent, this)).translate() ~ "\n";
```
jacob-carlborg: Is it possible to handle the newlines in central location?

jacob-carlborg:
I'm pretty sure that the `objc/methods` test needs to be updated to match the new order.

dstep/translator/Translator.d
ciechowoj: Yes, the `endlAwareAppend` function is supposed to do that.

dstep/translator/Translator.d
ciechowoj: `countTrailingEndls` and `countLeadingEndls` returns the number of `\n` at the end and the begin of given string respectively.

ciechowoj: Then `endlAwareAppend` tries to join two strings together (`Appender` and `string` actually) in the way there is never more than two `\n` between them. It should collaborate with "translate" methods. 

Translate method can return different combinations of end-lines e.g.:

- `content\n`
- `\ncontent\n`
- `content\n\n`

Signalling that the item should be followed by `\n`, preceded by `\n` and  and followed by `\n` and followed by double `\n` respectively. Then `endlAwareAppend` can intelligently merge two items by taking the sum of `\n` but not more than 2.

For example, let `++` be `endlAwareAppend` equivalent:

- `"\ncontent\n" ++ "\ncontent\n" === "\ncontent\n\ncontent\n"`
- `"enum FOO = 1;\n" ++ "enum BAR = 2;\n" === "enum FOO = 1;\nenum BAR = 2;\n"`
- `"enum FOO = 1;\n" ++ "\nclass A { }; \n" === "enum FOO = 1;\n\nclass A { };\n"`
- `"enum WANT_BLANK_AFTER_THIS = 0;\n\n" ++ "\nclass WANT_BLANK_BEFORE_AND_AFTER_THIS { }\n\n" === "enum WANT_BLANK_AFTER_THIS = 0;\n\nclass WANT_BLANK_BEFORE_AND_AFTER_THIS { }\n\n"`

Saying it another way you precede item by `\n` if you want blank before the item and follow the item by double `\n` if you want a blank newline after item and the `endlAwareAppend` takes care of joining it so that there is no more than one blank line.

ciechowoj: Maybe a good idea would be to add some unit tests for this function to show the behavior.

ciechowoj:
Hmm, it seems that the Objective-C methods are still handled in the wrong order...

dstep/translator/Translator.d
jacob-carlborg: Yes please.

ciechowoj:
I added the unit tests for `endlAwareAppend`. Making that ObjC methods to appear in order is more tricky. Btw. when I tries to run ObjC test on Linux I need a `limits.h` header. Where did you took the headers in "resources" from? Didn't you considered to add full standard C library? Maybe I could add the headers from "clang's" `libc++`? They are on MIT Licence.

jacob-carlborg:
The headers in `resources` are Clang's internal headers. I've added them on a case by case basis. I don't think it would be a good idea to add the C standard library. I don't know how the implementation looks like in regards to different platforms. There are at least three different standard libraries as well.

It's assumed the C standard library is always in the library search path. That's not the case with the internal headers. Those are referenced relative to the compiler.

ciechowoj:
> The headers in resources are Clang's internal headers.

Actually they aren't. They are part of standard C library. [https://en.wikipedia.org/wiki/C_standard_library](Here) you have the list of all standard headers.

ciechowoj:
Here my question, it would be the best to do it one way or another. Depend on system installed library (little problematic on Windows..., it would need some user attention) or put the whole library there.

ciechowoj:
Hmm, however there is something strange happening some of standard headers are on the libclang search path and some aren't. The files you attached aren't and `limits.h` isn't too.

jacob-carlborg:
They are built-in headers: http://clang.llvm.org/docs/FAQ.html#id3.

ciechowoj:
Oh, I see.

ciechowoj:
Well, I fixed the order of methods in ObjectiveC class.

ciechowoj:
Side question:
Is there any particular reason that the module names contains the class names? E.g. `module clang.c.CXErrorCode;` for `CXErrorCode` or `module clang.TranslationUnit` for `TranslationUnit`.

jacob-carlborg:
Not sure that I understand the question. The files in `clang.c` is a direct mapping of the C header files.

clang/Cursor.d
```diff
@@ -191,6 +209,87 @@ struct Cursor
     {
         return Cursor(clang_getCursorDefinition(cast(CXCursor) cx));
     }
+
+    void dumpAST(ref Appender!string result, size_t indent, File* file)
```
jacob-carlborg: Is this some kind of debug function?

unit_tests/Common.d
```diff
@@ -66,10 +85,134 @@ C code translated to:
...
+    string objCPath,
+    bool strict = true,
+    string file = __FILE__,
+    size_t line = __LINE__)
+{
+    import std.system;
+
+    string[] arguments = ["-ObjC", "-Iresources"];
+
+    if (os == OS.linux)
```
jacob-carlborg: `version(linux)`?

clang/Cursor.d
ciechowoj: Yes.

unit_tests/Common.d
```diff
@@ -66,10 +85,134 @@ C code translated to:
...
+        import std.string;
+
+        string start = "#include <...> search starts here:";
+        string stop = "End of search list.";
+        auto paths = output.findSplitAfter(start)[1].findSplitBefore(stop)[0].strip();
+        return paths.empty ? null : map!(a => "-I%s".format(a))(paths.splitLines()).array;
+    }
+
+    import std.process : executeShell;
+    auto result = executeShell("gcc -E -v - < /dev/null");
```
jacob-carlborg: Perhaps `gcc` should be `cc` or something more general to cover Clang as well.

ciechowoj:
> Not sure that I understand the question. The files in clang.c is a direct mapping of the C header files.

I mean every time I want to use let say `clang.Cursor.Cursor` I need to import `clang.Cursor`. Why don;t put all that stuff to `clang` module and save a trouble of making all that imports.
What is more sometimes it is desirable for the classes to access private members of others and it doesn't seem to be possible across the module boundary.

jacob-carlborg:
> I mean every time I want to use let say clang.Cursor.Cursor I need to import clang.Cursor. Why don;t put all that stuff to clang module and save a trouble of making all that imports.

I like to keep roughly a one to one mapping of types and modules, something like the single responsibility principle. I think Cursor, Type and Visitor are big enough modules as they are. To minimize the need for imports it's possible to create a module called `package` which publicly imports all modules in the package, then it's possible to import the package directly, i.e. `import clang`.

> What is more sometimes it is desirable for the classes to access private members of others and it doesn't seem to be possible across the module boundary.

That's what the `package` attribute is for.

unit_tests/Common.d

unit_tests/Common.d
ciechowoj: Right. Fixed. I wandered about the standard preprocessor `cpp` but I'm not sure how it interacts with `clang`.

dstep/translator/Translator.d
jacob-carlborg: I still think there is be a better way to do this. I'll have to think about this. I have solved it in one of my other projects, I just have to figure out if the same technique can be used in this project as well.

unit_tests/Common.d
```diff
@@ -66,10 +85,132 @@ C code translated to:
...
+        import std.string;
+
+        string start = "#include <...> search starts here:";
+        string stop = "End of search list.";
+        auto paths = output.findSplitAfter(start)[1].findSplitBefore(stop)[0].strip();
+        return paths.empty ? null : map!(a => "-I%s".format(a))(paths.splitLines()).array;
+    }
+
+    import std.process : executeShell;
+    auto result = executeShell("cc -E -v - < /dev/null");
```
jacob-carlborg: Aren't the directories returned by this always searched? Why is it necessary to explicitly add them?

dstep/translator/Translator.d
ciechowoj: Ok, I'll think about alternatives too.

unit_tests/Common.d
ciechowoj: The situation with the standard includes is real mess : (. If I do not add that search directories I get complaints like:
` /usr/include/GNUstep/GNUstepBase/GSConfig.h:376:10: fatal error: 'stdbool.h' file not found`

And even when I added that files to `resources` I began to get complaints about headers from `objc` subdirectory (seems objc runtime e.g.`objc/objc.h`) which I could find only in this location on my system:
`/usr/lib/gcc/x86_64-linux-gnu/4.8/include/objc/objc.h`
Despite the fact I have the standalone clang installation.

ciechowoj:
Good to know : ) . Thanks.

dstep/translator/Translator.d
ciechowoj: So, maybe instead of returning strings from translation subroutines we could return some abstract type like `Node`:

    struct Node
    {
        string spelling;
        bool endlHint; // tells if node wants space around itself
        Node[] children;
    }

Then during the translation process we would create a tree of such a nodes and then there could be a separate routine that takes a root of such a tree and efficiently produces output string. We could even handle making indentations there.

`endlHint` could be extended to something more than `bool` e.g. `enum`:

    enum EndlHint
    {
        Inline, // no newline at all
        NewLine, // should occupy separate line
        Group, // should has blank line before and after
        ...
    }

jacob-carlborg: They way I've implemented this in my [other project](https://github.com/d-widget-toolkit/jport/blob/dev/src/main/scala/dwt/jport/writers/NewlineWriter.scala) is that each node in the AST has a method that indicates if the node itself is multi-line or not. This is basically hard coded based on the type of the node. I also have information if there's a node coming after the current node at the same level (a next node). With this information it's trivial to handle the newlines: 

1. Add a newline after each node
2. If there's no next node we're done
3. If the current node or the next node is a multi-line node, add a newline

I also check the line numbers of the nodes to be able to group nodes in the output, but I don't think that's necessary for DStep, at least not for the first step.

I'll read through your comment tomorrow.

ciechowoj: Seems reasonable, but I see some issues.

> each node in the AST
So we would need to have some dstep AST (currently there are strings only) or use the clang AST but there may be no one to one mapping between dstep AST and clang AST.

> Add a newline after each node
What if there is a few nodes in one line?

> I also have information if there's a node coming after the current node at the same level (a next node).
I suppose it is the easiest to do if a full AST is known in advance.

jacob-carlborg: > So we would need to have some dstep AST (currently there are strings only) or use the clang AST but there may be no one to one mapping between dstep AST and clang AST.

In the case of DStep I imagine a function taking a cursor or a type which then does a if/switch and hard codes if a cursor type is multiline or not.

> What if there is a few nodes in one line?

This functionality is only for statements and other declarations that will end with a newline. For example, It's not used when translating the expressions in for loop. I basically have a "writer" for each lager component. Then a "translator" for something smaller like an expression. This functionality is only used between the "writers".

> I suppose it is the easiest to do if a full AST is known in advance.

Yes. But the case of my other project the library I'm using, which contains the AST, is using the visitor pattern. Then I'm visiting all nodes at a given level building up an array. Then I'm visiting all nodes in the array to do the actual processing. When the data structure is an array it's a lot easier to get the next and previous nodes.

This was all just an idea because I though the implementation was cleaner in my other project. But I have not looked into if that approach works in DStep.

jacob-carlborg: Your approach of a `Node` does sound like it could work.

ciechowoj: So I'll try to implement the `Node` approach and see how it works. One could think about the `Node` as very simple `AST`, so there seems to be some parallels with your approach.

jacob-carlborg: Sounds good.

ciechowoj:
Added `CodeBlock` class and bunch of unit tests for it. The code still needs to be changed to make use of it.

ciechowoj:
Well, I changed the code. It uses the `CodeBlock` for formatting.

unit_tests/Common.d
```diff
@@ -28,36 +29,76 @@ public import clang.Token;
...
+    size_t line = __LINE__)
 {
     import std.format : format;
+    import std.algorithm : map;
+
+    auto sep = "----------------";
+
+    if (unit.numDiagnostics != 0)
+    {
+        auto diagnostics = map!(a => a.toString())(unit.diagnosticSet);
```
jacob-carlborg: In general I prefer to use UFCS for all algorithm functions, like `map`.

clang/TranslationUnit.d
```diff
@@ -74,6 +77,16 @@ struct TranslationUnit
         return DiagnosticVisitor(cx);
     }
 
+    @property DiagnosticSet diagnosticSet ()
+    {
+        return DiagnosticSet(clang_getDiagnosticSetFromTU(cx));
+    }
+
+    @property size_t numDiagnostics ()
```
jacob-carlborg: Is this used because it's more efficient than creating a `DiagnosticSet` and call `length`?

dstep/translator/CodeBlock.d
```diff
@@ -0,0 +1,149 @@
...
+import std.string;
+
+enum EndlHint
+{
+    Empty, // does nothing
+    Group,
+    SingleLine,
+    MultiLine, // indent children
+    SubscopeWeak, // put braces if children takes more than single line
+    SubscopeStrong, // always put braces
```
jacob-carlborg: Since you already have comments on most members here, why not make it proper ddoc comments and put it on all members.

dstep/translator/CodeBlock.d
```diff
@@ -0,0 +1,149 @@
...
+        this.children = [];   
+    }
+
+    this(CodeBlock[] children)
+    {
+        this.spelling = null;
+        this.endlHint = EndlHint.Group;
+        this.children = children;
+    }
+}
```
jacob-carlborg: Are all these constructors necessary? By default, for all structs in D, it's possible to initialize any members. Either by position: `CodeBlock("spelling")`, `CodeBlock("spelling", EndlHint.Group")` or by using the names of the fields: `CodeBlock c = { endlHint: EndlHint.Group }`, `CodeBlock c = { children: [], endlHint: EndlHint.Group }`. The second syntax is only allowed in assignments. See the documentation for more information [1] and [2].

[1] https://dlang.org/spec/struct.html#static_struct_init
[2] https://dlang.org/spec/struct.html#struct-literal

dstep/translator/CodeBlock.d
```diff
@@ -0,0 +1,149 @@
...
+    bool isSubscope(EndlHint hint)
+    {
+        return hint == EndlHint.SubscopeWeak || hint == EndlHint.SubscopeStrong;
+    }
+
+    if (!blocks.empty)
+    {
+        compose(output, blocks[0], indent);
+     
+        for (size_t i = 1; i < blocks.length; ++i)
```
jacob-carlborg: I recommend a [`foreach` loop](https://dlang.org/spec/statement.html#ForeachRangeStatement) instead.

test_files/aggregate.h
```diff
@@ -0,0 +1,23 @@
...
+struct Bar
+{
+	int x;
+	int y;
+};
+
+const char* q;
+
+int (*e) (int a, int b);
+
```
jacob-carlborg: Why is there extra newlines in the beginning and end of the file?

unit_tests/CodeBlockTests.d
```diff
@@ -0,0 +1,169 @@
...
+    auto declFloat = CodeBlock("float b;", EndlHint.SingleLine, []);
+    auto group1 = CodeBlock("", EndlHint.Group, [declInt, declFloat]);
+
+    assertEq("int a;\nfloat b;\n\n", compose(group1));
+
+    auto declString = CodeBlock("string s;", EndlHint.SingleLine, []);
+    auto declFuncDecl = CodeBlock("void func();", EndlHint.SingleLine, []);
+    auto group2 = CodeBlock("", EndlHint.Group, [declString, declFuncDecl]);
+    auto group3 = CodeBlock("", EndlHint.Group, [group1, group2]);
+    assertEq(q"EOS
```
jacob-carlborg: When using delimited strings I prefer that the identifier is matching the language in the code is in. Some editors might highlight that.

unit_tests/FileTests.d
```diff
@@ -0,0 +1,139 @@
...
+        "test_files/typedef_struct.d",
+        "test_files/typedef_struct.h");
+}
+
+unittest
+{
+    assertTranslatesCFile(
+        "test_files/unions.d",
+        "test_files/unions.h");
+}
```
jacob-carlborg: Is this instead of the cucumber tests?

unit_tests/UnitTests.d
```diff
@@ -22,9 +24,36 @@ unittest
...
 }
 
-unittest
+unittest 
 {
-	assertTranslates(
+    assertTranslates(
         "#define SOME_STRING \"foobar\"", 
-        "extern (C):\n\nenum SOME_STRING = \"foobar\";");	
+        "extern (C):\n\nenum SOME_STRING = \"foobar\";");
```
jacob-carlborg: Looks to still be some trailing whitespace.

unit_tests/UnitTests.d
```diff
@@ -22,9 +24,36 @@ unittest
         "extern (C):\n\nenum FOO = 0;\nenum BAR = 1;");
 }
 
-unittest
+unittest
```
jacob-carlborg: Trailing whitespace added.

clang/TranslationUnit.d
ciechowoj: Truth say, I didn't make measurements, but I believe so. It maps to one libclang API call.

jacob-carlborg: That's good enough for more. I was mostly curious.

unit_tests/CodeBlockTests.d
ciechowoj: Could you give me an example? I've never seen that 'idiom' before. Should I replace `EOS` with something like `C` ?

unit_tests/FileTests.d
ciechowoj: No. The cucumber test are still enabled. But the problem with them is they give unreadable error messages - hard to see what is wrong at first glance (at second and third too...). So I implemented some custom assertions that show better information (e.g. clangs diagnostic messages if the code doesn't compile). I was thinking about completely replacing cucumber. What do you think about this?

unit_tests/CodeBlockTests.d
jacob-carlborg: Yes, or `D` if it's for D. I have not seen that idiom used in D (I've barely seen these type of string literals in D at all) but it's more common in Ruby. I plan to add support for that in TextMate. It's already supported in TextMate for Ruby and it's quite common for Markdown as well, as you can see here on GitHub:

```d
string a = "a block of code highlighted as D";
```

unit_tests/FileTests.d
jacob-carlborg: > I was thinking about completely replacing cucumber. What do you think about this?

Sure, go ahead, remove them. The tests that are Cucumber tests now are kind of pointless to have as Cucumber tests. But I know @Dicebot likes them :wink:.

dstep/translator/CodeBlock.d
ciechowoj: What about replacing them with explicitly named static methods? E.g.

    static CodeBlock empty();
    static CodeBlock singleLine(string spelling);

?

unit_tests/UnitTests.d
ciechowoj: Fixed.

unit_tests/UnitTests.d
ciechowoj: Fixed.

unit_tests/Common.d

dstep/translator/CodeBlock.d
ciechowoj: Fixed.

dstep/translator/CodeBlock.d
ciechowoj: Much shorter now : )

test_files/aggregate.h
ciechowoj: No particular reason - fixed.

dstep/translator/CodeBlock.d
jacob-carlborg: Doesn't matter that much. I was just thinking since D has so flexible struct initialization syntax.

dstep/translator/CodeBlock.d
```diff
@@ -0,0 +1,160 @@
...
+    SingleLine,
+
+    /// Multiline item, the children are indented.
+    MultiLine,
+
+    /// Puts braces if children takes more than single line, useful e.g for loops.
+    SubscopeWeak,
+
+    /// Always puts braces, useful e.g. for classes.
+    SubscopeStrong,
```
jacob-carlborg: Just remembered, the style convention in D is to start enum members with a lower case letter.

dstep/translator/CodeBlock.d
ciechowoj: So, I would leave it as is, at least for now.

unit_tests/CodeBlockTests.d
ciechowoj: Fixed (not sure if correctly).

dstep/translator/CodeBlock.d
ciechowoj: Fixed.

unit_tests/FileTests.d
ciechowoj: So maybe I will leave the cucumber for a bit more. Currently those unit test tests only `Translator` class, they do not invoke the dstep directly from outside (from the command line). So, maybe it would be better to have the rest of code covered with something.

Dicebot: Yay, please kill Cucumber! <3

jacob-carlborg: > they do not invoke the dstep directly from outside (from the command line)

That's a good point. So perhaps leave one Cucumber test for now?

ciechowoj: Done.

jacob-carlborg: Sorry, but I think one Objective-C Cucumber test should be left as well. It requires different flags.

ciechowoj: Good point, I'll fix it in a moment.

ciechowoj: Done.

jacob-carlborg:
Please squash the commits. But leave "Removes some of cucumber tests." as a separate commit.

ciechowoj:
I squashed the commits.
Unrelated:
I'm writing the GSOC proposal. Do you have any particular features that would be worth to implement in dstep? Do you think I could ask on the forum about it : ) ?

jacob-carlborg:
:ok_hand:

jacob-carlborg:
> I'm writing the GSOC proposal. Do you have any particular features that would be worth to implement in dstep? Do you think I could ask on the forum about it : ) ?

There's the [list](http://wiki.dlang.org/GSOC_2016_Ideas#C_to_D_conversion_tool_.28DStep.29) on the wiki page. The [todo list](https://github.com/jacob-carlborg/dstep/blob/master/todo.taskpaper) and the GitHub issues.

ciechowoj:
Yes, but it does't seem much for 3 months : D

ciechowoj:
Or maybe I'm underestimating.

jacob-carlborg:
* Extended support for the preprocessor
* C++ support

That should keep you busy for a while :wink:.

jacob-carlborg:
Oh, I forgot about the changelog. Could you please create a PR to update the changelog.

ciechowoj:
Ok.

jacob-carlborg:
BTW, I think the C++ support is at the top of the wish list if you ask in the forums.

ciechowoj:
I supposed so, but I would like to focus on polishing the C. C++ if there will be time.

ciechowoj:
Yes, I was considering that. But I believe it should be added as additional option the can be enabled in command line arguments. What do you mean by CLI arguments exactly?

ciechowoj:
I actually tried something like this last summer holidays. I used pycparser to parse libclang headers and wrote some code that was supposed to create python bindings in some smart way (e.g. by collapsing that pointer/size arguments or generating destructors (__del__) methods that invoke dispose_ functions).

jacob-carlborg:
> I supposed so, but I would like to focus on polishing the C. C++ if there will be time.

You just asked about more things to do because you thought it wouldn't be enough for three months, that's why I mentioned C++. But I would prefer polishing translating of C.

ciechowoj:
> You just asked about more things to do because you thought it wouldn't be enough for three months, that's why I mentioned C++. But I would prefer polishing translating of C.
Yes, I see : ).

ciechowoj:
> Something as simple dstep --deduce-enum-namespaces --remove-symbol-prefix to enable smarter conversion. My point was that you don't normally want to do stuff like that by default as there is small risk of generating broken bindings. But for any specific library it should be possible to pick set of CLI flags for smarter conversion that work good enough on case by case basis.

Right, I've just had troubles with decoding *CLI* (Command LIne) : D. But, yes such a feature would be very useful. Obviously there would be problem with functions as their name cannot be changes, dstep would need to generate second wrapper version of function with nicer name and interface which just call the first true C binding.

ciechowoj:
> Actually they can be changed. You can emit pragma(mangle, "original_name") to ensure it links properly. I am not sure if this is a good idea though, feels a bit dirty.

Oh, nice. Didn't know about that.

jacob-carlborg:
> Probably separating generated results in "raw" binding and wrapper is indeed better.

I agree. How about we focus on making the produced bindings as good as possible, i.e. minimizing the need for post processing, before doing more fancy stuff like wrappers?

ciechowoj:
Yes, sure, first things first.

Issue #47, opened 2016-01-19
Treatment of #define


russel:
From what I can see #define is removed but the values of them are used. So the C code:

#define LENGTH 4
char a[LENGTH];

becomes the D code:

char[4] a;

Might it be better to be more careful with these values and create an immutable variable to replace the #define value as would be done in C++.

immutable LENGTH = 4;
char[LENGTH] a;

jacob-carlborg:
DStep doesn't currently handle anything from the preprocessor at all.

russel:
So how does dstep get the 4 for the "char[4] a" if it hasn't at least remembered the value of LENGTH in some way?

UplinkCoder:
dstep does not. 
clang does run the perprocessor.

jacob-carlborg:
> So how does dstep get the 4 for the "char[4] a" if it hasn't at least remembered the value of LENGTH in some way?

It just calls the libclang function [clang_getArraySize](https://github.com/jacob-carlborg/dstep/blob/master/clang/Type.d#L169).

Dicebot:
@russel it is important to remember that in C pre-processing and compiling are done in two completely separate steps. By the time source code gets to compiler (and thus to dstep which is built on top of libclang) it already looks like `char[4] a` with all macros expanded.

To do what you want one needs to either plug into preprocessor and pass some state to compiler parser from it or write your own C pre-processor which replaces numeric defines with `enum` instead of doing actual text substitution. Or something similar.

russel:
@Dicebot  Ah, OK, crucial bit of information: DStep is working on the abstract syntax tree form of the code not the source itself. Sounds like a pre-pre-processor which converted #defines into either inline functions or constexpr values would be needed to solve the problem I have here.

jacob-carlborg:
@Dicebot Clang retains the full source information, that's the whole point of Clang. They tried create tools based GCC but it was difficult/impossible because it did not retain the complete source information. For example, the parser did some forms of const folding. Something like `const int a = 3 + 1;` looked like `const int a = 4;` in the AST.

The necessary information should be available in Clang. When I started with DStep libclang did not have an API to interact with the preprocessor. It still doesn't, but I recently noticed that they provide a translation unit flag [CXTranslationUnit_DetailedPreprocessingRecord](http://clang.llvm.org/doxygen/group__CINDEX__TRANSLATION__UNIT.html#gab1e4965c1ebe8e41d71e90203a723fe9) that seems to be interesting.

PR #54, opened 2016-03-10
[WIP] Prototype of support for macro constants.

ciechowoj:
I made a prototype implementation of macro constants. There is more work to be done, but it's proof of concept. Current code handles situations like in #47 .

    #define LENGTH 4
    char a[LENGTH];

    extern (C):

    enum LENGTH = 4;
    extern __gshared char[LENGTH] a;

The code is based on my previous pull request.


jacob-carlborg:
Let me see if I get this correctly. The first pull request you did for defines was to translate code like this: `#define LENGTH 4`. And this pull request if for translating the second line, `char a[LENGTH];`, but keeping the constant instead of the actual value in the output?

ciechowoj:
Yes. But currently this pull request is only prototype/proof of concept. I would be grateful if you could take a look at code and share you opinion about it if you have some time. It works in very simple case and there is plenty to do before it actually can be merged with master...

And by:

> The code is based on my previous pull request.

I meant that this pull request contains a commits from the other pull request that isn't currently merged (translate in order). But, your statement is correct, because you need to translate constant first before you can use them in expressions/statements (like array declaration) : ).

PR #56, opened 2016-03-16
Removes global Output.

ciechowoj:
Removes the `output` global object.


dstep/translator/IncludeHandler.d
```diff
@@ -126,8 +130,8 @@ class IncludeHandler
 
         auto imps = imports.map!(e => toImport(e));
 
-        return r.append(imps).filter!(e => e.any).unique.toArray;
-    }
+        return CodeBlock(r.append(imps).filter!(e => e.any).unique.map!(a => CodeBlock(a)).array);
+    }
```
jacob-carlborg: Trailing whitespace.

dstep/translator/Translator.d
```diff
@@ -222,14 +231,15 @@ class Translator
...
-    string typedef_ (Cursor cursor, String context = output)
+    CodeBlock typedef_ (Cursor cursor)
     {
-        context ~= "alias ";
-        context ~= translateType(cursor.type.canonicalType);
-        context ~= " " ~ cursor.spelling;
-        context ~= ";";
+        import std.format : format;
+
+        string result = "alias %s %s;".format(
```
jacob-carlborg: Please don't use UFCS for `format`.

dstep/translator/IncludeHandler.d

dstep/translator/Translator.d
ciechowoj: Fixed.

jacob-carlborg:
This doesn't just remove the global `output` object but also the concept of a single buffer. As far as I know it's more efficient to write to a single buffer than creating a lot of small strings that are concatenated together.

ciechowoj:
> This doesn't just remove the global output object but also the concept of a single buffer. As far as I know it's more efficient to write to a single buffer than creating a lot of small strings that are concatenated together.

Yes you are generally right. However there is no more strings created than it was before, the only difference is that they are stored in the memory for a longer time, because there is that intermediate `CodeBlock`-tree representation. Then in `compose` function they are concatenated one after another, so the performance is still O(n).

ciechowoj:
The change you are talking of was mostly made in `translate_in_order` patch. It was consequence of introducing that intermediate `CodeBlock`. The `CodeBlock` approach is somewhat simpler, but it wastes memory. I may rewrite it bask so that it doesn't waste memory. The idea is to keep the `Output` class and make more use of that `in` idiom used in one place. So, instead of writing this:

    CodeBlock translateClass(arguments...) 
    {
        CodeBlock result;

        [...]

        result.children ~= translateChild1(args...);
        result.children ~= translateChild2(args...);
        result.children ~= translateChild3(args...);

        [...]

        return result;
    }

Write it more like previously:

    void translateClass(Output output, arguments...) 
    {
        [...]

        output.subscopeString in {
            translateChild1(output, args...);
            translateChild2(output, args...);
            translateChild3(output, args...);
        };

        [...]
    }

    void translateChild1(Output ouput, arguments...)
    {
        output.singleLine("yada yada yada...");
    }

    class Output 
    {

        void singleLine(arguments...);
        void multiLine(arguments...);

        [etc...]

        string finalize(); // maybe data()
    public:
        Appender!string data;

        [some other state]
    }

What do you think.

jacob-carlborg:
I think you can continue to use `format`. There's an overload that takes a buffer as the first argument, if I recall correctly.

jacob-carlborg:
Close enough :smiley:.

ciechowoj:
> This doesn't just remove the global output object but also the concept of a single buffer. As far as I know it's more efficient to write to a single buffer than creating a lot of small strings that are concatenated together.

> I think you can continue to use format. There's an overload that takes a buffer as the first argument, if I recall correctly.

Hmm, so I'm misled now what the problem is : ). Is it that I used `format` too much instead of string concatenation? Or what?

ciechowoj:
> When you append to same string (or use formattedWrite on it) it only causes new allocations if string capacity is exceeded. If bunch of small strings get created separately, each will result in allocation of new GC root.

Right, I get it. 

So, is there anything more I should fix? As I said in above, to use one buffer I would need to make changes I described there.

jacob-carlborg:
> So, is there anything more I should fix

I don't think so. Although I 'm not sure if I understand the code in your comment [above](https://github.com/jacob-carlborg/dstep/pull/56#issuecomment-198674024).

ciechowoj:
TL;DR

The listings presents how interfaces of functions/methods that do translation of particular statements would look like depending how we do formatting (on the go/at the end).

Longer version:

There are (at least) two different ways to do the translation:

1. Translate everything to tree of `CodeBlock`s and concatenate to one string at the end.
2. Append translated entities to single buffer as fast as possible during translation.

Before I did any changes there was a some mix of 1 and 2. Currently we have 1 (mostly). There is possibility to implement it the "2." way. Depending on that choice the routines that translate statements would have different interfaces.1. Simpler to implement.
2.

The first listing in my post presents the function that uses `CodeBlock`, the second listing presents one that takes the reference to buffer to which it is supposed to append the result.

If there is intermediate representation there is no need to handle formatting during translation. However if we want to do formatting as we go we need to somehow put newlines and indentations and one way to implement this is to use that trick with `in`.

> This doesn't just remove the global output object but also the concept of a single buffer. As far as I know it's more efficient to write to a single buffer than creating a lot of small strings that are concatenated together.

I assumed the problem is I'm not doing the translation on-line and I'm concatenating the strings at the end. But then you (both) started discussion `format` and `GC` and I'm still not sure what is going on (I understand @Dicebot notice about GC). But maybe it doesn't really matter. IMO everything is ok.

And how did you created a link to my post? I cannot figure out : )

ciechowoj:
Moved `anonymousNames` and `includeHandler` global object from global scope to new `Context` class.

jacob-carlborg:
> And how did you create a link to my post? I cannot figure out : )

The timestamp on a post is a link.

ciechowoj:
> The timestamp on a post is a link.

Wow, that's a surprise : ).

jacob-carlborg:
> I assumed the problem is I'm not doing the translation on-line and I'm concatenating the strings at the end. But then you (both) started discussion format and GC and I'm still not sure what is going on (I understand @Dicebot notice about GC). But maybe it doesn't really matter. IMO everything is ok.

What I was referring to is basically the same as what Dicebot said. Here is a perfect example that demonstrates the issue [1]. Before your changes that code appended to a single buffer. That means that an allocation is only necessary if the capacity of the buffer has been exceeded. With your changes there's (at least) one allocation when building the string using `format` and additional allocations (possibly) when combining all strings that was created using `format`.

With only a single buffer there's more flexibility in that you can pre allocate memory for the buffer, if needed.

I'm not sure which solution is best, and I'm not even sure if this _is_ an issue. @Dicebot is using a single buffer something you think is worth pursuing?

[1] https://github.com/jacob-carlborg/dstep/pull/56/files#diff-7926c0d03fae867bba699d01bfccd928R225

jacob-carlborg:
My original idea with the "global" output object was to be able to have a single buffer used everywhere without having to clutter the interfaces by always passing around the buffer. Note that in D it's not global in the same sense as in C. It's a thread local module variable.

ciechowoj:
> What I was referring to is basically the same as what Dicebot said. Here is a perfect example that demonstrates the issue [1]. Before your changes that code appended to a single buffer. That means that an allocation is only necessary if the capacity of the buffer has been exceeded. With your changes there's (at least) one allocation when building the string using format and additional allocations (possibly) when combining all strings that was created using format.

Good, at least now I know we are talking about the same thing. So, as I said I'm open for suggestions.

> My original idea with the "global" output object was to be able to have a single buffer used everywhere without having to clutter the interfaces by always passing around the buffer. Note that in D it's not global in the same sense as in C. It's a thread local module variable.

Yes, but they still have the rest of problems of global object. E.g. it forces me to add that `reset` methods I did at the beginning of tests are ugly. And there are other situations when that global objects may become problematic e.g. translating multiple files one after another in single thread.

But yes, passing that `context` is problematic too. I was thinking about putting all that functions to a class so that `context` is implicit.

jacob-carlborg:
> Not crucial but I'd favor it. Right now it is all a non-issue because dstep simply doesn't operate on amount of data for performance to really matter. 

There's another open pull request that adds support for handling multiple input files at once. The amount of data may increase.

> But I think it is more pragmatical to write with more memory friendly API from the very start than to do major refactoring of half dstep later if it ever does become an issue.

That's what I'm worried about.

jacob-carlborg:
>  I was thinking about putting all that functions to a class so that context is implicit.

All the `translate...` methods? Most of them are already wrapped in a class.

ciechowoj:
> All the translate... methods? Most of them are already wrapped in a class.

The methods that translate types aren't, and there is bunch of functions that need(ed) a access to that context too.

Btw. I started to work on keeping macro constants in expressions, and the type-translating functions would need too take `Cursor` instead of `Type` (there is some information about type in the `Cursor` that type doesn't carry).

> Not crucial but I'd favor it. Right now it is all a non-issue because dstep simply doesn't operate on amount of data for performance to really matter. But I think it is more pragmatical to write with more memory friendly API from the very start than to do major refactoring of half dstep later if it ever does become an issue.

I'll rewrite it then.

ciechowoj:
Well, I've just rewritten the code so that it appends the result to one buffer. There is few places needing some refactoring, I'll fix them in nearby future.

dstep/translator/Context.d
```diff
@@ -0,0 +1,54 @@
+/**
+ * Copyright: Copyright (c) 2016 Wojciech Szęszoł. All rights reserved.
+ * Authors: Wojciech Szęszoł
+ * Version: Initial created: Mar 21, 2016
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+
+module dstep.translator.Context;
+
+import mambo.core._ : isBlank;
```
jacob-carlborg: The `_` module imports all modules in the core package. Feels a bit weird to do selective imports on that.

dstep/translator/Context.d
```diff
@@ -0,0 +1,54 @@
...
+        return name;
+    }
+
+    @property public IncludeHandler includeHandler()
+    {
+        return mIncludeHandler;
+    }
+
+    private string[Cursor] anonymousNames;
+    private IncludeHandler mIncludeHandler;
```
jacob-carlborg: Please append an underscore for instance variables that conflict with methods.

dstep/translator/Output.d
```diff
@@ -14,124 +14,287 @@ import mambo.core._;
...
+    private void indent(int shift)
+    {
+        foreach (x; 0..stack.length - 1 + shift)
+            accum.put("    ");
+    }
+
+    private Appender!string accum;
+    private Entity[] stack;
+    private Entity first = Entity.bottom;
+    private string weak = null;
```
jacob-carlborg: Please put all instance variables at the beginning of classes/structs.

dstep/translator/Output.d
```diff
@@ -14,124 +14,287 @@ import mambo.core._;
...
+            accum.put("    ");
+    }
+
+    private void indent(int shift)
+    {
+        foreach (x; 0..stack.length - 1 + shift)
+            accum.put("    ");
+    }
+
+    private Appender!string accum;
```
jacob-carlborg: Please avoid shortened words, perhaps `buffer` is a good alternative.

dstep/translator/Output.d
```diff
@@ -14,124 +14,287 @@ import mambo.core._;
...
+    public void separator()
+    {
+        if (stack.back == Entity.singleLine)
+            stack.back = Entity.separator;
+    }
 
-    alias currentContext this;
+    public void output(Output output)
+    {
+        if (output.stack == [Entity.singleLine])
```
jacob-carlborg: I think it's better to check the length of `stack` and compare the first element than allocating a new array to compare with.

jacob-carlborg:
There are a lot of calls to `format` which I expected to be `formattedWrite` instead. That's stopping you from using `formattedWrite`?

ciechowoj:
> There are a lot of calls to format which I expected to be formattedWrite instead. What's stopping you from using formattedWrite?

Nothing, I'm going to add it, it still work in progress.

jacob-carlborg:
> Nothing, I'm going to add it, it still work in progress.

I see.

ciechowoj:
Well, I've just fixed the issues you enumerated and replaced the most of `format`s with `formattedWrite`.

dstep/translator/Output.d
```diff
@@ -14,124 +14,302 @@ import mambo.core._;
...
+        private Output output;
+        private string sentinel;
+
+        ~this()
+        {
+            bool flushed = output.flush(false);
 
-        if (extraNewline)
-            newline ~= "\n";
+            auto back = output.stack.back;
```
jacob-carlborg: Trailing whitespace.

ciechowoj: Fixed.

ciechowoj:
Is there anything more I should fix here?

jacob-carlborg:
Not sure if I've misunderstood but I still see several `format` instead of `formattedWrite`.

ciechowoj:
Ok, I replaced a few more but the rest must stay. They are in places that do not know about `Output` or are not associated with the translation process.

ciechowoj:
I wonder if making `Output` a class was a good idea. I do not see any advantages beside that it isn't necessary to put `ref` when passing it as argument. What do you think about it? Maybe I should change it to struct?

jacob-carlborg:
What disadvantages do you see except for the allocation of an Output object?

ciechowoj:
The allocation and GC is my only concern.

jacob-carlborg:
Sure, converting to a struct would be fine.

ciechowoj:
I think, it must be left as class. Structs cannot have default constructors and I need one. I tried to initialize the fields directly, but as some of them have reference semantics they starts to be shared between different instances of class/struct.

ciechowoj:
Well, it could be changed for class but first some [bug](https://forum.dlang.org/post/mailman.1332.1460623226.26339.digitalmars-d-bugs@puremagic.com) in phobos needs to be fixed. For time being I would prefer to leave it as is.

jacob-carlborg:
It's fine to keep at as a class as well.

ciechowoj:
So, let it be.

jacob-carlborg:
There are merge conflicts.

ciechowoj:
Ok, I fixed the conflict by doing merge. Let me know if it's ok.

jacob-carlborg:
No, I don't want any extra pointless merge commits. I recommend rebasing instead of merging when updating. Please also squash the commits into one.

dstep/translator/Context.d
```diff
@@ -0,0 +1,54 @@
...
+        return name;
+    }
+
+    @property public IncludeHandler includeHandler()
+    {
+        return includeHandler_;
+    }
+
+    private string[Cursor] anonymousNames;
+    private IncludeHandler includeHandler_;
```
jacob-carlborg: Please put instance variables at the top.

unit_tests/OutputUnitTests.d
```diff
@@ -0,0 +1,519 @@
...
+    output.append(";");
+
+    assertEq(output.data, "int x;");
+}
+
+
+
+
+
+
```
jacob-carlborg: Please remove these extra newlines at the end.

ciechowoj:
Done.

jacob-carlborg:
:ok_hand:

jacob-carlborg:
Thanks.

ciechowoj:
Thanks for reviewing it, too : ).

PR #57, opened 2016-03-17
Adding parallel processing of many files.

genuinelucifer:
Currently it works fine if one file is passed. But if I pass more than one file then i get only ```extern(C):``` in the output.
I'm trying to find the mistake.
I would be thankful for any suggestion(s) regarding approach and/or the coding style (since it is my first PR to this project, I might have missed many coding styles standard for this projects)...

### DIFF
--- a/dstep/driver/Application.d
+++ b/dstep/driver/Application.d
@@ -8,10 +8,13 @@ module dstep.driver.Application;
 
 import std.getopt;
 import std.stdio : writeln, stderr;
-import Path = std.path : setExtension;
+import Path = std.path;
+import std.file;
+import std.parallelism;
 
 import DStack = dstack.application.Application;
 
+import mambo.arguments.Arguments;
 import mambo.core._;
 import mambo.util.Singleton;
 import mambo.util.Use;
@@ -34,12 +37,6 @@ class Application : DStack.Application
 
     private
     {
-        string[] inputFiles;
-
-        Index index;
-        TranslationUnit translationUnit;
-        DiagnosticVisitor diagnostics;
-
         Language language;
         string[] argsToRestore;
         bool helpFlag;
@@ -52,12 +49,27 @@ class Application : DStack.Application
 
         if (arguments.argument.input.hasValue)
         {
-            inputFiles ~= arguments.argument.input;
-            startConversion(inputFiles.first);
+            auto inputFiles = arguments.argument.input.values[1 .. $];
+            string outputDir = null;
+            if (arguments.output != "" && inputFiles.length > 0)
+            {
+                outputDir = arguments.output;
+                if (!exists(outputDir) || !outputDir.isDir)
+                    mkdirRecurse(outputDir);
+            }
+            foreach(string fileName; inputFiles)
+            {
+                auto conversionTask = task!startParsingFile(language, argsToRestore, fileName, true,
+                                                            arguments, compiler, compilerArgs, outputDir);
+                conversionTask.executeInNewThread();
+            }
+
+            startParsingFile(language, argsToRestore, arguments.argument.input.values[0], false, arguments,
+                             compiler, compilerArgs, outputDir);
         }
 
         else
-            startConversion("");
+            startParsingFile(language, argsToRestore, "", false, arguments, compiler, compilerArgs, null);
     }
 
     protected override void setupArguments ()
@@ -65,11 +77,11 @@ class Application : DStack.Application
         arguments.sloppy = true;
         arguments.passThrough = true;
 
-        arguments.argument("input", "input files");
+        arguments.argument("input", "input files").params(1, int.max);
 
         arguments('o', "output", "Write output to")
             .params(1)
-            .defaults(&defaultOutputFilename);
+            .defaults("");
 
         arguments('x', "language", "Treat subsequent input files as having type <language>")
             .params(1)
@@ -79,55 +91,15 @@ class Application : DStack.Application
         arguments("objective-c", "Treat source input file as Objective-C input.");
     }
 
+    static void startParsingFile(Language lang, string[] argsToRestore, string fileName, bool createOutputFileName,
+                                 Arguments args, Compiler comp, string[] compilerArguments, string outputDir)
+    {
+        auto parseFile = ParseFile(lang, argsToRestore, fileName, createOutputFileName,
+                                   args, comp, compilerArguments, outputDir);
+        parseFile.startConversion();
+    }
+
 private:
-
-    string defaultOutputFilename ()
-    {
-        return Path.setExtension(arguments.argument.input, "d");
-    }
-
-    void startConversion (string file)
-    {
-        if (arguments["objective-c"])
-            argsToRestore ~= "-ObjC";
-
-        index = Index(false, false);
-        translationUnit = TranslationUnit.parse(
-            index, 
-            file, 
-            compilerArgs,
-            compiler.extraHeaders);
-
-        // hope that the diagnostics below handle everything
-        // if (!translationUnit.isValid)
-        //     throw new DStepException("An unknown error occurred");
-
-        diagnostics = translationUnit.diagnostics;
-
-        scope (exit)
-            clean;
-
-        if (handleDiagnostics && file.any)
-        {
-            Translator.Options options;
-            options.outputFile = arguments.output;
-            options.language = language;
-
-            auto translator = new Translator(translationUnit, options);
-            translator.translate;
-        }
-    }
-
-    void clean ()
-    {
-        translationUnit.dispose;
-        index.dispose;
-    }
-
-    bool anyErrors ()
-    {
-        return diagnostics.length > 0;
-    }
 
     void handleArguments ()
     {
@@ -186,22 +158,111 @@ private:
         return remainingArgs ~ extraArgs;
     }
 
-    bool handleDiagnostics ()
+    struct ParseFile
     {
-        bool translate = true;
-
-        foreach (diag ; diagnostics)
+        private
         {
-            auto severity = diag.severity;
+            string inputFile;
+            string outputFile;
+            string outputDirectory;
+            Index index;
+            TranslationUnit translationUnit;
+            DiagnosticVisitor diagnostics;
 
-            with (CXDiagnosticSeverity)
-                if (translate)
-                    translate = !(severity == CXDiagnostic_Error || severity == CXDiagnostic_Fatal);
-
-            writeln(stderr, diag.format);
+            Language language;
+            string[] argsToRestore;
+            Compiler compiler;
+            string[] compilerArgs;
+            Arguments arguments;
         }
 
-        return translate;
+        this(Language language, string[] argsToRestore, string fileName, bool createOutputFileName, Arguments arguments,
+             Compiler compiler, string[] compilerArgs, string outputDir)
+        {
+            this.language = language;
+            this.argsToRestore = argsToRestore.dup;
+            this.arguments = arguments;
+            this.compiler = compiler;
+            this.compilerArgs = compilerArgs.dup;
+            
+            inputFile = fileName;
+
+            if (outputDir != null)
+                outputFile = Path.buildPath(outputDir, defaultOutputFilename());
+            else if (!createOutputFileName && arguments.output != "")
+                outputFile = arguments.output;
+            else
+                outputFile = defaultOutputFilename();
+        }
+
+        void startConversion ()
+        {
+            if (arguments["objective-c"])
+                argsToRestore ~= "-ObjC";
+            index = Index(false, false);
+            translationUnit = TranslationUnit.parse(
+                index, 
+                inputFile, 
+                compilerArgs,
+                compiler.extraHeaders);
+
+            // hope that the diagnostics below handle everything
+            // if (!translationUnit.isValid)
+            //     throw new DStepException("An unknown error occurred");
+
+            diagnostics = translationUnit.diagnostics;
+
+            if (handleDiagnostics && exists(inputFile))
+            {
+                Translator.Options options;
+                options.outputFile = outputFile;
+                options.language = language;
+
+                auto translator = new Translator(translationUnit, options);
+                translator.translate;
+            }
+        }
+
+        ~this()
+        {
+            clean();
+        }
+
+    private:
+
+        string defaultOutputFilename ()
+        {
+            return Path.setExtension(Path.baseName(inputFile), "d");
+        }
+
+        void clean ()
+        {
+            translationUnit.dispose;
+            index.dispose;
+        }
+
+        bool anyErrors ()
+        {
+            return diagnostics.length > 0;
+        }
+
+        bool handleDiagnostics ()
+        {
+            bool translate = true;
+        
+            foreach (diag ; diagnostics)
+            {
+                auto severity = diag.severity;
+
+                with (CXDiagnosticSeverity)
+                    if (translate)
+                        translate = !(severity == CXDiagnostic_Error || severity == CXDiagnostic_Fatal);
+
+                writeln(stderr, diag.format);
+            }
+
+            return translate;
+        }
     }
 
     override protected void showHelp ()
@@ -213,6 +274,7 @@ private:
         println();
         println("Options:");
         println("    -o, --output <file>          Write output to <file>.");
+        println("    -o, --output <directory>     Write all the files to <directory>, in case of multiple input files.");
         println("    -ObjC, --objective-c         Treat source input file as Objective-C input.");
         println("    -x, --language <language>    Treat subsequent input files as having type <language>.");
         println("    -h, --help                   Show this message and exit.");
--- a/test_files/multiThreadTest1.h
+++ b/test_files/multiThreadTest1.h
@@ -1,0 +1,1 @@
+void func1(void);
--- a/test_files/multiThreadTest2.h
+++ b/test_files/multiThreadTest2.h
@@ -1,0 +1,1 @@
+void func2(void);
--- a/unit_tests/Common.d
+++ b/unit_tests/Common.d
@@ -14,6 +14,7 @@ import std.path;
 import std.conv;
 import std.algorithm;
 import std.array;
+import std.typecons;
 
 import clang.c.Index;
 
@@ -251,8 +252,7 @@ string[] findExtraGNUStepPaths(string file, size_t line)
 }
 
 void assertRunsDStep(
-    string expectedPath,
-    string actualPath,
+    Tuple!(string, string)[] filesPaths,
     bool strict,
     string[] arguments,
     string file = __FILE__,
@@ -262,17 +262,37 @@ void assertRunsDStep(
     import std.path : baseName;
     import std.format : format;
     import clang.Util : namedTempDir;
-    import std.file : readText;
+    import std.file : readText, mkdirRecurse;
 
-    assertFileExists(expectedPath, file, line);
-    assertFileExists(actualPath, file, line);
+    string[] actualPaths;
+
+    foreach (Tuple!(string, string) filesPath; filesPaths)
+    {
+        assertFileExists(filesPath[0], file, line);    //Expected Paths
+        assertFileExists(filesPath[1], file, line);    //Actual Paths
+        actualPaths ~= filesPath[1];
+    }
 
     string outputDir = namedTempDir("dstepUnitTest");
-    string outputPath = buildPath(outputDir, baseName(expectedPath));
+
+    string[] outputPaths;
+    if (filesPaths.length == 1)
+        outputPaths ~= buildPath(outputDir, baseName(filesPaths[0][0]));
+    else
+    {
+        foreach (Tuple!(string, string) filesPath; filesPaths)
+        {
+            outputPaths ~= buildPath(outputDir, baseName(filesPath[0]));
+        }
+    }
 
     scope(exit) rmdirRecurse(outputDir);
 
-    auto command = ["./bin/dstep", actualPath] ~ arguments ~ ["-o", outputPath];
+    auto command = ["./bin/dstep"] ~ actualPaths ~ arguments;
+    if (outputPaths.length == 1)
+        command ~= ["-o", outputPaths[0]];
+    else
+        command ~= ["-o", outputDir];
     auto result = execute(command);
 
     auto sep = "----------------";
@@ -298,47 +318,50 @@ DStep output:
         throw new AssertError(message, file, line);
     }
 
-    if (!exists(outputPath) || !isFile(outputPath))
+    foreach (i, outputPath; outputPaths)
     {
-        auto templ = q"/
-Output file `%4$s` doesn't exist.
-%1$s
-DStep command:
-%2$s
-%1$s
-DStep output:
-%3$s/";
+        if (!exists(outputPath) || !isFile(outputPath))
+        {
+            auto templ = q"/
+    Output file `%4$s` doesn't exist.
+    %1$s
+    DStep command:
+    %2$s
+    %1$s
+    DStep output:
+    %3$s/";
 
-        auto message = format(
-            templ,
-            sep,
-            join(command, " "),
-            result.output,
-            outputPath);
+            auto message = format(
+                templ,
+                sep,
+                join(command, " "),
+                result.output,
+                outputPath);
 
-        throw new AssertError(message, file, line);
-    }
+            throw new AssertError(message, file, line);
+        }
 
-    string expected = readText(expectedPath);
-    string actual = readText(outputPath);
+        string expected = readText(filesPaths[i][0]);
+        string actual = readText(outputPath);
 
-    if (!compareString(expected, actual, strict))
-    {
-        auto fmt = q"/
-Source code translated to:
-%1$s
-%2$s
-%1$s
-Expected D code:
-%1$s
-%3$s
-%1$s
-DStep command:
-%4$s/";
+        if (!compareString(expected, actual, strict))
+        {
+            auto fmt = q"/
+    Source code translated to:
+    %1$s
+    %2$s
+    %1$s
+    Expected D code:
+    %1$s
+    %3$s
+    %1$s
+    DStep command:
+    %4$s/";
 
-        string commandString = join(command, " ");
-        string message = format(fmt, sep, actual, expected, commandString);
-        throw new TranslateAssertError(message, file, line);
+            string commandString = join(command, " ");
+            string message = format(fmt, sep, actual, expected, commandString);
+            throw new TranslateAssertError(message, file, line);
+        }
     }
 }
 
@@ -400,8 +423,7 @@ void assertRunsDStepCFile(
     string[] arguments = ["-Iresources"];
 
     assertRunsDStep(
-        expectedPath,
-        cPath,
+        [tuple(expectedPath, cPath)],
         strict,
         arguments,
         file,
@@ -428,8 +450,23 @@ void assertRunsDStepObjCFile(
     }
 
     assertRunsDStep(
-        expectedPath,
-        objCPath,
+        [tuple(expectedPath, objCPath)],
+        strict,
+        arguments,
+        file,
+        line);
+}
+
+void assertRunsDStepCFiles(
+    Tuple!(string, string)[] filesPaths,
+    bool strict = false,
+    string file = __FILE__,
+    size_t line = __LINE__)
+{
+    string[] arguments = ["-Iresources"];
+
+    assertRunsDStep(
+        filesPaths,
         strict,
         arguments,
         file,
--- a/README.markdown
+++ b/README.markdown
@@ -44,6 +44,12 @@ For translating Objective-C headers add the `-ObjC` flag.
 
     $ dstep Foo.h -o Foo.d -ObjC
 
+For translating multiple files at once, simply pass all the files to dstep.
+In this case though, `-o` (if given) would point to output directory name.
+The directory will be created if it doesn't exist.
+
+    $ dstep Foo1.h Foo2.h Foo3.h .... FooN.h -o ./outputDirectory/
+
 Use `-h` for usage information. Any flags recognized by Clang can be used.
 
 ## Limitations/Known issues
--- a/changelog.md
+++ b/changelog.md
@@ -7,6 +7,7 @@
 * Basic unit tests were added
 * Most of cucumber tests was replaced with D-based tests.
 * Statements are translated in original 'C' order now.
+* Multiple input files can be processed in different threads.
 
 ## Version 0.2.1
 ### New/Changed Features
--- a/test_files/multiThreadTest1.d
+++ b/test_files/multiThreadTest1.d
@@ -1,0 +1,3 @@
+extern (C):
+
+void func1 ();
--- a/test_files/multiThreadTest2.d
+++ b/test_files/multiThreadTest2.d
@@ -1,0 +1,3 @@
+extern (C):
+
+void func2 ();
--- a/unit_tests/IntegrationTests.d
+++ b/unit_tests/IntegrationTests.d
@@ -6,6 +6,7 @@
  */
 
 import Common;
+import std.typecons;
 
 unittest
 {
@@ -16,4 +17,8 @@ unittest
     assertRunsDStepObjCFile(
         "test_files/objc/methods.d",
         "test_files/objc/methods.h");
+
+    assertRunsDStepCFiles(
+        [tuple("test_files/multiThreadTest1.d", "test_files/multiThreadTest1.h"),
+         tuple("test_files/multiThreadTest2.d", "test_files/multiThreadTest2.h")]);
 }


genuinelucifer:
Most probably due to arguments (the way it is defined and used) it is getting a bit out of hand currently. Will try to do the necessary changes to make it work.

genuinelucifer:
Although I am unable to find exactly where ```arguments``` is defined. So, that I could make copies of it and pass to every thread, its own local copy. Is it in one of the libraries used?

dstep/driver/Application.d
```diff
@@ -34,30 +36,35 @@ class Application : DStack.Application
...
-        TranslationUnit translationUnit;
-        DiagnosticVisitor diagnostics;
-
-        Language language;
-        string[] argsToRestore;
-        bool helpFlag;
-        Compiler compiler;
+	    bool helpFlag;
+	Language language;
+	string[] argsToRestore;
```
jacob-carlborg: Please use spaces (four) for indentation.

genuinelucifer: Sorry. Mixed up tabs and spaces. Will correct it.

jacob-carlborg:
`arguments` is defined here: https://github.com/jacob-carlborg/dstack/blob/master/dstack/application/Application.d#L40-L48

dstep/driver/Application.d
```diff
@@ -34,30 +36,35 @@ class Application : DStack.Application
...
     protected override void run ()
     {
         handleArguments;
-
         if (arguments.argument.input.hasValue)
         {
-            inputFiles ~= arguments.argument.input;
-            startConversion(inputFiles.first);
+	    int j = 0;
+	    if(remainingArgs.length >0)
```
jacob-carlborg: Always put a space before the opening parenthesis for statements.

dstep/driver/Application.d
```diff
@@ -80,130 +87,159 @@ class Application : DStack.Application
...
+
+	    string[] extraArgs ()
+	    {
+		return compiler.extraIncludePaths.map!(e => "-I" ~ e).toArray;
+	    }
+
+	    string[] compilerArgs ()
+	    {
+		return remainingArgs ~ extraArgs;
+	    }
```
jacob-carlborg: I don't think these three arguments related functions are necessary to have in this class. I think they will be the same for all instances.

dstep/driver/Application.d
```diff
@@ -80,130 +87,159 @@ class Application : DStack.Application
...
+	    default:
+	        throw new DStepException(`Unrecognized language "` ~ language ~ `"`);
+	}
+
+	argsToRestore ~= "-x";
+	argsToRestore ~= language;
     }
-
-    bool handleDiagnostics ()
+    class ParseFile
```
jacob-carlborg: A struct would be a better fit than a class.

.gitignore
```diff
@@ -20,3 +20,7 @@ clangs
 bin/test
 bin/unittest
 .DS_Store
+*.so.*
+*.so
```
jacob-carlborg: This is already present at line 10.

.gitignore
```diff
@@ -20,3 +20,7 @@ clangs
...
+/bin/*
+*~
```
jacob-carlborg: What is this for?

.gitignore
genuinelucifer: Sorry didn't see.

.gitignore
genuinelucifer: It is to not add the temporary files. Many temporary files are created
which end wirh '~', to ignore them, i put it.

dstep/driver/Application.d
genuinelucifer: Sorry. Will do.

dstep/driver/Application.d
genuinelucifer: Ok. Will change.

.gitignore
jacob-carlborg: Are they created by DStep or something else?

dstep/driver/Application.d
genuinelucifer: Yes,seems so. Will change.

.gitignore
genuinelucifer: By the texteditor.

genuinelucifer:
I'll look into dstack and try to figure out how to implement this.

.gitignore
jacob-carlborg: Please remove that. It should not be part of project specific ignore files. If you want to ignore that I recommend either adding a global ignore file to ignore that across all git projects or add it to the private ignore file for the project. You should be able to find this by searching online.

genuinelucifer: Ohk.  Will check.
Actually i just added everything i found git added when i did 'git add .'
Will change it.

genuinelucifer:
@jacob-carlborg Please review. It is working fine for me now.

jacob-carlborg:
Most things I commented on are not fixed as far as I can see.

dstep/driver/Application.d
```diff
@@ -34,30 +38,40 @@ class Application : DStack.Application
...
 
     protected override void run ()
     {
         handleArguments;
-
         if (arguments.argument.input.hasValue)
         {
-            inputFiles ~= arguments.argument.input;
-            startConversion(inputFiles.first);
+            int j = 0;
```
jacob-carlborg: I don't thinks this is used.

dstep/driver/Application.d
```diff
@@ -34,30 +38,40 @@ class Application : DStack.Application
...
+            auto remArgs = remainingArgs;
```
jacob-carlborg: You're not really supposed to use `remainingArgs` like this. Instead you can access all files passed on the command line as `arguments.argument.input.values`. I think you need to increase the maximum numbers of allowed values for the `input` argument in the `setupArguments` method. See [this](https://github.com/jacob-carlborg/mambo/blob/master/mambo/arguments/Arguments.d#L264) for reference.

dstep/driver/Application.d
```diff
@@ -34,30 +38,40 @@ class Application : DStack.Application
...
+            if(remArgs.length > 0)
+            foreach(string fileName; remArgs)
```
jacob-carlborg: Is it not possible to use a parallel foreach? Something like this:

```d
auto intputFiles = arguments.argument.input.values;
foreach (file ; inputFiles.parallel) { startParsingFile ... }
```

dstep/driver/Application.d
```diff
@@ -79,55 +93,13 @@ class Application : DStack.Application
...
-            translator.translate;
-        }
-    }
-
-    void clean ()
+    static void startParsingFile(Language lang, string[] argsTR, string fileName, bool createOutputFileName, Arguments args, Compiler comp)
     {
-        translationUnit.dispose;
-        index.dispose;
+        auto pF = new ParseFile(lang, argsTR, fileName, createOutputFileName, args, comp);
```
jacob-carlborg: There's no reason to use `new`.

dstep/driver/Application.d
```diff
@@ -176,34 +154,122 @@ private:
...
+                compiler.extraHeaders);
+
+            // hope that the diagnostics below handle everything
+            // if (!translationUnit.isValid)
+            //     throw new DStepException("An unknown error occurred");
+
+            diagnostics = translationUnit.diagnostics;
+
+            scope (exit)
+                clean;
```
jacob-carlborg: Now when this is a struct, is it possible to move this to a destructor?

jacob-carlborg:
I'm wondering if it's better that the `-o` flag specifies which directory to place all the resulting files in when there are multiple input files.

dstep/driver/Application.d
genuinelucifer: Sorry, will do.

dstep/driver/Application.d
genuinelucifer: I used parallel foreach in the same code, but for some reason it didn't work! That's why changed it into tasks... I'll try again though.

dstep/driver/Application.d
genuinelucifer: Sorry, just a habit. Will change.

dstep/driver/Application.d
genuinelucifer: Sure. Will do.

genuinelucifer:
Sure seems better if ```-o``` tells the directory in case of multiple files.

dstep/driver/Application.d
genuinelucifer: Sorry, had used it in previous code, forgot to remove...

genuinelucifer:
I did the changes. Also, please tell me how much to keep the ```max``` value for ```Arguments``` class in mambo. For my usage, I kept it to 512 and worked. But, I don't know how to decide the exact amount to be kept in mambo.
.
Also, I tried to use ```parallel``` again. This time the problem is that, if I give 2 input files. The first output file is fine. But the second output file contains the combined output from 1st AND 2nd file. While using this version of ```task``` works as required. I'll try to come up with a solution to using ```parallel``` though if it is required.

genuinelucifer:
And one of the buildbots is passing and other one is failing. I saw the output but can't figure out what actually is the error that is shown. Please help.

jacob-carlborg:
> And one of the buildbots is passing and other one is failing. I saw the output but can't figure out what actually is the error that is shown. Please help.

The error messages are not very good. But the issue is that DStep fails to produce a result in the case of `features/objc/method.feature`, for some reason.

dstep/driver/Application.d
```diff
@@ -34,30 +38,34 @@ class Application : DStack.Application
...
+                    auto conversionTask = task!startParsingFile(language, argsToRestore, fileName, createNewFileName, arguments, compiler, compilerArgs);
+                    conversionTask.executeInNewThread();
+                }
+            }
         }
 
         else
-            startConversion("");
+        {
+            auto conversionTask = task!startParsingFile(language, argsToRestore, "", false, arguments, compiler, compilerArgs);
```
jacob-carlborg: I don't think it's necessary to use a task for this condition.

dstep/driver/Application.d
```diff
@@ -34,30 +38,34 @@ class Application : DStack.Application
...
-
         if (arguments.argument.input.hasValue)
         {
-            inputFiles ~= arguments.argument.input;
-            startConversion(inputFiles.first);
+            auto inputFiles = arguments.argument.input.values;
+            bool createNewFileName = inputFiles.length > 1 ? true : false;
+            foreach(string fileName; inputFiles)
+            {
+                if(exists(fileName))
```
jacob-carlborg: The previous code didn't check if the file existed. Might be confusing if a file doesn't exist and is silently ignored.

dstep/driver/Application.d
```diff
@@ -34,30 +38,34 @@ class Application : DStack.Application
...
     {
         handleArguments;
-
         if (arguments.argument.input.hasValue)
         {
-            inputFiles ~= arguments.argument.input;
-            startConversion(inputFiles.first);
+            auto inputFiles = arguments.argument.input.values;
+            bool createNewFileName = inputFiles.length > 1 ? true : false;
+            foreach(string fileName; inputFiles)
```
jacob-carlborg: Please have an empty newline before and after statements like if, foreach and so on.

genuinelucifer:
Again one build is passing, other one is not. The tests are failing on clang 3.7 and the best part is I have clang 3.7 only... How do I run these feature tests on my machine? Or what could be the probable cause for this failure?

jacob-carlborg:
One building is running on Linux and the other one on OS X. The one on OS X is running additional tests, converting Objective-C to D. It's one of those that is failing.

You either need OS X to run those test or install GNUStep on Linux. I think you can run basically the same tests by running `dub test`, @ciechowoj recently added that capability. That might give you better error messages.

jacob-carlborg:
> Also, please tell me how much to keep the max value for Arguments class in mambo. For my usage, I kept it to 512 and worked. But, I don't know how to decide the exact amount to be kept in mambo

That's a good question. You can always use `int.max` and it will most likely be limited by something else :smiley:. Otherwise, what's the maximum number of files one could expect there to be in a C library.

ciechowoj:
> Again one build is passing, other one is not. The tests are failing on clang 3.7 and the best part is I have clang 3.7 only... How do I run these feature tests on my machine? Or what could be the probable cause for this failure?

If you are on linux try to install GNUStep and comment `if OSX` in `features/step_definitions/common.rb`.

> You either need OS X to run those test or install GNUStep on Linux. I think you can run basically the same tests by running dub test, @ciechowoj recently added that capability. That might give you better error messages.

The unit tests seems to pass.

    Testing with libclang version 3.6.0
    Running unit tests
    Running cucumber tests

    int test ()
    {
        println("Running cucumber tests ");

        auto result = execute("cucumber");

        if (result.status != 0)
            println(result.output);

        return result.status;
    }

genuinelucifer:
I ran the tests on my machine and they are all passing.
I am out of ideas.
What should I look for to ensure that the build would pass on the builbot? Any Ideas?

ciechowoj:
I cloned you repository and it doesn't pass the tests on my (linux) machine. Make sure you are making your changes against the master branch HEAD. You could also run the `cucumber` directly in project main directory, then it will show you which test are really run. Verify that MacOS tests are run.

genuinelucifer:
Due to some dependency issues I am unable to install GNUStep on my Ubuntu.....
I am working to sort it out. (Some problem because of the way I installed libbz2 to run dub)..
.
Meanwhile, it would really help if @jacob-carlborg could review my draft for the GSoC proposal which I submitted to Google a few days back. :)

jacob-carlborg:
Sure. just send it to me.

genuinelucifer:
I have submitted it to Google and it probably would be visible to you.
Or, please provide me your email-id so I could send it to you..

jacob-carlborg:
> I have submitted it to Google and it probably would be visible to you.

Well, you need to send me a link or something for me to find it :wink: 

> Or, please provide me your email-id so I could send it to you..

You can find my email in the git history for one of my commits.

genuinelucifer:
> You can find my email in the git history for one of my commits.

Sent you a mail with the link to the proposal :)

jacob-carlborg:
> Due to some dependency issues I am unable to install GNUStep on my Ubuntu.....
I am working to sort it out. (Some problem because of the way I installed libbz2 to run dub)..

I have two suggestions (unless you can figure out the dependency issues):

* Copy the header files from GNUStep and place it somewhere Clang can find them. I think the header files are all what's required

* Setup and run the tests inside a Docker container

genuinelucifer:
Finally fixed all the messy dependencies (As I tried putting the headers but it came up with its own list of errors!!)....
.
But I had to change require "aruba/cucumber" to "cucumber" in ```env.rb```.
Also, now I get the file not found errors . A small part of which is :
```
lucifer@ubuntu:~/GSoC/dstep$ cucumber
Feature: Array

  Scenario: Convert array         # features/array.feature:3
  File exists @ sys_fail2 - (../../libclang.so, tmp/aruba/libclang.so) (Errno::EEXIST)
  /home/lucifer/GSoC/dstep/features/support/env.rb:36:in `symlink'
  /home/lucifer/GSoC/dstep/features/support/env.rb:36:in `Before'
    Then I test the file "arrays" # features/step_definitions/common.rb:39

```
It seems it can't find the libclang.so file. So, I copy pasted that file in all of the directories ```/home/lucifer``` , ```/home/lucifer/GSoC```, ```/home/lucifer/GSoC/dstep```
Still the same error.
My libclang.so is a symlink and it points to libclang.so.3.7 in the same folder. So I copied that file too in all the above mentioned folders.
Still, the same errors.. Any suggestions?

jacob-carlborg:
The test is trying to create a symlink from `dstep/libclang.so` to `dstep/tmp/aruba`. The test will create the target directory. Have you installed aruba? If not, run `bundle install` to install all Ruby dependencies.

genuinelucifer:
I did a ```bundle install``` and then too ```cucumber``` gives the same errors.

genuinelucifer:
Sorry. My libclang.so.3.7 was not in path. It was in the same directory but didn't work. I just copied it to /usr/lib and it worked. Thanks for the help...

jacob-carlborg:
It should work if `libclang.so` is in the DStep root directory.

genuinelucifer:
It was in the DStep directory, but didn't work :(
.
But after copying to lib it is working fine and I can see that one test is failing on cucumber.

genuinelucifer:
Well, the error in cucumber is:
```
Feature: Include

  Scenario: Convert include        # features/include.feature:3
    Then I test the file "include" # features/step_definitions/common.rb:39
      No such file or directory @ rb_sysopen - /home/lucifer/GSoC/dstep/tmp/aruba/include.d (Errno::ENOENT)
      ./features/step_definitions/common.rb:24:in `/^the files "([^"]*)" and "([^"]*)" should be equal$/'
      ./features/step_definitions/common.rb:43:in `/^I test the file "([^"]*)"$/'
      features/include.feature:4:in `Then I test the file "include"'

```
But when I check for ```/home/lucifer/GSoC/dstep/tmp/aruba/include.d```, it actually exists!!

ciechowoj:
Try to run dstep manually on that file, with the same flags as cucumber is using.

genuinelucifer:
Sorry, how do I get the flags?

genuinelucifer:
And also, sorry again...
No file is present in ```dstep/tmp/aruba/```
All the *.d and *.h files are present in ```dstep/tmp/aruba/test_files/```!!

ciechowoj:
> Sorry, how do I get the flags?

Make dstep to print them to stderr/check them in cucumber feature definitions.

If I remember correctly I was getting similar errors to yours and it was because the file was not translating.

jacob-carlborg:
If a file doesn't exist it's because it wasn't translated. The files in `dstep/tmp/aruba/test_files` are copied from `dstep/test_files` and are the test files. The header files to translate and the expected result, the D files. The result of a translation should end up in `dstep/tmp/aruba`

genuinelucifer:
@jacob-carlborg I checked ```dstep/tmp/aruba``` only has one file ```variables.d```
How are other tests passing and only one test failing.
I have done some changes (which I am pushing now) but am unable to figure out any cause of the file not being translated. Please have a look.

jacob-carlborg:
If a test fails due to a file not existing, run that test manually, run dstep on that test file and see what happens.

jacob-carlborg:
You might want to take into consideration that the "global" output object is thread local, but (I think) there's not direct mapping on tasks and threads. Not sure if this causing any problems, but it could. @ciechowoj is working on removing the global output object.

genuinelucifer:
Running dstepon include.h gives following d file:
```
import core.stdc.stdio;

extern (C):

extern __gshared size_t a;
extern __gshared FILE b;
```
.
Also, since I separated the translation of 1st file to current thread. The output object is actually in the same thread and no other thread exists (as only one file is parsed)..
I am unable to think of some other solution.

jacob-carlborg:
The above translation is correct. If you look at the output from Travis you can see that it's the Objective-C "method" test that fails.

genuinelucifer:
If I run the same file with ```-ObjC``` I get the following output:
```
import core.stdc.stdio;

extern (Objective-C):

extern __gshared size_t a;
extern __gshared FILE b;

```

genuinelucifer:
Is the output wrong. Or am I supposed to deal with some other file?

jacob-carlborg:
I don't really understand what you're doing. Please specify the exact command you're running and what output you get.

genuinelucifer:
I copied the ```include.h``` to bin directory of dstep and did:
```./dstep include.h -ObjC```
and got the file with the contents as I posted earlier.

**Comment by jacob-carlborg** [comment]
Yeah, that's expected. When you specify the language, as you do with `-ObjC`, DStep will assume all code is in that language, Objective-C in this case, and use the language to write the extern declaration: `
extern (Objective-C)`.

**Comment by genuinelucifer** [comment]
Okay. So the file is being processed. But cucumber still can't find the
file. Any suggestions on where to look for the root of the problem?

**Comment by jacob-carlborg** [comment]
Not sure why you're looking at the `include.h` file, it's `objc/methods.h` that fails.

**Comment by genuinelucifer** [comment]
Cause when I ran cucumber include test was failing.
And now I copied all clang headers to /usr/include
After this arrays test is failing. Others are passed!!!

**Comment by genuinelucifer** [comment]
And wow. Now all the tests are passing with final cucumber output as:
```
19 scenarios (6 undefined, 13 passed)
19 steps (6 undefined, 13 passed)
0m1.998s

You can implement step definitions for undefined steps with these snippets:

Then(/^I test the Objective\-C file "(.*?)" in "(.*?)"$/) do |arg1, arg2|
  pending # express the regexp above with the code you wish you had
end


```

**Comment by genuinelucifer** [comment]
All the objective C tests are shown in yellow (undefined).
What could have caused them to be undefined? Probably because of working late night, I am seeing cucumber magically change its output. I will try tomorrow again.

**Comment by jacob-carlborg** [comment]
That definition is guarded by a check for OS X. Look in `common.rb` at the bottom.

Since you seem to get some odd behavior could it be some multi thread issue?

**Comment by genuinelucifer** [comment]
Ok. I did not remove that gaurd and ran in linux. Silly mistake.
.
And now I am getting errors regarding some header file not found when using ```ObjC`` with methods.h
Will try to figure out and finish this PR within 1-2 days.

**Comment by jacob-carlborg** [comment]
You need GNUStep for that.

**Comment by genuinelucifer** [comment]
I have installed GNUStep. I guess I messed up some files by keeping them in
path and some I copied to usr folders.

**Comment by genuinelucifer** [comment]
I am getting following error on processing methods.h with ```./dstep methods.h -ObjC```
```
File(977710, "")methods.h:1:9: fatal error: 'Foundation/NSObjCRuntime.h' file not found
```
Although I have install GNUStep from terminal, I downloaded source for GNUStepBase and copied all folders in ```Headers``` to ```/usr/include```
But then I get error that could not find ```GSConfig.h```. I found that the original file is named ```GSConfig.h.in``` So I renamed it to remove the ```.in``` part.
Now I am getting lots and lots of errors in the GSConfig file (syntax errors).
.
What Should I do?
I know this PR has taken ages now, but these errors are just..... unbelievable..... :(

**Comment by ciechowoj** [comment]
I had similar problem to yours with GNUStep on linux. For testing purposes (in your case) you could add `-I` flags when invoking dstep, e.g. `-I/usr/include/GNUstep`. It isn't very good to copy around the header files in your system directly, neither is renaming `.in` files to `.h` files . When I had similar problem, I just used `find` command to locate the file in my system and passed it with `-I` flag then (of course for testing purposes, to locate other errors, test for `ObjC` aren't run on linux by default).

**Comment by ciechowoj** [comment]
E.g. on my system:

    user@machine / $ sudo find -name NSObjCRuntime.h
    [sudo] password for user: 
    ./usr/include/GNUstep/Foundation/NSObjCRuntime.h
    user@machine / $ sudo find -name GSConfig.h
    ./usr/include/GNUstep/GNUstepBase/GSConfig.h

**Comment by ciechowoj** [comment]
I would suggest you to update your code to the most recent version of master branch, as some changes to tests were made.

**Comment by genuinelucifer** [comment]
@ciechowoj The problem is when I ```sudo apt-get install GNUstep``` there are no header files in my gnustep directory. That's why I had to download the Headers as well.
.
Weel, i uninstalled and then installed GNUstep again and still:
```
lucifer@ubuntu:/usr/include/GNUstep$ ls
Cynthiune  DBKit  Frameworks  FSNode  GNUMail  Inspector  Operation
```
There is not ```Foundation``` directory.
And
```sudo find -name NSObjCRuntime.h``` doesn't give any results!!

**Comment by genuinelucifer** [comment]
I had to ```sudo apt-get install libgnustep-base-dev``` to finally get those headers. And the errors for not finding of those Foundation headers is gone.
BUT, now it shows problems in ```/usr/include/inttypes.h``` it has a line ```#include_next<inttypes.h>``` and the compiler complaints about not finding ```/usr/include/inttypes.h```!!!!!!!!!!!!!!!!!!!!!!!!!
The same file trying to include itself raising error about itself not being found!!!!!!!!

**Comment by ciechowoj** [comment]
That is strange, you could try to install libc++ and GNU libc++ library and pass the path to the directory directly.

> BUT, now it shows problems in /usr/include/inttypes.h it has a line #include_next<inttypes.h> and the compiler complaints about not finding /usr/include/inttypes.h!!!!!!!!!!!!!!!!!!!!!!!!!

There is some probability that you've broken something by copying the header files.

**Comment by jacob-carlborg** [comment]
I'm suspecting it's a Clang internal header. These are normally shipped with Clang. For convenience DStep is including some of these, I've basically added the ones that I've needed so far.

If you download Clang from LLVM site it's available in the `<clang>/lib/clang/3.7.0/include` path. You can copy the file to `<dstep>/resources` and then add it here [1]. Copy the file from Clang 3.5.0.

[1] https://github.com/jacob-carlborg/dstep/blob/master/clang/Compiler.d#L39

**Comment by jacob-carlborg** [comment]
I'm suspecting `#include_next` indicates it should look for the header file in the next include path, it's not including itself.

**Comment by genuinelucifer** [comment]
Thanks @jacob-carlborg . Finally it seems to have been solved.
.
BUT! yet another error :(
```
File(977710, "")/usr/include/GNUstep/GNUstepBase/GSConfig.h:236:3: error: unknown type name 'uint8_t'
File(977710, "")/usr/include/GNUstep/GNUstepBase/GSConfig.h:239:3: error: unknown type name 'uint8_t'
File(977710, "")/usr/include/GNUstep/Foundation/NSObjCRuntime.h:91:9: error: unknown type name 'intptr_t'
File(977710, "")/usr/include/GNUstep/Foundation/NSObjCRuntime.h:92:9: error: unknown type name 'uintptr_t'
File(977710, "")/usr/include/GNUstep/GNUstepBase/GSObjCRuntime.h:58:11: fatal error: 'objc/objc.h' file not found
```
I am afraid building ObjC files on linux seems a faar fetched thing! :(
Just to confirm, is it possible to test the objectiveC files on linux?

**Comment by jacob-carlborg** [comment]
> Just to confirm, is it possible to test the objectiveC files on linux?

I have no idea :smiley:. I never bother trying that since I have a Mac.

**Comment by genuinelucifer** [comment]
Finally I was able to succesfully run dstep on methods.h
I had to add ```#include<stdint.h>``` to ```GSConfig.h```
.
But cucumber still reports that the test fails. (total 3 tests fail)

**Comment by ciechowoj** [comment]
> Just to confirm, is it possible to test the objectiveC files on linux?

I needed to change the flags in `common.rb` to:

`EXTRA_OBJC_FLAGS = '-I/usr/include/GNUstep -I/usr/lib/gcc/x86_64-linux-gnu/4.8/include -I/usr/lib/gcc/x86_64-linux-gnu/4.8/include-fixed'`

Then the tests pass.

**Comment by ciechowoj** [comment]
However in the most recent version of dstep, the most of the tests is run outside the cucumber and appropriate flags are passed automatically.

**Comment by genuinelucifer** [comment]
Also, i am able to build methods.h perfect well now. (after passing many
include directories)
So, i don't know what's wrong about the test in the build bot. How should I
figure that out if it's running ok for me?

**Comment by ciechowoj** [comment]
Do other tests pass as well?

**Comment by genuinelucifer** [comment]
Well, using cucumber they dont.
But if i manually use dstep, i am able to get an output file. And cucumber
complaints about file does not exist.

**Comment by ciechowoj** [comment]
Well, so try to pass all that additional include directories `EXTRA_OBJC_FLAGS` (remember to remove that `OSX` ifdefs) and check if cucumber passes then.

**Comment by ciechowoj** [comment]
If it passes on your machine and doesn't pass on the bot, I would try to debug it on the bot. After all, you always can enable some verbose mode and/or add some printfs/writelns here and there and verify the travis output.

**Comment by ciechowoj** [comment]
And update the code to the lastest version of dstep, I added some code that prints additional diagnostic information. And as the test are run from `test.d` you do not even need to know cucumber and ruby to make some debugging.

**Comment by genuinelucifer** [comment]
Finally rebased and with just a small change, all the tests are passing on my machine now (after removing the OSX gaurd). :) Thanks @ciechowoj 
.
Waiting for the build bot output. I will add the OSX gaurd back in next commit.

**Comment by ciechowoj** [comment]
That's great : ).

**Comment by genuinelucifer** [comment]
Again :(
Linux build passes. But OSX build fails on that test! That test passed on my machine just after giving the path to objc.h (inc extra_objc_flags) probably something like that might be required for OSX?
ANy help...

File: dstep/driver/Application.d [hunk 15]
```diff
@@ -48,16 +46,22 @@ class Application : DStack.Application
 
     protected override void run ()
     {
+        derr.writefln(arguments.rawArgs);
```
genuinelucifer: Moreover, I do not see the output from this neither on my machine nor on the build bot Am I supposed to print to ```stderr``` in some other way?

jacob-carlborg:
I downloaded you branch and tried to run DStep on `test_files/objc/methods.h`. This is the error I get on OS X:

```
[./bin/dstep,-ObjC]
File(10019B630, "")/System/Library/Frameworks/Foundation.framework/Headers/NSObjCRuntime.h:10:10: fatal error: 'stdarg.h' file not found
```

`stdarg.h` is an internal Clang header which DStep includes (clang/Compiler.d). It might be some issue with multi threading.

genuinelucifer:
Thanks. I'll try to fix this.

genuinelucifer:
All I could find was that this is a common problem with clang that it can't find some header files... All of the solutions were to either move the header files that download with clang to a location of to pass the directory so that clang uses them.
.
So I am trying to ```locate``` the file.
But I am unable to see any output from the command on build bot!!

jacob-carlborg:
Have you been able to run the tests on the master branch on your local machine?

genuinelucifer:
Yes.

jacob-carlborg:
Including the Objective-C tests that are not normally run on Linux?

ciechowoj:
Currently (on the dstep master branch) the Objective-C tests are run, if the GNUStep is installed. Otherwise (GNUstep not installed) the warning is emitted to stderr that GNUStep couldn't be found but tests pass nevertheless.

jacob-carlborg:
@ciechowoj ah, sorry. @genuinelucifer then it looks your changes somehow messes with the Clang internal headers that DStep handles. I suggest you inspect the flags DStep is passing to libclang and make sure these are correct.

genuinelucifer:
> Including the Objective-C tests that are not normally run on Linux?

Yes.
.
.
Can I run sudo on travis? I need to use ```locate``` to find the correct directory to pass to clang. But OSX needs the locate database to be built first!

jacob-carlborg:
> Can I run sudo on travis?

You can on OS X but not on Linux. The Linux builds are running inside Docker containers and for security reasons sudo is not allowed.

genuinelucifer:
@jacob-carlborg 
I just rebased onto master and due to the latest changes and removal of ```cucumber```, my code passes the build!! :D
Please consider reviewing again.

dstep/driver/Application.d
jacob-carlborg: I guess this was some form of debugging that can be removed now?

dstep/driver/Application.d
```diff
@@ -48,16 +46,22 @@ class Application : DStack.Application
...
         handleArguments;
 
         if (arguments.argument.input.hasValue)
         {
-            inputFiles ~= arguments.argument.input;
-            startConversion(inputFiles.first);
+            auto inputFiles = arguments.argument.input.values[1 .. $];
+            foreach(string fileName; inputFiles)
```
jacob-carlborg: Put a space after the `foreach` keyword. Surround the foreach statement with an empty newline.

dstep/driver/Application.d
```diff
@@ -48,16 +46,22 @@ class Application : DStack.Application
...
+            {
+                auto conversionTask = task!startParsingFile(language, argsToRestore, fileName, true, arguments, compiler, compilerArgs);
+                conversionTask.executeInNewThread();
+            }
+            startParsingFile(language, argsToRestore, arguments.argument.input.values[0], false, arguments, compiler, compilerArgs);
```
jacob-carlborg: Why is the first file handled separately?

dstep/driver/Application.d
```diff
@@ -48,16 +46,22 @@ class Application : DStack.Application
...
         handleArguments;
 
         if (arguments.argument.input.hasValue)
         {
-            inputFiles ~= arguments.argument.input;
-            startConversion(inputFiles.first);
+            auto inputFiles = arguments.argument.input.values[1 .. $];
+            foreach(string fileName; inputFiles)
+            {
+                auto conversionTask = task!startParsingFile(language, argsToRestore, fileName, true, arguments, compiler, compilerArgs);
```
jacob-carlborg: This line is a bit too long, please split it up.

dstep/driver/Application.d
```diff
@@ -79,54 +83,17 @@ class Application : DStack.Application
         arguments("objective-c", "Treat source input file as Objective-C input.");
     }
 
-private:
-
-    string defaultOutputFilename ()
+    static void startParsingFile(Language lang, string[] argsTR, string fileName, bool createOutputFileName, Arguments args, Compiler comp, string[] compilerArguments)
```
jacob-carlborg: Please avoid these abbreviations like `argsTR` and `pF`.

jacob-carlborg:
I don't see any tests.

dstep/driver/Application.d
genuinelucifer: Will do.

dstep/driver/Application.d
genuinelucifer: I thought it would reduce total number of threads spawned. Also, if only 1 file is parsed, no need to create extra thread.
Should I include first file in new thread?

dstep/driver/Application.d
genuinelucifer: Will do.

dstep/driver/Application.d
genuinelucifer: Will change.

genuinelucifer:
> I don't see any tests.

I couldn't figure out how to write test for this? I need to parse more than 1 file at once and check both their outputs. Where do I write the test?

dstep/driver/Application.d
jacob-carlborg: I see. My original idea of using `parallel` would handle the thread scheduling automatically. Meaning a new task would not necessarily create a new thread. But that might be a good idea anyway.

ciechowoj:
Just a suggestion: I believe you could write a function similar to `assertRunsDStepCFile` that takes array of file pairs and does appropriate checks. 

    assertRunsDStepCFiles([tuple("test_files/aggregate.d", "test_files/aggregate.h"), tuple("test_files/methods.d", "test_files/methods.h")]);

jacob-carlborg:
I just merged a pull request by ciechowoj which removes the cucumber tests. It adds a new file called `IntegrationTests`.

genuinelucifer:
> could write a function similar to assertRunsDStepCFile that takes array of file pairs and does appropriate checks.

Thanks @ciechowoj

genuinelucifer:
@jacob-carlborg A small change has to be done to ```mambo``` also. I just sent the PR

genuinelucifer:
@jacob-carlborg I believe the tests are now complete. Please review.
.
@ciechowoj I ended up doing a pretty similar thing as you suggested. I had to change the existing functions though to check output files in exact path when multiple input files are given...

ciechowoj:
I believe it would be better to have list of tuples that two lists - easier to match output to input file when there is more files.

test_files/multiThreadTest2.h
```diff
@@ -0,0 +1,4 @@
+int retme()
+{
+	return 2;
```
jacob-carlborg: Please use spaces for indentation.

jacob-carlborg: I recommend not having any body.

unit_tests/Common.d
```diff
@@ -253,29 +253,61 @@ string[] findExtraGNUStepPaths(string file, size_t line)
...
 
     string outputDir = namedTempDir("dstepUnitTest");
-    string outputPath = buildPath(outputDir, baseName(expectedPath));
+
+    string[] outputPaths;
+    if(expectedPaths.length == 1)
+        outputPaths ~= buildPath(outputDir, baseName(expectedPaths[0]));
+    else
+    {
+        foreach(expectedPath; expectedPaths)
```
jacob-carlborg: Please put a space after `foreach`, `if` and similar.

jacob-carlborg:
@ciechowoj yeah, I recommend a list of tuples as well.

genuinelucifer:
@ciechowoj @jacob-carlborg I changed the lists to list of tuples.
Please review.
.
Also, should I squash the commits to one?

test_files/multiThreadTest1.h
```diff
@@ -0,0 +1,3 @@
+void func1()
```
jacob-carlborg: I recommend not having a body and `void` as the parameter.

jacob-carlborg:
Yes, squash the commits. I also don't see any tests for the behavior of the `-o` flag. That if there's multiple input files the `-o` flag should create a directory where to put the resulting files and if there's only one input file it should be have as it currently does, naming the output file.

test_files/multiThreadTest1.h
genuinelucifer: > I recommend not having a body and void as the parameter.
>

Oh ok. I thought u wanted empty body. I'll change it.

genuinelucifer:
> I also don't see any tests for the behavior of the -o flag.

In assertRunsDStep function, if there is one file, it passes outputPath for
output file. Else it passes outputDir.

> the -o flag should create a directory where to put the resulting files

Well, currently it requires that the directory should exist else it fails.
Also, if a filename is passed along with multiple input files then that
filename is choosen for the first file and rest of the filenames are made
by using  .d as extension.
I'll change this behavior to always create the output directory.

jacob-carlborg:
If the user passes multiple files and the -o flag, something like `dstep foo.h bar.h -o result.d` then the directory would be named `result.d`.

test_files/multiThreadTest1.h
jacob-carlborg: No. I mean `void func1(void);`. DStep currently doesn't handle the body of a function. It don't want there to be any confusion.

unit_tests/Common.d
```diff
@@ -265,17 +265,38 @@ void assertRunsDStep(
...
+    {
+        assertFileExists(filesPath[0], file, line);
+        assertFileExists(filesPath[1], file, line);
+        actualPaths ~= filesPath[1];
+    }
 
     string outputDir = namedTempDir("dstepUnitTest");
-    string outputPath = buildPath(outputDir, baseName(expectedPath));
+
+    string[] outputPaths;
```
jacob-carlborg: Why are there multiple output paths?

genuinelucifer: One for each file.
An output path is complete location of a file which is checked for the
test.

genuinelucifer:
@jacob-carlborg I suquashed the commits and incorporated the function to always create output directory when multiple input files are passed.

dstep/driver/Application.d
```diff
@@ -186,22 +160,108 @@ private:
...
+            this.argsToRestore = argsToRestore.dup;
+            this.arguments = arguments;
+            this.compiler = compiler;
+            this.compilerArgs = compilerArgs.dup;
+            
+            inputFile = fileName;
+
+            if(outputDir != null)
+                outputFile = buildPath(outputDir, defaultOutputFilename());
+            else if(!createOutputFileName && arguments.output != "")
```
jacob-carlborg: There are still several cases when there's a missing space between an `if` and the opening parenthesis.

dstep/driver/Application.d
```diff
@@ -186,22 +160,108 @@ private:
         return remainingArgs ~ extraArgs;
     }
 
-    bool handleDiagnostics ()
+    struct ParseFile
     {
-        bool translate = true;
+        private
+        {
+            string inputFile, outputFile, outputDirectory;
```
jacob-carlborg: Please have only one variable declaration per each line.

genuinelucifer:
@jacob-carlborg I did the changes and squashed it back. But now this PR has merge conflicts. I'll try to rebase and fix the conflicts in a few hours.

jacob-carlborg:
Yeah, I merged another big PR.

genuinelucifer:
@jacob-carlborg I have rebased the PR. Please consider merging if it's acceptable now.

jacob-carlborg:
:ok_hand:

jacob-carlborg:
Actually, a couple of minor things:

1. Could you please change the commit message. The feature is not parallel processing of input files, but rather multiple input files in the first place. I don't think "Enforcing" sound good in this case

1. Update the changelog 
1. Update the help text/usage information to somehow indicate that multiple input files are supported. Something like `<input files...>`

genuinelucifer:
Is it okay now?

jacob-carlborg:
With the "help text" i meant the help text at the bottom of Application.d that is printed when the `--help` flag is given. But it was good that you update the readme as well.

genuinelucifer:
> With the "help text" i meant the help
> text at the bottom of Application.d that
> is printed when the --help flag is given.
> But it was good that you update the
> readme as well.

I saw that after pushing the changes. Is it complete now? I changed some
text in the readme as well.

jacob-carlborg:
Looking good now, I'll wait with merging until the tests are complete, just to be sure.

jacob-carlborg:
Thanks.

genuinelucifer:
Although, i would take a couple of days break for a few other things.
But, it would be great if you could point me to some other issue/feature
that I could work on next.

jacob-carlborg:
@genuinelucifer I think a natural extension to this PR would be to be able to pass a directory to DStep, which scans the directory for files to translate.

genuinelucifer:
> scans the directory for files to translate.

Ok. So a '-i <directory>` to check for all the  '.h' files and then
translate them?
Or without a new flag. Simply check if inout file name is indeed a
directory and if yes then translate all files in it. In this case if a file
also exists with same name then that could be given priority.

Which one would you prefer?

jacob-carlborg:
> Which one would you prefer?

The second approach. It's not possible to have a directory and a file with the same name.

genuinelucifer:
Oops. I thought it might be possible in linux. Never tried.

jacob-carlborg:
An alternative approach could be to allow this: `dstep foo/*`, and for recursively scan a directory: `dstep foo/**/*`. The first should already be handled by the shell but I don't think the second is handled by the shell. Also, the shell does no globbing on Windows.

genuinelucifer:
> the shell does no globbing on Windows.
>

So, i should hard code the recursive search for both cases? Because i could
never make dstep work on Windows and hence i won't be able to test it. And
build bots are for mac and linux.
I guess it shouldn't be too hard. I'll try and submit a preliminary PR
probably within 1-2 days.

jacob-carlborg:
Never mind the globbing, the shell already does that.

genuinelucifer:
Oh great. :)

jacob-carlborg:
But I still think it's convenient to be able to pass a whole directory to translate. That should work on Windows as well, if we ever get DStep to work on Windows.

genuinelucifer:
Hmm. Right. Currently we only translate header files right? No source
files.

jacob-carlborg:
Yes, only header files.

jacob-carlborg:
I think we can assume `.h` for all files. If the user wants to translate some other extension they can still use globbing or similar.

genuinelucifer:
Oh hmm. Great.

jacob-carlborg:
I noticed one thing that might be a problem with the current implementation of multiple files. If I have a file structure looking like this:
```
foo
├── bar
│   └── bar.h
└── foo
    └── foo.h
```
And run DStep like `dstep foo/bar/bar.h foo/foo/foo.h -o bar` then the resulting file structure will look like this:

```
bar
├── bar.d
└── foo.d
```
Looks like all files are placed in the top level of the target directory. That will cause problems if `bar` and `foo` in the source file structure have files with the same names. Then these files will overwrite each other in the target directory.

genuinelucifer:
Oh hmm. But isn't it that the user should avoid? Translating files with
same name at the same time?
I mean even if we copy the directory structure then too it could be
dificult (or undesired) in some situations.
Probably a new flag for this?

jacob-carlborg:
> Oh hmm. But isn't it that the user should avoid? Translating files with same name at the same time?

I think it's a valid use case when using globbing and what I suggested as a new feature today, to handle a whole directory as input argument.

> I mean even if we copy the directory structure then too it could be dificult (or undesired) in some situations.

I think it's undesired for a single input file. But for multiple input files, I'm not sure. Maybe a new flag for this.

PR #60, opened 2016-04-07
Rest of cucumber tests.

ciechowoj:
PR removes rest of cucumber tests and adds equivalent unit tests.


clang/Util.d
```diff
@@ -80,8 +80,17 @@ mixin template CX ()
...
 
 extern (C) int mkstemps(char*, int);
+extern (C) char* mkdtemp(char*);
 extern (C) int close(int);
 
+private void randstr (char[] slice)
+{
+    import std.random;
+
+    for (uint i = 0; i < slice.length; ++i)
```
jacob-carlborg: `foreach`? Or is it even possible to use array operations, something like: `slice[] = uniform!("[]")('A', 'Z');`, and skip the loop completely.

clang/Util.d
```diff
@@ -112,7 +122,35 @@ File namedTempFile(string prefix, string suffix)
     scope (exit) close(fd);
 
     if (fd == -1)
-        throw new NamedTempFileException("Cannot create \"%s\" temporary file.".format(path));
+    {
+        auto message = format("Cannot create temporary file \"%s\".", path);
```
jacob-carlborg: Please store the path as an instance variable instead of in the message.

unit_tests/FileTests.d
```diff
@@ -149,3 +149,14 @@ unittest
...
+
+unittest
+{
+    assertRunsDStepCFile(
+        "test_files/aggregate.d",
+        "test_files/aggregate.h");
+
+    assertRunsDStepObjCFile(
+        "test_files/objc/methods.d",
+        "test_files/objc/methods.h");
```
jacob-carlborg: These tests are more like integration tests. Perhaps they should have a separate directory?

clang/Util.d

clang/Util.d
ciechowoj: I'm not sure if by instance variable you mean "pass the result directly as the argument", but if is so it is fixed. I sometimes add such an auxiliary variables to shorten the line length and avoid unnecessary nesting in function argument list.

unit_tests/FileTests.d
ciechowoj: Fixed.

clang/Util.d
jacob-carlborg: No, I mean to store the path as an instance variable in `NamedTempDirException`. It should be possible to retrieve the path without having to parse the message.

clang/Util.d
jacob-carlborg: Hmm, when I added the comment the diff showed that `randstr` was added. Now the diff shows that `randstr` is removed. How is that possible?

ciechowoj: I added `randstr` only to fill the range with random characters, however as you showed me a way to do it directly with `uniform`, I removed `randstr` and used uniform directly.

clang/Util.d

clang/Util.d
Dicebot: I don't think array slice assignment of uniform works like that. Most likely it will call RNG once and use the same value for all elements. Needs checking at least.

jacob-carlborg: Yeah, you're right. It will only call `uniform` once.

clang/Util.d
```diff
@@ -80,14 +80,32 @@ mixin template CX ()
...
 {
-    this (string message, string file = __FILE__, size_t line = __LINE__)
+    this (string path, string file = __FILE__, size_t line = __LINE__)
     {
-        super(message, file, line);
+        this.path = path;
+        super(format("Cannot create temporary file \"%s\".", path), file, line);
     }
+
+    public string path;
```
jacob-carlborg: Please always put instance variables at the top, before all methods and constructors. `public` is also not necessary. Set it to const or immutable as well to make it read only.

unit_tests/Common.d
```diff
@@ -77,6 +101,24 @@ class TranslateAssertError : AssertError
     }
 }
 
+string readFile(string path)
+{
+    import std.stdio : File;
+    auto file = File(path, "r");
+    char[] buffer = new char[file.size];
+    return file.rawRead(buffer).idup;
+}
```
jacob-carlborg: Any reason not to use `std.file.read`?

clang/Util.d

unit_tests/Common.d

clang/Util.d

jacob-carlborg:
Instead of the latest commit, cb75f3cdfe8a2679c674ba983f0ef1cfaa11240e, could you please amend the changes to the commit(s) instead.

ciechowoj:
Done.

jacob-carlborg:
:ok_hand:

ciechowoj:
Thanks : ) .

PR #62, opened 2016-05-14
Issue 47

ciechowoj:
Adds support for macro-definition constants in array sizes. E.g.

    #define FOO = 1
    char a[FOO];

translates (more or less) to

    enum FOO = 1
    char[FOO] a;

### DIFF
--- a/clang/SourceLocation.d
+++ b/clang/SourceLocation.d
@@ -57,13 +57,13 @@ struct SourceLocation
         return format("SourceLocation(file = %s, line = %d, column = %d, offset = %d)", s.file, s.line, s.column, s.offset);
     }
 
-    static bool lexicalLess(in SourceLocation a, in SourceLocation b)
+    bool lexicalLess(in SourceLocation that)
     {
         File fileA, fileB;
         uint offsetA, offsetB;
 
-        clang_getSpellingLocation(a.cx, &fileA.cx, null, null, &offsetA);
-        clang_getSpellingLocation(b.cx, &fileB.cx, null, null, &offsetB);
+        clang_getSpellingLocation(cx, &fileA.cx, null, null, &offsetA);
+        clang_getSpellingLocation(that.cx, &fileB.cx, null, null, &offsetB);
 
         return fileA != fileB ? fileA < fileB : offsetA < offsetB;
     }
--- a/clang/SourceLocation.d
+++ b/clang/SourceLocation.d
@@ -31,6 +31,15 @@ struct SourceLocation
         return spell;
     }
 
+    @property Spelling expansion() const
+    {
+        Spelling spell;
+
+        clang_getExpansionLocation(cx, &spell.file.cx, &spell.line, &spell.column, &spell.offset);
+
+        return spell;
+    }
+
     @property size_t offset() const
     {
         return spelling.offset;
@@ -46,5 +55,16 @@ struct SourceLocation
         import std.format: format;
         auto s = spelling;
         return format("SourceLocation(file = %s, line = %d, column = %d, offset = %d)", s.file, s.line, s.column, s.offset);
+    }
+
+    static bool lexicalLess(in SourceLocation a, in SourceLocation b)
+    {
+        File fileA, fileB;
+        uint offsetA, offsetB;
+
+        clang_getSpellingLocation(a.cx, &fileA.cx, null, null, &offsetA);
+        clang_getSpellingLocation(b.cx, &fileB.cx, null, null, &offsetB);
+
+        return fileA != fileB ? fileA < fileB : offsetA < offsetB;
     }
 }
--- a/clang/Type.d
+++ b/clang/Type.d
@@ -129,6 +129,15 @@ struct Type
     {
         return ArrayType(this);
     }
+
+    @property bool isArray ()
+    {
+        return
+            kind == CXTypeKind.CXType_ConstantArray ||
+            kind == CXTypeKind.CXType_IncompleteArray ||
+            kind == CXTypeKind.CXType_VariableArray ||
+            kind == CXTypeKind.CXType_DependentSizedArray;
+    }
 }
 
 struct FuncType
@@ -158,6 +167,12 @@ struct ArrayType
     Type type;
     alias type this;
 
+    this (Type type)
+    {
+        assert(type.isArray);
+        this.type = type;
+    }
+
     @property Type elementType ()
     {
         auto r = clang_getArrayElementType(cx);
@@ -167,6 +182,20 @@ struct ArrayType
     @property long size ()
     {
         return clang_getArraySize(cx);
+    }
+
+    @property size_t numDimensions ()
+    {
+        size_t result = 1;
+        auto subtype = elementType();
+
+        while (subtype.isArray)
+        {
+            ++result;
+            subtype = subtype.array.elementType();
+        }
+
+        return result;
     }
 }
 
--- a/dstep/translator/MacroIndex.d
+++ b/dstep/translator/MacroIndex.d
@@ -15,36 +15,31 @@ import clang.SourceLocation;
 import clang.SourceRange;
 import clang.TranslationUnit;
 
-class InvalidArgumentError : object.Error
-{
-    this (string message, string file = __FILE__, ulong line = __LINE__)
-    {
-        super(message, file, line);
-    }
-}
-
 class MacroIndex
 {
     private static bool pred(in Cursor a, in Cursor b)
     {
-        return SourceLocation.lexicalLess(a.location, b.location);
+        return a.location.lexicalLess(b.location);
     }
 
     private TranslationUnit unit;
-    private alias CursorRedBlackTree = RedBlackTree!(Cursor, (in Cursor a, in Cursor b) => pred(a, b));
+    private alias CursorRedBlackTree =
+        RedBlackTree!(Cursor, (a, b) => pred(a, b));
     private CursorRedBlackTree expansions;
     private Cursor[string] definitions;
 
     private static string uniqueID(in Cursor cursor)
     {
         import std.format : format;
-        return "%s@%s:%d".format(cursor.spelling, cursor.location.path, cursor.location.offset);
+        return format(
+            "%s@%s:%d",
+            cursor.spelling,
+            cursor.location.path,
+            cursor.location.offset);
     }
 
     this(TranslationUnit unit)
     {
-        import std.stdio;
-
         this.unit = unit;
 
         expansions = new CursorRedBlackTree();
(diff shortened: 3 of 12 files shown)


ciechowoj:
Making this PR I found a bug with the order of array dimension sizes. The tests `array.h/array.d` are broken. I'm not sure how fast I can fix it, so @jacob-carlborg, you can review and accept this PR, or wait and I include the fix for the issue here when it is done.

clang/SourceLocation.d
```diff
@@ -47,4 +56,15 @@ struct SourceLocation
         auto s = spelling;
         return format("SourceLocation(file = %s, line = %d, column = %d, offset = %d)", s.file, s.line, s.column, s.offset);
     }
+
+    static bool lexicalLess(in SourceLocation a, in SourceLocation b)
```
jacob-carlborg: Why is this a static method?

unittest.d
```diff
@@ -1,4 +1,7 @@
+import std.stdio;
+import Common;
```
jacob-carlborg: Are the imports there for a reason?

jacob-carlborg:
> Making this PR I found a bug with the order of array dimension sizes.

Could you please explain the bug?

clang/SourceLocation.d
ciechowoj: It is used as predicate in some another place, I prefer predicated to be binary functions, but I suppose I can make a symmetry and make it a usual method.

ciechowoj:
int a[4];
    int b[2][3];
    int c[2][3][4];

    extern const char *const sys_errlist[];

translates to

	extern (C):

	extern __gshared int[4] a;
	extern __gshared int[3][2] b;
	extern __gshared int[4][3][2] c;
	extern __gshared const char*[] sys_errlist;

pay attention to array sizes.

unittest.d
ciechowoj: Sorry, I forgot to remove it. Fixed.

clang/SourceLocation.d
jacob-carlborg: It's already wrapped in a predicate where it's actually used.

dstep/translator/MacroIndex.d
```diff
@@ -0,0 +1,101 @@
...
+
+class MacroIndex
+{
+    private static bool pred(in Cursor a, in Cursor b)
+    {
+        return SourceLocation.lexicalLess(a.location, b.location);
+    }
+
+    private TranslationUnit unit;
+    private alias CursorRedBlackTree = RedBlackTree!(Cursor, (in Cursor a, in Cursor b) => pred(a, b));
```
jacob-carlborg: I think you can replace the lambda with just `pred`.

clang/SourceLocation.d
ciechowoj: Yes, but the other predicate takes a `Cursor` instead of `SourceLocation`.

dstep/translator/MacroIndex.d
ciechowoj: I wanted to do that but then the compiler complains with this:

    dstep/translator/MacroIndex.d(34,40): Error: template instance std.container.rbtree.RedBlackTree!(Cursor, pred) does not match template declaration RedBlackTree(T, alias less = "a < b", bool allowDuplicates = false) if (is(typeof(binaryFun!less(T.init, T.init))))

And I do not know why really : /

ciechowoj: Wrong, it complains because `pred` is private. I didn't wanted to make it public.

jacob-carlborg: Fair enough, can you remove the parameter types in the lambda?

ciechowoj: Done.

jacob-carlborg:
I was just looking for the documentation to figure out the order D uses for the rows and columns in a multidimensional array. Unfortunately the [only example](https://dlang.org/spec/arrays.html#rectangular-arrays) I could found uses the same number of rows and columns 😞.

ciechowoj:
Wow, that is a surprise. So, there is no a bug, however, hold on with accepting this PR, as replacing the macros doesn't work for multidimensional arrays yet. I planned to fix it together with fixing that nonexistent bug.

dstep/translator/MacroIndex.d
```diff
@@ -0,0 +1,101 @@
...
+        if (!equal.empty)
+            result ~= equal.array;
+
+        foreach (itr; greater)
+        {
+            if (itr.file == cursor.file && itr.location.offset < extent.end.offset)
+                result ~= itr;
+            else
+                break;
+        }
```
jacob-carlborg: Looks like this foreach can be replaced with a filter.

ciechowoj: `filter` would have slightly worse performance, it would need to go to the end of range every time, even if it can finish after finding first instance. I replaced the loop with `until`.

ciechowoj:
Well, I added support for multidimensional arrays.

On another topic (don't want to create separate issues):

@jacob-carlborg - what do you think about removing dependency on mambo, dstack and tango from dstep?

What do you think about replacing all enums like `CXCursorKind.CXCursor_IntegerLiteral` with saner replacement like `CursorKind.IntegerLiteral`. I know there is `with` but I don't consider it nice and current seems to be too verbose.

jacob-carlborg:
> what do you think about removing dependency on mambo, dstack and tango from dstep?

What's the reason to remove those dependencies? One reason I can think of is to make DStep an official D tool. if that's a goal. Any other?

> What do you think about replacing all enums like CXCursorKind.CXCursor_IntegerLiteral with saner replacement like CursorKind.IntegerLiteral. I know there is with but I don't consider it nice and current identifiers seem to be too verbose.

Hmm. We can do something about the identifiers. Just not sure the best why to do this. I see a couple of options:

1. Let DStep automatically to the transformation. The tool Apple is using to import C interfaces to Swift is doing something similar

1. Manually do the transformation

1. Add aliases so the members are accessible without qualifying them with the enum name

There's also the question if this should be done directly in the bindings or in a layer on top of the bindings.

ciechowoj:
> What's the reason to remove those dependencies? One reason I can think of is to make DStep an official D tool. if that's a goal. Any other?

1. As you mentioned.
2. Maintainability - currently if someone wants to contribute to dstep he needs to learn a bit of those libraries.
3. Some things are duplicated in Phobos and mambo/tango and some of that duplicates are inconsistent to each other. This gives me some headaches from time to time.
4. Some features that I would like to implement in dstep would require making changes to dstack.
5. `dub` would have less dependencies to check : )

I'm not saying that I would like to do such a migration at the moment, but I would like to know if I should replace some dependencies of there is an occasion. 

> Hmm. We can do something about the identifiers. Just not sure the best why to do this. I see a couple of options:
>
> 1. Let DStep automatically to the transformation. The tool Apple is using to import C interfaces to Swift is doing something similar

This could be controlled by command line switch.

> Manually do the transformation
> 
> Add aliases so the members are accessible without qualifying them with the enum name

Still ugly (enum elements) .

> There's also the question if this should be done directly in the bindings or in a layer on top of the bindings.

So, I would like to add that prettier enums to the layer on the top of clang bindings. I'm only asking if you accept a PR with such changes.

ciechowoj:
@Dicebot The prettier enums would be present alongside the original ones but in the higher layer of code. There is a wrapper around clang in the dstep (I suppose you know what I mean) and all names are pretty with exception to that enums, which drives me a bit crazy every time I need to type it : ).

> And switching dstep from to be phobos-only is something I wanted to do long time ago too, just never got to it in priority list. I believe current usage of dstack/tango is disproportional to added contribution/maintenance burden of transitive dependencies.

Agree 100%.

clang/SourceLocation.d
jacob-carlborg: I would still like this to be a non-static method.

ciechowoj: As you wish : )

clang/Type.d
```diff
@@ -12,6 +12,14 @@ import clang.c.Index;
 import clang.Cursor;
 import clang.Util;
 
+class TypeIsNotAnArrayError : object.Error
+{
```
jacob-carlborg: I think the exception should include the type.

dstep/translator/MacroIndex.d
```diff
@@ -0,0 +1,106 @@
...
+                    definitions[uniqueID(cursor)] = *def;
+            }
+            else if (cursor.kind == CXCursorKind.CXCursor_MacroDefinition)
+            {
+                recent[cursor.spelling] = cursor;
+            }
+        }
+    }
+
+    Cursor queryDefinition(in Cursor expansion) const
```
jacob-carlborg: Is this method used?

clang/Type.d
ciechowoj: Changed.

dstep/translator/MacroIndex.d
ciechowoj: No, I intended to use it in future - removed.

clang/Type.d
jacob-carlborg: No. But on the other hand, can we use an assert instead? It's a logical error to call `numDimensions` on something that isn't an array?

clang/Type.d
```diff
@@ -129,6 +138,36 @@ struct Type
...
+    @property bool isArray ()
+    {
+        return
+            kind == CXTypeKind.CXType_ConstantArray ||
+            kind == CXTypeKind.CXType_IncompleteArray ||
+            kind == CXTypeKind.CXType_VariableArray ||
+            kind == CXTypeKind.CXType_DependentSizedArray;
+    }
+
+    @property size_t numDimensions ()
```
jacob-carlborg: I think this belongs to `ArrayType`.

ciechowoj: You are right, now it's better.

clang/Type.d
```diff
@@ -127,8 +127,18 @@ struct Type
 
     @property ArrayType array ()
     {
+        assert(isArray);
```
jacob-carlborg: I think it's better to put the assert in a constructor in `ArrayType`. There's nothing from stopping anyone to use `ArrayType` directly without going through `array`.

ciechowoj: Good observation.

dstep/translator/MacroIndex.d
```diff
@@ -0,0 +1,97 @@
...
+
+import std.container.rbtree;
+
+import clang.c.Index;
+import clang.Cursor;
+import clang.SourceLocation;
+import clang.SourceRange;
+import clang.TranslationUnit;
+
+class InvalidArgumentError : object.Error
```
jacob-carlborg: This seems not to be used.

ciechowoj: Removed.

jacob-carlborg:
If you prefix one of the commit messages with `Fix #47: ...` GitHub will automatically close the issue. Could you please update the changelog as well.

ciechowoj:
Done.

jacob-carlborg:
👌

ciechowoj:
Thank you.

ciechowoj:
> @ciechowoj regarding enum members - changing them to prettier may conflict with eventual goal of generating clang bindings with dstep itself (unless the hypothetical namespace stripping support is implemented too).

@Dicebot could you elaborate? Especially what do you mean by 'namespace stripping'?

PR #63, opened 2016-05-24
Comments

ciechowoj:
Adds support for global comments.

The PR extends an `Output` class to add comments between translated items. The `Output` class is given an instance of `CommentIndex` class, which can report comments from given range of code.

The `Output` class remembers the location of the recent cursor (translated item) and when the next item is added it asks `CommentIndex` for comments (in between them) and inserts them in between considered cursors (methods that adds items (`singleLine` etc.) were extended with a parameter to pass their locations to `Output` class and there is a `flushLocation` family of methods for that purpose too). The `Output` takes some care to format the comments correctly (identation, spacing, etc.).

The second part of the PR is fixing the order of preprocessor related cursors, as they are reported by default at the very beginning of the list returned by `TranslationUnit.all` (shame I didn't noticed it when doing translation of macro constants). PR implements another visitor that interleaves preprocessor and normal cursor in proper order, it is returned by `TranslationUnit.allInOrder` method.


jacob-carlborg:
There's a lot of commits here. Are the commits individually self contained so I can review them one by one?

ciechowoj:
I believe it would be better if you review all the changes at once. The commits are more or less self contained, but sometimes the have things from other ones.

jacob-carlborg:
In basically every pull request you've made I've mentioned that I would like to have self contained commits that are well organized. It's much easier for both you and me if the commits are in that shape from the beginning.

ciechowoj:
> In basically every pull request you've made I've mentioned that I would like to have self contained commits that are well organized. It's much easier for both you and me if the commits are in that shape from the beginning.

I'll behave.

> Rebase and amend are you're friends and with practice you'll improve.

The problem is that when I make some pull request, then do some work based on in on another branch and then modify the original branch the rebase becomes nightmare : ( .

Anyway I'll put more effort to improve the quality of my commits.

jacob-carlborg:
> The problem is that when I make some pull request, then do some work based on in on another branch and then modify the original branch the rebase becomes nightmare : ( .

Yeah, that can be problematic. Just make sure you rebase the original branch in your new branch when you update the new branch.

ciechowoj:
I cleaned up the commits.

jacob-carlborg:
Thanks, I'll take a look.

jacob-carlborg:
Is the comment introspection API [1]  only for documentation style comments (doxygen)?

[1] http://clang.llvm.org/doxygen/group__CINDEX__COMMENT.html

ciechowoj:
How is reviewing going?

jacob-carlborg:
Sorry, I've been quite busy with non D related things.

clang/TranslationUnit.d
```diff
@@ -113,6 +115,139 @@ struct TranslationUnit
...
+            {
+                auto ptr = cursor.path in stacked;
+
+                if (ptr !is null && *ptr)
+                {
+                    while (locationStack[$-1].path != cursor.path)
+                    {
+                        stacked[locationStack[$-1].path] = false;
+                        locations ~= locationStack[$-1];
+                        locationStack = locationStack[0..$-1];
```
jacob-carlborg: Please use spaces around operators.

clang/TranslationUnit.d
```diff
@@ -113,6 +115,139 @@ struct TranslationUnit
...
+
+    SourceRange extent (uint startOffset, uint endOffset)
+    {
+        CXFile file = clang_getFile(cx, spelling.toStringz);
+        auto start = clang_getLocationForOffset(cx, file, startOffset);
+        auto end = clang_getLocationForOffset(cx, file, endOffset);
+        return SourceRange(clang_getRange(start, end));
+    }
+
+    package SourceLocation[] includeLocationsImpl(Range)(Range cursors)
```
jacob-carlborg: What is this function used for?

ciechowoj: In essence it returns list of locations of include directives.

clang/TranslationUnit.d
```diff
@@ -113,6 +115,139 @@ struct TranslationUnit
...
+            }
+        }
+
+        Entry[][string] map;
+
+        uint index = 0;
+        foreach (location; locations)
+        {
+            map[location.path] ~= Entry(index, location);
+            ++index;
```
jacob-carlborg: Is there a reason why the index in the foreach loop is not used?

clang/TranslationUnit.d
ciechowoj: It is used in the next function that returns a comparator that allows to compare locations of cursors globally, across different include files. It is needed because libclang report all preprocessor cursors at the beginning and I needed to reorder them to correctly translate them.

ciechowoj: The comparator is used in `InOrderVisitor` from `Visitor.d`.

jacob-carlborg: Yeah, that's what the name says 😃. I have to admit I feel a bit lost here. Perhaps you could write a summary of what's going on in this PR in the conversation tab.

ciechowoj: Ok, I'll do it in a moment it the first post.

dstep/translator/CommentIndex.d
```diff
@@ -0,0 +1,124 @@
...
+                auto location = token.location;
+
+                if (location.isFromMainFile)
+                {
+                    Comment comment = {
+                        token.spelling,
+                        token.extent,
+                        location.line,
+                        location.column,
+                        location.offset };
```
jacob-carlborg: Please avoid this initialization syntax. Either use `Comment(...)` or `{ key: value}`.

dstep/translator/CommentIndex.d
```diff
@@ -0,0 +1,124 @@
...
+    bool isHeaderCommentPresent()
+    {
+        return !comments.empty && comments.front.offset == 0;
+    }
+
+    SourceRange queryHeaderCommentExtent()
+    {
+        return translUnit.extent(
+            comments.front.offset,
+            comments.front.offset + cast (uint) comments.front.content.length);
```
jacob-carlborg: Don't put a space after `cast`.

dstep/translator/Record.d
```diff
@@ -29,61 +29,70 @@ class Record (Data) : Declaration
...
-            this.recordDefinitions[cursor] = true;
-        else
-            context.isFwdDeclaration = true;
-
-        context.name = translateIdentifier(name);
-
-        dg(context);
-
-        output.output(context.data);
+        output.singleLine("struct %s;", spelling);
```
jacob-carlborg: Should `type` be used here as well instead of `struct`?

dstep/translator/Record.d
```diff
@@ -29,61 +29,70 @@ class Record (Data) : Declaration
...
 
-    void translateVariable (Cursor cursor, Data context)
+    private void translateVariable (Output output, Cursor cursor)
     {
-        Output output = new Output();
         translator.variable(output, cursor);
-        context.instanceVariables ~= output;
+    }
+
+    private string type ()
```
jacob-carlborg: Is there a better name than `type` for this? I don't want to confuse it with a cursor type.

clang/TranslationUnit.d
ciechowoj: Ok, wrote the summary.

test_files/aggregate.d
```diff
@@ -18,4 +19,5 @@ struct Bar
 }
 
 extern __gshared const(char)* q;
-extern __gshared int function (int, int) e;
\ No newline at end of file
+
+extern __gshared int function (int, int) e;
```
jacob-carlborg: Is this now keeping the newlines from the C header file?

clang/TranslationUnit.d
ciechowoj: Around `..` too? Like `[prefixSize..$]` -> `[prefixSize .. $]`.

test_files/aggregate.d
ciechowoj: Yes.

unit_tests/CommentUnitTests.d
```diff
@@ -0,0 +1,243 @@
...
+
+// Test header comment handling.
+unittest
+{
+	assertTranslates(
+q"C
+/* Header comment. */
+
+/* Comment before variable. */
+int variable;
```
jacob-carlborg: What is considered a header comment? If `/* Comment before variable. */` is removed, is `/* Header comment. */` still a header comment?

test_files/aggregate.d
ciechowoj: As `Output` class is getting info about locations of translated cursors, it tries to keep the newlines. I believe that feature is a bit improved in the subsequent PRs.

jacob-carlborg: Does it remove excessive newlines as well?

jacob-carlborg:
Are `//` comments supported?

ciechowoj:
> Are // comments supported?

Very likely, however I'm focusing on C, didn't done the tests.

> Does it support comments inside declarations, i.e. void foo(/*const*/ int a);?

No at the moment. Only global comments are handled correctly. There is some preliminary work in subsequent PRs to add such support.

clang/TranslationUnit.d
ciechowoj: No, fixed. Good to know that such a feature of D exists.

unit_tests/CommentUnitTests.d
ciechowoj: The header comment is the comment in the first line, if there is no such comment (e.g. fist line empty) there is no header comment. 

> If /* Comment before variable. */ is removed, is /* Header comment. */ still a header comment?

Yes.

test_files/aggregate.d
ciechowoj: Yes.

unit_tests/OutputUnitTests.d
```diff
@@ -511,3 +513,307 @@ unittest
...
+    CommentIndex index = makeCommentIndex(
+q"C
+/* This is header comment.
+ * Aaaaaaaaaaaaaaaaaaaaaaa.
+ * Aaaaaaaaaaaaaa. Aaaaaa.
+ */
+
+int func(int x);
+
+class A { };
```
jacob-carlborg: `class`?

dstep/translator/CommentIndex.d
ciechowoj: Fixed.

clang/TranslationUnit.d
jacob-carlborg: I prefer that.

dstep/translator/CommentIndex.d
ciechowoj: Fixed.

clang/TranslationUnit.d
```diff
@@ -113,6 +115,135 @@ struct TranslationUnit
...
+            {
+                auto ptr = cursor.path in stacked;
+
+                if (ptr !is null && *ptr)
+                {
+                    while (locationStack[$ - 1].path != cursor.path)
+                    {
+                        stacked[locationStack[$-1].path] = false;
+                        locations ~= locationStack[$-1];
+                        locationStack = locationStack[0..$-1];
```
jacob-carlborg: Here as well, spaces around operators.

clang/TranslationUnit.d
jacob-carlborg: 😃

test_files/aggregate.d
jacob-carlborg: Awesome 😃

unit_tests/CommentUnitTests.d
jacob-carlborg: Ok, I see. Perhaps worth adding documentation to the `isHeaderCommentPresent` method to describe that

unit_tests/OutputUnitTests.d
ciechowoj: Fixed.

jacob-carlborg:
> Very likely, however I'm focusing on C, didn't done the tests.

It's supported in C99 [1] and I have yet to find a compiler that doesn't support them. Even when compiling with `clang -std=c89` single line comments are supported.

[1] https://en.wikipedia.org/wiki/C99

unit_tests/OutputUnitTests.d
```diff
@@ -511,3 +513,307 @@ unittest
...
+// block-statement, even if there is extra space in the original.
+unittest {
+    CommentIndex index = makeCommentIndex(
+q"C
+
+int func(int x);
+
+
+struct A {
+    void method();
```
jacob-carlborg: This does not compile for me using Clang.

jacob-carlborg:
Is there functionality to get include directives just for completeness? Because I don't think it's useful to translate include directives.

unit_tests/OutputUnitTests.d
ciechowoj: > It's supported in C99 [1] and I have yet to find a compiler that doesn't support them. Even when compiling with clang -std=c89 single line comments are supported.

[1] https://en.wikipedia.org/wiki/C99

You say it, you got it. Added appropriate tests.

ciechowoj: > This does not compile for me using Clang.

Hmm, right. Probably clang is run in C++ mode...

ciechowoj:
> Is the functionality to get include directives just for completeness? Because I don't think it's useful to translate include directives.

What particular functionality, sorry I don't understand. 

If it goes about improved handling of includes there is some work in `packages` PR.

jacob-carlborg:
> What particular functionality? Sorry I don't understand. 

`includeLocations` in `TranslationUnit`, but it's only the locations.

unit_tests/OutputUnitTests.d
ciechowoj: I removed it to a field `int field;`... However, it doesn't do much harm.

ciechowoj:
> No at the moment. Only global comments are handled correctly. There is some preliminary work in subsequent PRs to add such support.

I meant global comments, comments inside structs and enums.

unit_tests/OutputUnitTests.d
jacob-carlborg: Cool. I don't what it to look like C++ is supported when it's not.

unit_tests/OutputUnitTests.d
```diff
@@ -511,3 +513,307 @@ unittest
...
+// Test saving header position.
+unittest
+{
+    CommentIndex index = makeCommentIndex(
+q"C
+/* This is header comment.
+ * Aaaaaaaaaaaaaaaaaaaaaaa.
+ * Aaaaaaaaaaaaaa. Aaaaaa.
+ */
+
```
jacob-carlborg: What will happen if there's not newline here? Will it be a header comment or a comment for the function?

ciechowoj:
> What particular functionality? Sorry I don't understand.
includeLocations in TranslationUnit, but it's only the locations.

Oh, now I see. It is only for completeness. It is a bit more efficient for private functions to call `includeLocationsImpl` directly.

unit_tests/OutputUnitTests.d
ciechowoj: Header comment.

ciechowoj: Actually, current implementation doesn't comprehend something like `comment for the function`. It only tries to match the source code layout as close as possible. The header comment is the only exception, I added it because commonly used license comment looks silly placed after `extern (C):` and import block.

jacob-carlborg: Ok, I see.

jacob-carlborg:
Are there any future plans to translate Doxygen comments to Ddoc?

dstep/translator/Record.d
ciechowoj: Yes... do you have another idea? I do not like it much too, but I spend a several seconds thinking about that and didn't come up with anything better.

ciechowoj:
No, in part because I didn't have such a idea. However I may add that feature to the end my list of features : ).

dstep/translator/Record.d
jacob-carlborg: I have another similar tool where I call it `typeName`, but I'm not satisfied with that either.

jacob-carlborg:
> No, in part because I didn't have such a idea. However I may add that feature to the end my list of features : ).

That should hopefully be a bit easier since libclang has an API for that: http://clang.llvm.org/doxygen/group__CINDEX__COMMENT.html.

ciechowoj:
Yes, this API should be useful for implementing that.

unit_tests/CommentUnitTests.d
ciechowoj: Added documentation.

dstep/translator/CommentIndex.d
```diff
@@ -0,0 +1,129 @@
...
+            return lower.upperBound(begin - 1);
+    }
+
+    SourceLocation queryLastLocation()
+    {
+        return lastTokenRange.end;
+    }
+
+    /**
+     * Returns true if a header comment is present.
```
jacob-carlborg: Please separate the first line from the second two lines. In Ddoc, the first paragraph is the summary of the symbol and should be kept short.

dstep/translator/Record.d
ciechowoj: What about `typeKeyword` ?

jacob-carlborg:
I know there's almost no documentation in DStep currently, but in the future I would like to see some more documentation.

Please update the change log.

dstep/translator/CommentIndex.d
ciechowoj: Done.

dstep/translator/Record.d
jacob-carlborg: Sounds good.

ciechowoj:
Done (the second thing).

dstep/translator/Record.d
ciechowoj: Fixed.

dstep/translator/Record.d
ciechowoj: Done.

ciechowoj:
I'm going go to sleep, if you find more issues, I'll fix them tomorrow.

dstep/translator/CommentIndex.d
```diff
@@ -0,0 +1,130 @@
...
+
+    private TranslationUnit translUnit;
+    private Comment[] comments;
+    private SourceRange lastTokenRange;
+
+    this(TranslationUnit translUnit)
+    {
+        this.translUnit = translUnit;
+
+        lastTokenRange = translUnit.cursor.extent;
```
jacob-carlborg: Would this be the same as `translUnit.cursor.tokens.back.extent`?

dstep/translator/CommentIndex.d
```diff
@@ -0,0 +1,130 @@
...
+
+        foreach (token; translUnit.cursor.tokens)
```
jacob-carlborg: Looks like this loop can be converted to two filters and a map.

ciechowoj: You are right, fixed.

jacob-carlborg:
Looking good 👌. Anything else or is it ready to be merged?

ciechowoj:
Ready.

jacob-carlborg:
Thanks.

ciechowoj:
Great : )!

Issue #65, opened 2016-06-01
Handling advanced pre-processor macros.


ciechowoj:
I'm wondering how to handle 'advanced' pre-processor stuff. I have two problems. The first problem is libclang only parses an active branch of pre-processor condition. E.g. 

```C
#define FOO

#ifdef FOO
// This will appear in AST.
#else
// This won't appear in AST.
#endif
```

I suppose it isn't possible to implement translation of pre-processor that handles all possible cases, but most common cases should be doable. The only idea I have is to load the content of file we are translating and parse macros manually (possibly using clang to parse the code *inside* ifdefs) to some AST form, suitable for further processing. This should be perfectly doable. The next problem is a design/implementation one.

Currently there is a thin wrapper around c libclang API. I think it would be nice if pre-processor information (that we extract from file) are available through the same API. What I propose is to extend the classes (structs) like `Cursor`, `SourceLocation`, `Token` etc. in a way that they can contain data created by the user of the thin wrapper (currently an instances of that structs can only be created by libclang). Then, one could iterate through AST containing all the information in a transparent way, without worrying if the data comes from libclang or from the pre-processor parser. For example I would like `Cursor` to have a constructor that allows to create an instance of it from user-supplied data. E.g.

```D
struct Cursor
{
    this(string spelling, CXCursorKind kind, SourceRange, ...);

    /+ ... +/
}
```
An alternative would be to add separate API and then interleave them in actual translator code, however it seems to be messy.

What do you think about it?

jacob-carlborg:
> The only idea I have is to load the content of file we are translating and parse macros manually (possibly using clang to parse the code inside ifdefs) to some AST form, suitable for further processing. This should be perfectly doable.

I would hope it's possible to using libclang to lex the file to get the tokens of the branch that is not compiled. Then run the tokens through libclang again to do the full parsing and analyze.

There's a problem regardless which method is used. You might get compile errors if you're trying to analyze the branch that is normally not analyzed. Example:

```c
#if _WIN32
#include <windows.h>
void foo(DWORD param); // DWORD is declared in windows.h
#endif
```

Running the above translation on any other platform than windows will fail to compile if the branch is analyzed.

As far as the design. At some point you need to run libclang on the source code/tokens and the wrapper types can be created like today. Or am I missing something?

ciechowoj:
[...]

PR #67, opened 2016-06-08
Defines

ciechowoj:
Implement translation of `functional` macrodefinitions.

### DIFF
--- a/clang/Token.d
+++ b/clang/Token.d
@@ -4,7 +4,6 @@
  * Version: Initial created: Feb 14, 2016
  * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
  */
-
 module clang.Token;
 
 import std.typecons;
--- a/clang/Token.d
+++ b/clang/Token.d
@@ -18,6 +18,20 @@ import clang.Util;
 import clang.Visitor;
 import clang.Cursor;
 
+enum TokenKind
+{
+    punctuation = CXTokenKind.CXToken_Punctuation,
+    keyword = CXTokenKind.CXToken_Keyword,
+    identifier = CXTokenKind.CXToken_Identifier,
+    literal = CXTokenKind.CXToken_Literal,
+    comment = CXTokenKind.CXToken_Comment,
+}
+
+TokenKind toD(CXTokenKind kind)
+{
+    return cast (TokenKind) kind;
+}
+
 struct Token
 {
     private struct Container
@@ -38,7 +52,7 @@ struct Token
         }
     }
 
-    private const RefCounted!Container container;
+    private RefCounted!Container container;
     size_t index;
 
     @property private Container* containerPtr() const
@@ -54,14 +68,16 @@ struct Token
 
     @property string spelling() const
     {
+        import clang.Util : toD;
+
         return toD(clang_getTokenSpelling(
             containerPtr.translationUnit,
             containerPtr.tokens[index]));
     }
 
-    @property CXTokenKind kind() const
+    @property TokenKind kind() const
     {
-        return clang_getTokenKind(containerPtr.tokens[index]);
+        return clang_getTokenKind(containerPtr.tokens[index]).toD;
     }
 
     @property SourceLocation location() const
@@ -89,7 +105,7 @@ struct Token
 
 struct TokenRange
 {
-    private const RefCounted!(Token.Container) container;
+    private RefCounted!(Token.Container) container;
     private size_t begin;
     private size_t end;
 
@@ -106,7 +122,7 @@ struct TokenRange
     }
 
     private this(
-        const RefCounted!(Token.Container) container,
+        RefCounted!(Token.Container) container,
         size_t begin,
         size_t end)
     {
@@ -130,12 +146,12 @@ struct TokenRange
         return begin >= end;
     }
 
-    @property Token front() const
+    @property Token front()
     {
         return Token(container, begin);
     }
 
-    @property Token back() const
+    @property Token back()
     {
         return Token(container, end - 1);
     }
@@ -150,7 +166,7 @@ struct TokenRange
         --end;
     }
 
-    @property TokenRange save() const
+    @property TokenRange save()
     {
         return this;
     }
@@ -160,12 +176,12 @@ struct TokenRange
         return end - begin;
     }
 
-    Token opIndex(size_t index) const
+    Token opIndex(size_t index)
     {
         return Token(container, begin + index);
     }
 
-    TokenRange opSlice(size_t begin, size_t end) const
+    TokenRange opSlice(size_t begin, size_t end)
     {
         return TokenRange(container, this.begin + begin, this.begin + end);
     }
--- a/clang/Token.d
+++ b/clang/Token.d
@@ -29,7 +29,7 @@ enum TokenKind
 
 TokenKind toD(CXTokenKind kind)
 {
-    return cast (TokenKind) kind;
+    return cast(TokenKind) kind;
 }
 
 struct Token
--- a/clang/Type.d
+++ b/clang/Type.d
@@ -6,7 +6,7 @@
  */
 module clang.Type;
 
-import mambo.core._;
+import std.bitmanip;
 
 import clang.c.Index;
 import clang.Cursor;
@@ -14,17 +14,54 @@ import clang.Util;
 
 struct Type
 {
+    static assert(Type.init.kind == CXTypeKind.CXType_Invalid);
+
     mixin CX;
 
-    @property string spelling ()
+    private Type* pointee_;
+
+    mixin(bitfields!(
+        bool, "isConst", 1,
+        bool, "isVolatile", 1,
+        bool, "isClang", 1,
+        uint, "", 5));
+
+    string spelling = "";
+
+    this (CXType cx)
     {
-        auto r = clang_getTypeDeclaration(cx);
-        return Cursor(r).spelling;
+        this.cx = cx;
+        spelling = Cursor(clang_getTypeDeclaration(cx)).spelling;
+        isConst = clang_isConstQualifiedType(cx) == 1;
+        isClang = true;
     }
 
-    @property bool isTypedef ()
+    this (CXTypeKind kind, string spelling)
     {
-        return kind == CXTypeKind.CXType_Typedef;
+        cx.kind = kind;
+        this.spelling = spelling;
+    }
+
+    static Type makePointer(Type pointee)
+    {
+        Type result = Type(CXTypeKind.CXType_Pointer, "");
+        result.pointee_ = new Type();
+        *result.pointee_ = pointee;
+        return result;
+    }
+
+    @property bool isAnonymous ()
+    {
+        return spelling == "";
+    }
+
+    @property bool isArray ()
+    {
+        return
+            kind == CXTypeKind.CXType_ConstantArray ||
+            kind == CXTypeKind.CXType_IncompleteArray ||
+            kind == CXTypeKind.CXType_VariableArray ||
+            kind == CXTypeKind.CXType_DependentSizedArray;
     }
 
     @property bool isEnum ()
@@ -32,45 +69,32 @@ struct Type
         return kind == CXTypeKind.CXType_Enum;
     }
 
-    @property Type canonicalType ()
+    @property bool isExposed ()
     {
-        auto r = clang_getCanonicalType(cx);
-        return Type(r);
-    }
-
-    @property Type pointeeType ()
-    {
-        auto r = clang_getPointeeType(cx);
-        return Type(r);
-    }
-
-    @property bool isValid ()
-    {
-        return kind != CXTypeKind.CXType_Invalid;
+        return kind != CXTypeKind.CXType_Unexposed;
     }
 
     @property bool isFunctionType ()
     {
-        return canonicalType.kind == CXTypeKind.CXType_FunctionProto;
+        return canonical.kind == CXTypeKind.CXType_FunctionProto;
     }
 
     @property bool isFunctionPointerType ()
     {
-        with (CXTypeKind)
-            return kind == CXType_Pointer && pointeeType.isFunctionType;
+        return kind == CXTypeKind.CXType_Pointer && pointee.isFunctionType;
     }
 
     @property bool isObjCIdType ()
     {
         return isTypedef &&
-            canonicalType.kind ==  CXTypeKind.CXType_ObjCObjectPointer &&
+            canonical.kind == CXTypeKind.CXType_ObjCObjectPointer &&
             spelling == "id";
     }
 
     @property bool isObjCClassType ()
     {
         return isTypedef &&
-            canonicalType.kind ==  CXTypeKind.CXType_ObjCObjectPointer &&
+            canonical.kind == CXTypeKind.CXType_ObjCObjectPointer &&
             spelling == "Class";
     }
 
@@ -79,9 +103,9 @@ struct Type
         with(CXTypeKind)
             if (isTypedef)
             {
-                auto c = canonicalType;
+                auto c = canonical;
                 return c.kind == CXType_Pointer &&
-                    c.pointeeType.kind == CXType_ObjCSel;
+                    c.pointee.kind == CXType_ObjCSel;
             }
 
             else
@@ -93,31 +117,55 @@ struct Type
         return isObjCIdType || isObjCClassType || isObjCSelType;
     }
 
+    @property bool isPointer ()
+    {
+        return kind == CXTypeKind.CXType_Pointer;
+    }
+
+    @property bool isTypedef ()
+    {
+        return kind == CXTypeKind.CXType_Typedef;
+    }
+
+    @property bool isValid ()
+    {
+        return kind != CXTypeKind.CXType_Invalid;
+    }
+
     @property bool isWideCharType ()
     {
-        with (CXTypeKind)
-            return kind == CXType_WChar;
+        return kind == CXTypeKind.CXType_WChar;
     }
 
-    @property bool isConst ()
+    @property Type canonical()
     {
-        return clang_isConstQualifiedType(cx) == 1;
+        if (isClang)
+            return Type(clang_getCanonicalType(cx));
+        else
+            return Type.init;
     }
 
-    @property bool isExposed ()
+    @property Type pointee()
     {
-        return kind != CXTypeKind.CXType_Unexposed;
-    }
-
-    @property bool isAnonymous ()
-    {
-        return spelling.isEmpty;
+        if (pointee_)
+        {
+            return *pointee_;
+        }
+        else
+        {
+            if (isClang)
+                return Type(clang_getPointeeType(cx));
+            else
+                return Type.init;
+        }
     }
 
     @property Cursor declaration ()
     {
-        auto r = clang_getTypeDeclaration(cx);
-        return Cursor(r);
+        if (isClang)
+            return Cursor(clang_getTypeDeclaration(cx));
+        else
+            return Cursor.empty;
     }
 
     @property FuncType func ()
@@ -130,13 +178,10 @@ struct Type
         return ArrayType(this);
     }
 
-    @property bool isArray ()
+    @property string toString() const
     {
-        return
-            kind == CXTypeKind.CXType_ConstantArray ||
-            kind == CXTypeKind.CXType_IncompleteArray ||
-            kind == CXTypeKind.CXType_VariableArray ||
-            kind == CXTypeKind.CXType_DependentSizedArray;
+        import std.format: format;
+        return format("Type(kind = %s, spelling = %s, isConst = %s)", kind, spelling, isConst);
     }
 }
 
--- a/dstep/translator/Context.d
+++ b/dstep/translator/Context.d
@@ -15,6 +15,7 @@ import clang.TranslationUnit;
 
 import dstep.translator.CommentIndex;
 import dstep.translator.IncludeHandler;
+import dstep.translator.MacroDefinition;
 import dstep.translator.MacroIndex;
 import dstep.translator.Options;
 import dstep.translator.Output;
@@ -33,6 +34,8 @@ class Context
     private TypedefIndex typedefIndex_ = null;
     private Translator translator_ = null;
     private Output globalScope_ = null;
+    public MacroDefinition[string] macroDefinitions;
+    string macroLinkage = "extern (D)";
 
     public this(TranslationUnit translUnit, Options options, Translator translator)
     {
@@ -46,6 +49,11 @@ class Context
         typedefIndex_ = new TypedefIndex(translUnit);
         translator_ = translator;
         globalScope_ = new Output();
+    }
+
+    public string macroLinkagePrefix()
+    {
+        return macroLinkage == "" ? "" : macroLinkage ~ " ";
     }
 
     public string getAnonymousName (Cursor cursor)
--- a/dstep/translator/Context.d
+++ b/dstep/translator/Context.d
@@ -34,10 +34,12 @@ class Context
     private TypedefIndex typedefIndex_ = null;
     private Translator translator_ = null;
     private Output globalScope_ = null;
+    private bool[string] typeNames_;
+
     public MacroDefinition[string] macroDefinitions;
     string macroLinkage = "extern (D)";
 
-    public this(TranslationUnit translUnit, Options options, Translator translator)
+    public this(TranslationUnit translUnit, Options options, Translator translator = null)
     {
         this.translUnit = translUnit;
         macroIndex = new MacroIndex(translUnit);
@@ -47,8 +49,14 @@ class Context
             commentIndex_ = new CommentIndex(translUnit);
 
         typedefIndex_ = new TypedefIndex(translUnit);
-        translator_ = translator;
+
+        if (translator !is null)
+            translator_ = translator;
+        else
+            translator_ = new Translator(translUnit, options);
+
         globalScope_ = new Output();
+        typeNames_ = collectTypeNames(translUnit);
     }
 
     public string macroLinkagePrefix()
@@ -93,6 +101,11 @@ class Context
     public CommentIndex commentIndex()
     {
         return commentIndex_;
+    }
+
+    @property public bool[string] typeNames()
+    {
+        return typeNames_;
     }
 
     public TypedefIndex typedefIndex()
@@ -175,9 +188,9 @@ string cursorScopeString(Context context, Cursor cursor)
 }
 
 /**
-  * Returns true, if there is a variable, of type represented by cursor, among
-  * children of its parent.
-  */
+ * Returns true, if there is a variable, of type represented by cursor, among
+ * children of its parent.
+ */
 bool variablesInParentScope(Cursor cursor)
 {
     import std.algorithm.iteration : filter;
@@ -206,9 +219,72 @@ bool shouldBeAnonymous(Context context, Cursor cursor)
 }
 
 /**
-  * Returns true, if cursor is in the global scope.
-  */
+ * Returns true, if cursor is in the global scope.
+ */
 bool isGlobal(Cursor cursor)
 {
     return cursor.semanticParent.kind == CXCursorKind.CXCursor_TranslationUnit;
 }
+
+/**
+ * The collectTypeNames function scans the whole AST of the translation unit and produces
+ * a set of the type names in global scope.
+ *
+ * The type names are required for the parsing of C code (e.g. macro definition bodies),
+ * as C grammar isn't context free.
+ */
+
+bool[string] collectTypeNames(TranslationUnit translUnit)
+{
+    void collectBasicTypes(ref bool[string] result)
+    {
+        result["void"] = true;
+        result["char"] = true;
+        result["signed char"] = true;
+        result["unsigned char"] = true;
+        result["short"] = true;
+        result["signed short"] = true;
+        result["unsigned short"] = true;
+        result["signed"] = true;
+        result["unsigned"] = true;
+        result["int"] = true;
+        result["signed int"] = true;
+        result["unsigned int"] = true;
+        result["long"] = true;
+        result["signed long"] = true;
+        result["unsigned long"] = true;
+        result["long long"] = true;
+        result["signed long long"] = true;
+        result["unsigned long long"] = true;
+        result["float"] = true;
+        result["double"] = true;
+    }
+
+    void collectTypeNames(ref bool[string] result, Cursor parent)
+    {
+        foreach (cursor, _; parent.all)
+        {
+            switch (cursor.kind)
+            {
+                case CXCursorKind.CXCursor_TypedefDecl:
+                    result[cursor.spelling] = true;
+                    break;
+
+                case CXCursorKind.CXCursor_StructDecl:
+                    result[cursor.spelling] = true;
+                    break;
+
+                default:
+                    break;
+            }
+        }
+    }
+
+    bool[string] result;
+
+    collectBasicTypes(result);
+    collectTypeNames(result, translUnit.cursor);
+
+    return result;
+}
+
--- a/dstep/translator/Context.d
+++ b/dstep/translator/Context.d
@@ -4,7 +4,6 @@
  * Version: Initial created: Mar 21, 2016
  * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
  */
-
 module dstep.translator.Context;
 
 import mambo.core._;
@@ -34,10 +33,9 @@ class Context
     private TypedefIndex typedefIndex_ = null;
     private Translator translator_ = null;
     private Output globalScope_ = null;
-    private bool[string] typeNames_;
+    private Cursor[string] typeNames_;
 
     public MacroDefinition[string] macroDefinitions;
-    string macroLinkage = "extern (D)";
 
     public this(TranslationUnit translUnit, Options options, Translator translator = null)
     {
@@ -56,12 +54,7 @@ class Context
             translator_ = new Translator(translUnit, options);
 
         globalScope_ = new Output();
-        typeNames_ = collectTypeNames(translUnit);
-    }
-
-    public string macroLinkagePrefix()
-    {
-        return macroLinkage == "" ? "" : macroLinkage ~ " ";
+        typeNames_ = collectGlobalTypes(translUnit);
     }
 
     public string getAnonymousName (Cursor cursor)
@@ -103,7 +96,7 @@ class Context
         return commentIndex_;
     }
 
-    @property public bool[string] typeNames()
+    @property public Cursor[string] typeNames()
     {
         return typeNames_;
     }
@@ -227,51 +220,36 @@ bool isGlobal(Cursor cursor)
 }
 
 /**
- * The collectTypeNames function scans the whole AST of the translation unit and produces
+ * The collectGlobalTypes function scans the whole AST of the translation unit and produces
  * a set of the type names in global scope.
  *
  * The type names are required for the parsing of C code (e.g. macro definition bodies),
  * as C grammar isn't context free.
  */
 
-bool[string] collectTypeNames(TranslationUnit translUnit)
+Cursor[string] collectGlobalTypes(TranslationUnit translUnit)
 {
-    void collectBasicTypes(ref bool[string] result)
-    {
-        result["void"] = true;
-        result["char"] = true;
-        result["signed char"] = true;
-        result["unsigned char"] = true;
-        result["short"] = true;
-        result["signed short"] = true;
-        result["unsigned short"] = true;
-        result["signed"] = true;
-        result["unsigned"] = true;
-        result["int"] = true;
-        result["signed int"] = true;
-        result["unsigned int"] = true;
-        result["long"] = true;
-        result["signed long"] = true;
-        result["unsigned long"] = true;
-        result["long long"] = true;
-        result["signed long long"] = true;
-        result["unsigned long long"] = true;
-        result["float"] = true;
-        result["double"] = true;
-    }
-
-    void collectTypeNames(ref bool[string] result, Cursor parent)
+    void collectGlobalTypes(ref Cursor[string] result, Cursor parent)
     {
         foreach (cursor, _; parent.all)
         {
             switch (cursor.kind)
             {
                 case CXCursorKind.CXCursor_TypedefDecl:
-                    result[cursor.spelling] = true;
+                    result[cursor.spelling] = cursor;
                     break;
 
                 case CXCursorKind.CXCursor_StructDecl:
-                    result[cursor.spelling] = true;
+                    result["struct " ~ cursor.spelling] = cursor;
+                    break;
+
+                case CXCursorKind.CXCursor_UnionDecl:
+                    result["union " ~ cursor.spelling] = cursor;
+                    break;
+
+
+                case CXCursorKind.CXCursor_EnumDecl:
+                    result["enum " ~ cursor.spelling] = cursor;
                     break;
 
                 default:
@@ -280,10 +258,9 @@ bool[string] collectTypeNames(TranslationUnit translUnit)
         }
     }
 
-    bool[string] result;
+    Cursor[string] result;
 
-    collectBasicTypes(result);
-    collectTypeNames(result, translUnit.cursor);
+    collectGlobalTypes(result, translUnit.cursor);
 
     return result;
 }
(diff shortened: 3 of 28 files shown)


clang/Token.d
```diff
@@ -18,6 +18,20 @@ import clang.Util;
...
+    punctuation = CXTokenKind.CXToken_Punctuation,
+    keyword = CXTokenKind.CXToken_Keyword,
+    identifier = CXTokenKind.CXToken_Identifier,
+    literal = CXTokenKind.CXToken_Literal,
+    comment = CXTokenKind.CXToken_Comment,
+}
+
+TokenKind toD(CXTokenKind kind)
+{
+    return cast (TokenKind) kind;
```
jacob-carlborg: Please don't put a space after `cast`.

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1404 @@
...
+                tokens = tokens[1..$];
+                return true;
+            }
+        }
+    }
+
+    return false;
+}
+
+bool accept(ref TokenRange tokens, ref string spelling, TokenKind kind)
```
jacob-carlborg: Doesn't this cause a conflict with the first overload of `accept`?

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1404 @@
...
+class Identifier : Expression
+{
+    this (string spelling)
+    {
+        this.spelling = spelling;
+    }
+
+    string spelling;
+
+    override string transl(ref bool[string] imports)
```
jacob-carlborg: Please avoid abbreviations.

ciechowoj: Is `Expr` suffix on some types ok?

dstep/translator/MacroDefinition.d
ciechowoj: Doesn't seem so, however I hope it isn't due to some bug in dmd. The first overload requires explicit template parameters. I can change it, if you want.

clang/Token.d
ciechowoj: Fixed it in separate commit. I'll rebase it after the review is complete.

dstep/translator/MacroDefinition.d
jacob-carlborg: Yeah, I guess that's ok.

dstep/translator/MacroDefinition.d
jacob-carlborg: If you don't specify any template parameters (since it's a variadic template parameter) it should cause a conflict. If the intention is that at least one template parameter should always be passed, then add a template constraint to enforce that.

ciechowoj: I'm not sure if [this](https://dlang.org/spec/template.html#variadic_template_specialization) does'n cover the case, if we consider a non-template version as a template without any parameters.

ciechowoj: Anyway, I fixed it.

unit_tests/MacroTranslTests.d
```diff
@@ -0,0 +1,475 @@
...
+C");
+
+    assertTMD(q"D
+int FOO()
+{
+    return 0;
+}
+D", q"C
+#define FOO() 0
+C");
```
jacob-carlborg: I don't see a difference compared to the previous assertion.

unit_tests/MacroTranslTests.d
```diff
@@ -0,0 +1,475 @@
...
+int FOO()
+{
+    return 0 + 1;
+}
+D", q"C
+#define FOO() 0 + 1
+C");
+
+    assertTMD(q"D
+int FOO(int a)
```
jacob-carlborg: How do you determine that `a` should be `int`. Based on `1` in the body? What if a long is passed when calling the macro, that will not compile in D. Isn't using a template closer to how the macro works?

unit_tests/MacroTranslTests.d
```diff
@@ -0,0 +1,475 @@
...
+#define FOO(a, b, c) (\
+    ((a) * 100) \
+  + ((b) * 10) \
+  + c)
+C");
+
+    assertTMD(q"D
+string STRINGIZE(T)(in T major, in T minor)
+{
+    import std.conv;
```
jacob-carlborg: When doing a local import, it's recommended to do a selective import as well.

unit_tests/MacroTranslTests.d
ciechowoj: Fixed.

unit_tests/MacroTranslTests.d
```diff
@@ -0,0 +1,475 @@
...
+#define FOO_MAX ((foo_t)(-1))
+#define MAX_UINT UINT_31_MAX
+C", q"D
+extern (C):
+
+alias uint uint_32;
+alias ulong foo_t;
+
+enum UINT_31_MAX = cast(uint_32) 0x7fffffffL;
+enum UINT_32_MAX = cast(uint_32) (-1);
```
jacob-carlborg: Is the parentheses around `-1` necessary here? You seem to have been able remove redundant parentheses in other cases.

unit_tests/MacroTranslTests.d
ciechowoj: You are right, I probably wanted be too smart.

jacob-carlborg:
Shouldn't all macros take their parameters by reference?

ciechowoj:
> Shouldn't all macros take their parameters by reference?

Macros are mostly pure functions (at least the currently handled ones), I don't see any advantage (maybe speed) and you cannot pass rvalue to reference parameter.

unit_tests/MacroTranslTests.d
ciechowoj: Fixed.

jacob-carlborg:
> Macros are mostly pure functions (at least the currently handled ones), I don't see any advantage (maybe speed) and you cannot pass rvalue to reference parameter.

You have a test for a macro `#define foo(x) ++x` which is translated to `auto foo(T)(in T x){ return ++x; }`. That has not the same behaviour as the C macro. First problem, the parameter is declared as `in` which means `scope const`. The body won't compile since you cannot modify a const parameter. Second problem is when the macro/function is used:

```d
int a = 0;
foo(a);
assert(a == 1);
```

The above assertion will pass in C since the macro is expanded inline. But it will fail in D since the integer is copied to the function and the caller won't see the modification.

ciechowoj:
Good point. I'll see what can be done about it, but it do not seem to be trivial to fix : /.

unit_tests/MacroTranslTests.d
ciechowoj: Fixed.

jacob-carlborg:
I think the correct translation is to take all arguments by reference, unless you can figure out that the body doesn't modify the arguments. Using `ref`instead of `in` should be a trivial fix. Figuring out exactly when `ref` is needed, not so much 😃. For me it's perfectly acceptable to use `ref` for all parameters. If you want to and have the time, you can come back later and try to only insert `ref` when is actually necessary.

ciechowoj:
Ok, I may change `in` to `ref` and fix it later in another PR. What do you think?

unit_tests/MacroTranslTests.d
ciechowoj: Fixed.

ciechowoj:
Ok, I fixed if for `++` and `--`, currently other assigning operators aren't supported, so the risk that reference is necessary and it isn't translated that way is small. I corrected the naming of template parameters too (they have different names now). 

To illustrate problem with references consider following code:

```
#define foo(x) ((x) + 1)
#define bar(x) ((x) * 2)
#define baz(x) (foo(bar(x)) % 3)
```

If all parameters are by reference straightforward translation of `baz` will not work.

```
auto baz(T)(ref T x)
{
    return foo(bar(x)) % 3;
}
```

RValue returned by `bar` cannot be passed as reference parameter of `foo`. D doesn't allow temporaries to be passed to reference parameters.

jacob-carlborg:
> RValue returned by bar cannot be passed as reference parameter for foo. D doesn't allow temporaries to be passed to reference parameters.

What about [auto ref](https://dlang.org/spec/template.html#auto-ref-parameters)?

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1403 @@
...
+    {
+        import std.format : format;
+
+        if (surplus)
+            return format("%s", subexpr.translate(imports));
+        else
+            return format("(%s)", subexpr.translate(imports));
+    }
+
+    bool surplus() const
```
jacob-carlborg: What does `surplus` mean, is it an abbreviation of something?

ciechowoj:
Looks promising, I'll read more about it and  change the code tomorrow.

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1403 @@
...
+
+    ExprType guessExprType()
+    {
+        return ExprType(ExprType.kind.generic);
+    }
+
+    void guessParamTypes(ref ExprType[string] params, ExprType type)
+    { }
+
+    override string toString() { return ""; }
```
jacob-carlborg: Please put the body on its own line, fully expanded as the rest of the methods.

dstep/translator/MacroDefinition.d
ciechowoj: Synonym to `unused`/`excessive`.

[1](http://www.merriam-webster.com/dictionary/surplus)
[2](http://www.thesaurus.com/browse/surplus)

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1403 @@
...
+        directive.expr.debraced.translate());
+}
+
+string translateDirectiveParamList(string[] params, ExprType[string] types)
+{
+    import std.algorithm.iteration : map;
+    import std.format : format;
+    import std.array : join;
+
+    return map!(a => format("%s %s", types[a].asParamType(), a))(params).join(", ");
```
jacob-carlborg: I prefer to use UFCS for algorithm functions.

dstep/translator/MacroDefinition.d
jacob-carlborg: Hmm, never heard of it before.

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1403 @@
...
+}
+
+string translateDirectiveTypeList(string[] params, ref ExprType[string] types)
+{
+    import std.algorithm.iteration : filter, map;
+    import std.algorithm.searching : count;
+    import std.format : format;
+    import std.array : join;
+    import std.array : appender;
```
jacob-carlborg: You can put `appender` on the previous line.

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1403 @@
...
+            else
+                translateFunctDirective(output, context, definition);
+        }
+
+        context.macroDefinitions[definition.spelling] = definition;
+    }
+}
+
+
+
```
jacob-carlborg: Please remove excessive newlines.

unit_tests/MacroTranslTests.d
```diff
@@ -0,0 +1,457 @@
...
+
+private alias assertTMD = assertTranslatesMacroDefinition;
+
+// Translate basic macro definitions.
+unittest
+{
+    assertTMD(q"D
+D", q"C
+#define FOO
+C");
```
jacob-carlborg: Isn't it more natural to have the C code as the first parameter? Like here https://github.com/jacob-carlborg/dstep/blob/master/unit_tests/UnitTests.d#L14-L16.

unit_tests/MacroTranslTests.d
```diff
@@ -0,0 +1,457 @@
...
+    return ++a + --b + &c + *d + +e - (-f);
+}
+D", q"C
+#define FOO(a, b, c, d, e, f) ++a + --b + &c + *d + +e - (-f)
+C");
+
+    assertTMD(q"D
+size_t FOO(T)(in T a)
+{
+    return sizeof(a);
```
jacob-carlborg: `sizeof` in D is a property, i.e. `a.sizeof`.

dstep/translator/Context.d
```diff
@@ -204,3 +225,58 @@ bool isGlobal(Cursor cursor)
...
+                result[cursor.spelling] = true;
+                break;
+
+            default:
+                break;
+        }
+    }
+}
+
+bool[string] collectTypeNames(TranslationUnit translUnit)
```
jacob-carlborg: Could you please document/describe how these "collect" functions work.

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1403 @@
...
+import std.array : Appender;
+
+import clang.c.Index;
+import clang.Cursor;
+import clang.Token;
+
+import dstep.translator.Context;
+import dstep.translator.Output;
+
+bool accept(Args...)(ref TokenRange tokens, ref string spelling, TokenKind kind) if (Args.length > 0)
```
jacob-carlborg: Could you please document/describe how these "accept" functions work.

unit_tests/MacroTranslTests.d
ciechowoj: I would like to change all the assertion to look as here. I believe the order in `assertTranslates` was unlucky. Most unittesting frameworks takes *expected* value as first parameter.

dstep/translator/MacroDefinition.d
ciechowoj: Ok, I'll provide documentation tomorrow.

unit_tests/MacroTranslTests.d
jacob-carlborg: That's true. But I think in this case, when there are two string literals, the input and the expected, it feels more natural with the input value (the C code) first. The actual result is not in the assertion, it's the input value and the expected result. So it's still not like other frameworks.

dstep/translator/MacroDefinition.d
ciechowoj: Fixed.

dstep/translator/MacroDefinition.d
ciechowoj: Fixed.

dstep/translator/MacroDefinition.d
ciechowoj: Fixed.

dstep/translator/MacroDefinition.d
ciechowoj: Fixed.

unit_tests/MacroTranslTests.d
ciechowoj: You have a point. Fixed.

unit_tests/MacroTranslTests.d
ciechowoj: Fixed.

ciechowoj:
> What about auto ref?

Changed all arguments to `auto ref`.

dstep/translator/Context.d
ciechowoj: Done.

dstep/translator/MacroDefinition.d
ciechowoj: Done.

ciechowoj:
Unrelated stuff:

I tried to generate documentation with `dub build --build=docs` and it doesn't work (dub says that dmd returns 1 in some point). Have you ever tried to generate docs for dstep? I suppose it would be nice if it works.

jacob-carlborg:
No, I have not tried generating the documentation. You can try with the verbose flag to get the actual commands Dub is running. Then try manually running the individual commands to see if you get some more information.

ciechowoj:
I've just tried to do like so, but it seems like problem with dmd and beyond printing some warnings from dependencies it doesn't say nothing useful : P.

jacob-carlborg:
I get some errors if I remove the warnings flag (-w), mostly from Tango 😃.

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1429 @@
...
+import clang.Token;
+
+import dstep.translator.Context;
+import dstep.translator.Output;
+
+/**
+  * The accept family of functions parse tokens. In the case of successful parsing
+  * advances beginning of tokens to the next token and returns true.
+  *
+  * The parsing is successful, if the first token in tokens has a kind kind and
```
jacob-carlborg: * "kind" is repeated twice
* "has a kind" is that any kind of the given kind

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1429 @@
...
+  * its spelling matches on of the strings passed as Args.
```
jacob-carlborg: on -> one ?

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1429 @@
...
+  * It assigns the spelling of the token to the spelling parameter.
+  */
+
+bool accept(Args...)(ref TokenRange tokens, ref string spelling, TokenKind kind) if (Args.length > 0)
```
jacob-carlborg: Is there a reason when `Args` is a template parameter instead of `string[]...`?

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1429 @@
...
+
+import clang.c.Index;
+import clang.Cursor;
+import clang.Token;
+
+import dstep.translator.Context;
+import dstep.translator.Output;
+
+/**
+  * The accept family of functions parse tokens. In the case of successful parsing
```
jacob-carlborg: Please align the `*` to the first `*` in `/**`. This will match the module header/documentation.

dstep/translator/Context.d
```diff
@@ -204,3 +225,66 @@ bool isGlobal(Cursor cursor)
...
     return cursor.semanticParent.kind == CXCursorKind.CXCursor_TranslationUnit;
 }
+
+/**
+  * The collectTypeNames function scans the whole AST of the translation unit and produces
+  * a set of the type names in global scope.
+  *
+  * The type names are required for parsing C code (e.g. macro definition bodies),
+  * as C grammar isn't context free.
+  */
```
jacob-carlborg: * Same here with the alignment of `*`
* I still don't understand how this function is used

dstep/translator/MacroDefinition.d
ciechowoj: I wanted to say a kind `kind`. Fixed.

dstep/translator/MacroDefinition.d
ciechowoj: Fixed.

dstep/translator/MacroDefinition.d
ciechowoj: The only reason is performance (some branching can be avoided), if you consider it premature optimization, I can change it.

dstep/translator/MacroDefinition.d
ciechowoj: Fixed.

dstep/translator/Context.d
ciechowoj: To parse C-style casts (and `sizeof`s) you need to have information about type names. Without it you do not know if `(int)(4)` is a cast or a function call. This function collects that info.

dstep/translator/MacroDefinition.d
jacob-carlborg: Most of my concern is that the API of the function is not very clear. There's no indication in the documentation or in a template constraint that `Args` should only be strings. The name `Args` is also very generic, i.e. that you could pass in any type of arguments.

dstep/translator/Context.d
jacob-carlborg: So it's basically a list of all the types in the translation unit? And it's an associative array to emulate a set?

ciechowoj: Exactly.

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1429 @@
...
+import clang.Token;
+
+import dstep.translator.Context;
+import dstep.translator.Output;
+
+/**
+ * The accept family of functions parse tokens. In the case of successful parsing
+ * advances beginning of tokens to the next token and returns true.
+ *
+ * The parsing is successful, if the first token in tokens has is of specified kind
```
jacob-carlborg: * has is -> is?
* specified -> the specified

dstep/translator/Context.d
jacob-carlborg: Ok, now I get it 😃.

dstep/translator/Context.d
```diff
@@ -198,9 +219,72 @@ bool shouldBeAnonymous(Context context, Cursor cursor)
...
+        result["unsigned int"] = true;
+        result["long"] = true;
+        result["signed long"] = true;
+        result["unsigned long"] = true;
+        result["long long"] = true;
+        result["signed long long"] = true;
+        result["unsigned long long"] = true;
+        result["float"] = true;
+        result["double"] = true;
+    }
```
jacob-carlborg: * It seems that (un)signed can be specified in any position, before, after and in between (in the case of long long)
* `short int`, `long int`, `long double` are also valid types
* `const` can be specified for any type
* `__complex__` and `_Complex` in any combination of the floating point types are valid as well. The floating point type does not need to be specified
* `_Bool` and `bool` (if `<stdbool.h>` is included) are valid types in C99. I tried compiling code containing these types with `-std=c89 -Wall` and got no errors or warnings

dstep/translator/Context.d
```diff
@@ -198,9 +219,72 @@ bool shouldBeAnonymous(Context context, Cursor cursor)
...
+                case CXCursorKind.CXCursor_TypedefDecl:
+                    result[cursor.spelling] = true;
+                    break;
+
+                case CXCursorKind.CXCursor_StructDecl:
+                    result[cursor.spelling] = true;
+                    break;
+
+                default:
+                    break;
```
jacob-carlborg: What about unions and enums?

ciechowoj: Right, it is a bit leaky at the moment, there is probably a problem with tag namespace of strucutres (e.g. `struct Foo x;`). I'll fix it when I cope with constraining that `Args...` thing to be a string...

dstep/translator/MacroDefinition.d
ciechowoj: Fixed.

dstep/translator/MacroDefinition.d
ciechowoj: Fixed.

dstep/translator/Context.d
```diff
@@ -198,9 +219,72 @@ bool shouldBeAnonymous(Context context, Cursor cursor)
...
+
+/**
+ * The collectTypeNames function scans the whole AST of the translation unit and produces
+ * a set of the type names in global scope.
+ *
+ * The type names are required for the parsing of C code (e.g. macro definition bodies),
+ * as C grammar isn't context free.
+ */
+
+bool[string] collectTypeNames(TranslationUnit translUnit)
```
jacob-carlborg: I found this snippet in the newsgroup as way to emulate sets. It clearly shows indentation that it's a set and not an associative array.

```d
alias Set(T) = void[0][T];

void add(T)(ref void[0][T] set, T value) {
    set[value] = (void[0]).init;
}

bool contains(T)(inout(void[0][T]) set, T value) {
    return (value in set) !is null;
}

Set!(string) a;
a.add("asd");
assert(a.contains("asd"));
a.remove("asd");
```

ciechowoj: > It seems that (un)signed can be specified in any position, before, after and in between (in the case of long long)
> short int, long int, long double are also valid types
> const, volatile and _Atomic (C11) can be specified for types as well
> __complex__ and _Complex in any combination of the floating point types are valid as well. The > floating point type does not need to be specified
> _Bool and bool (if <stdbool.h> is included) are valid types in C99. I tried compiling code containing these types with -std=c89 -Wall and got no errors or warnings
> When a type consist of multiple words, i.e. long int, any order of the words are valid
> What about the spaces. If the source code specifies "long long" (seems like multiple spaces are removed by GitHub), will the spelling contain that as well or does Clang remove the extra spaces?

> What about unions and enums?

Well, I need to write parsing of the types. It will not be hard, but there is another problem. Even if I parse the the types, then I need to translate them. But, there are routines for translating types in dstep already, most in `dstep/translator/Type.d`. But they only accept information in the form of `Type` and `Cursor` objects. It would be stupid to have separate code path to translate types in defines.

This is related to what I said in #65 (desing problem). One way would be to make all the functions templates, but it doesn't seem doable in practice.

I propose to change a `Type` type from struct to class and then create two new classes that inherits from `Type`: `CLangType` and let say `UserType`. `CLangType` would talk to cx and access clang interface directly, `UserType` would have a constructor allowing an user to create it and fill with data directly.

Another way would be to create a `TypeImpl` struct and store both pointer to `TypeImpl` and `cx` inside `Type` and in every method check which one to use with if.

```
struct TypeImpl
{
    string spelling;
    bool isConst;
    etc...
};

struct Type
{
    mixin CX;
    TypeImpl* impl;

    string spelling() 
    {
        return impl ? impl.spelling : /* get spelling from cx */;
    }
}

```

Another way is to fill `Type` with all required fields and copy them on creation.

```
struct Type
{
    string spelling;
    bool isConst;
    bool isVolatile;
    Type* canonicalType;
    etc.

    Type(void* returned_from_clang)
    {
        // get data from libclang and fill fields
    }
}
```

What do you think about it?

jacob-carlborg: I don't like the second suggestion with `TypeImpl`. I'm leaning towards the first suggestion with classes. The last suggestion isn't so bad either. The last one has the advantage that we don't need to come up with names for two new classes (I'm not a big fan of `UserType`).

ciechowoj:
I would prefer the first one to, its easiest to implement. The name `UserType` can be changed if you have better idea, I have mixed feeling about it too.

ciechowoj:
Maybe `CustomType` or `ForeignType` (as type outside clang).

jacob-carlborg:
I think `CustomType`, and the casing for the other one should be `ClangType`.

ciechowoj:
Ok.

ciechowoj:
Eventually I used third suggestion..., I started with changing the struct to the class but it doesn't wont to work so well, there is too much difference between reference and value semantics.

cast operator (in macro-definitions) works only for basic types for now. Structs, unions enums and pointers are on their way.

ciechowoj:
Ok, I added support for struct, unions, enums and typedefs.

ciechowoj:
> const, volatile and _Atomic (C11) can be specified for types as well
volatile isn't supported by D, there is no API for _Atomic in libclang so I'm giving up on them.

jacob-carlborg:
> volatile isn't supported by D, there is no API for _Atomic in libclang so I'm giving up on them.

The question is if it needs to be recognized as part of the type anyway. I'm not sure if `const` can be kept in the translation, because const is transitive in D, while not in C. In the same sense it might be necessary to drop `volatile` and `_Atomic` as well i the translation.

ciechowoj:
Yes, situation with volatile and _Atomic is pretty clear, but I have no idea what to do about `const`...

ciechowoj:
I would prefer not to change anything about translation of `const`, as this PR is about parsing macro definitions and translation code is as it was before.

jacob-carlborg:
> I would prefer not to change anything about translation of const, as this PR is about parsing macro definitions and translation code is as it was before.

Absolutely.

clang/Type.d
```diff
@@ -93,31 +110,53 @@ struct Type
...
 
-    @property bool isConst ()
+    @property Type canonical()
     {
-        return clang_isConstQualifiedType(cx) == 1;
+        return canonical_ ? *canonical_ : Type(clang_getCanonicalType(cx));
     }
 
-    @property bool isExposed ()
+    @property void canonical(Type type)
```
jacob-carlborg: Is it necessary to have setters, can everything be handled in the constructor?

clang/Type.d
```diff
@@ -130,13 +169,10 @@ struct Type
...
     {
-        return
-            kind == CXTypeKind.CXType_ConstantArray ||
-            kind == CXTypeKind.CXType_IncompleteArray ||
-            kind == CXTypeKind.CXType_VariableArray ||
-            kind == CXTypeKind.CXType_DependentSizedArray;
+        import std.format: format;
+        return format("Type(kind = %s, spelling = %s, isConst = %s)", kind, spelling, isConst);
     }
 }
```
jacob-carlborg: For the future: it looks like you have changed order of the methods, which causes unnecessary noise in the diff, making it harder to review.

clang/Type.d
ciechowoj: Everything can be handled in constructor, however I thought it would be better to have a setters because some fields do not require initialization every time, so there would be many (a few) unused parameters in constructors of a few different versions of constructors. But I can change it.

clang/Type.d
```diff
@@ -4,73 +4,90 @@
  * Version: Initial created: Jan 29, 2012
  * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
  */
+
```
jacob-carlborg: I prefer that there's no newline between the documentation and the symbol it's attached to.

clang/Type.d
ciechowoj: Ok, I'll try to remember about that. I've just grouped some kinds of methods together.

clang/Type.d
jacob-carlborg: So far a type has been immutable, in the sense that it did not have any setters.

clang/Type.d
jacob-carlborg: That's understandable, but with the amount of code in your pull requests, every diff counts. It's better to make as few changes as possible, especially when it comes to style related changes. I know that can be hard but it's better to leave that for a completely separate pull request. It's also in your best interest because it will make my life easier and speed up the review process 😃.

dstep/translator/Context.d
```diff
@@ -44,8 +49,19 @@ class Context
...
+        else
+            translator_ = new Translator(translUnit, options);
+
         globalScope_ = new Output();
+        typeNames_ = collectGlobalTypes(translUnit);
+    }
+
+    public string macroLinkagePrefix()
+    {
+        return macroLinkage == "" ? "" : macroLinkage ~ " ";
```
jacob-carlborg: Why isn't the space included from the beginning?

clang/Type.d
ciechowoj: Fixed.

dstep/translator/Context.d
```diff
@@ -33,8 +34,12 @@ class Context
     private TypedefIndex typedefIndex_ = null;
     private Translator translator_ = null;
     private Output globalScope_ = null;
+    private Cursor[string] typeNames_;
 
-    public this(TranslationUnit translUnit, Options options, Translator translator)
+    public MacroDefinition[string] macroDefinitions;
+    string macroLinkage = "extern (D)";
```
jacob-carlborg: Is it necessary to set the linkage?

ciechowoj: The whole file is wrapped to `extern (C):`, so I thought it would be wise to override it for the translations of macros to not cause a clash with other symbols (e.g. from static library, mangled in C-style).

> Why isn't the space included from the beginning?

I thought about a CLI switch to control that behavior and I do not want the user to be obliged to remember about formatting issues.

dstep/translator/MacroDefinition.d
```diff
@@ -1188,7 +1221,417 @@ Expression parseCondExpr(ref TokenRange tokens, bool[string] table)
...
+        acceptIdentifier(local, spelling))
+    {
+        if (auto ptr = ("union " ~ spelling in table))
+        {
+            type.kind = CXTypeKind.CXType_Record;
+            type.spelling = spelling;
+            tokens = local;
+            return true;
+        }
+    }
```
jacob-carlborg: These two if statements are exactly the same, except for the keyword `struct/union`. If you add an additional variable, `keywordType` or similar and pass that in to the `accept` function and two template arguments `struct` and `union` then `keywordType` will contain which keyword that actually match on return?

dstep/translator/MacroDefinition.d
```diff
@@ -1188,7 +1221,417 @@ Expression parseCondExpr(ref TokenRange tokens, bool[string] table)
...
+
+    alias qualifiers = AliasSeq!(
+        "const",
+        "volatile",
+        "_Atomic");
+
+    if (accept!(qualifiers)(tokens, spelling))
+        return true;
+    else
+        return false;
```
jacob-carlborg: Doesn't `accept` already return a bool?

dstep/translator/MacroDefinition.d
```diff
@@ -1188,7 +1221,417 @@ Expression parseCondExpr(ref TokenRange tokens, bool[string] table)
...
+    }
+
+    if (specifiers.contains("signed"))
+    {
+        type = Type(CXTypeKind.CXType_Int, "int");
+        return true;
+    }
+
+    return false;
+}
```
jacob-carlborg: Does this handle `long int`?

dstep/translator/Context.d
jacob-carlborg: Do you mean that the user could set the linkage for a translated macro?

dstep/translator/MacroDefinition.d
ciechowoj: Fixed.

dstep/translator/MacroDefinition.d
ciechowoj: You are right, fixed.

dstep/translator/Context.d
ciechowoj: Yes that was the aim, I kinda forgot to add a flag to the driver code. What do you think about that idea?

jacob-carlborg: Currently I cannot come up with a use case for that. Perhaps keep it simple and not make that possible.

dstep/translator/MacroDefinition.d
ciechowoj: Yes here are tests:
[link](https://github.com/jacob-carlborg/dstep/pull/67/files#diff-21c4f0499bb291726704dbb88c44efaeR97)

jacob-carlborg: Ah, I see now, thanks.

ciechowoj: Eh, that link isn't very good, the tests are in unit_tests/MacroParsingTests.d:128 .

jacob-carlborg: I found the test anyway 😃.

dstep/translator/Context.d
ciechowoj: Done.

dstep/translator/Context.d
ciechowoj: Fixed.

clang/Type.d
ciechowoj: Fixed.

unit_tests/Assert.d
```diff
@@ -0,0 +1,143 @@
...
+    if (children.length != 1)
+        throw new AssertError("Assertion failure", file, line);
+
+    Cursor cursor = children[0];
+
+    assert(cursor.kind == CXCursorKind.CXCursor_MacroDefinition);
+
+    auto definition = parsePartialMacroDefinition(cursor.tokens, context.typeNames);
+
+    bool[string] imports;
```
jacob-carlborg: I guess this could use the new set "type".

clang/Type.d
```diff
@@ -93,31 +115,39 @@ struct Type
...
-    @property bool isAnonymous ()
+    @property bool isWideCharType ()
     {
-        return spelling.isEmpty;
+        return kind == CXTypeKind.CXType_WChar;
+    }
+
+    @property Type canonical()
+    {
+        return Type(clang_getCanonicalType(cx));
```
jacob-carlborg: This will only work if the type is create from a Clang type. Perhaps adding an assertion?

clang/Type.d
```diff
@@ -93,31 +115,39 @@ struct Type
...
+    @property Type pointee()
+    {
+        return pointee_ ? *pointee_ : Type(clang_getPointeeType(cx));
     }
 
     @property Cursor declaration ()
     {
-        auto r = clang_getTypeDeclaration(cx);
-        return Cursor(r);
+        return Cursor(clang_getTypeDeclaration(cx));
```
jacob-carlborg: Same here, will only work for Clang types.

ciechowoj: I assumed that libclang detect that the type is invalid and return invalid/empty cursor/type, but now its explicit.

unit_tests/Assert.d
ciechowoj: I guess you are right, fixed.

clang/Type.d
```diff
@@ -93,31 +117,55 @@ struct Type
...
+            else
+                return Type.init;
+        }
     }
 
     @property Cursor declaration ()
     {
-        auto r = clang_getTypeDeclaration(cx);
-        return Cursor(r);
+        if (isClang)
```
jacob-carlborg: Isn't it better to us an `assert`, because you're not really supposed to call this method unless the type is coming from Clang?

ciechowoj: Some code other depends on this. E.g. to check if the type has associated declaration.

jacob-carlborg: Ok, I see.

unit_tests/MacroTranslTests.d
```diff
@@ -0,0 +1,566 @@
...
+C", q"D
+extern (D) int FOO(T)(auto ref T a)
+{
+    return a == 1 ? 0 : 1 ? 2 : 3;
+}
+D");
+
+}
+
+// Translate cast operator (FIXME: type inference)
```
jacob-carlborg: What's up with this `FIXME`?

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1859 @@
...
+        if (subexpr !is null)
+        {
+            UnaryExpr expr = new UnaryExpr;
+            expr.subexpr = subexpr;
+            expr.operator = spelling;
+            tokens = local;
+            return expr;
+        }
+
+        // FIXME: unary-expression ::= sizeof ( type-name )
```
jacob-carlborg: What's up with this `FIXME`?

dstep/translator/MacroDefinition.d
```diff
@@ -0,0 +1,1859 @@
...
+        "char",
+        "short",
+        "int",
+        "long",
+        "float",
+        "double",
+        "signed",
+        "unsigned",
+        // "__complex__", TBD
+        // "_Complex", TBD
```
jacob-carlborg: What's the status of these two "complex", do you plan to implement them in this PR?

unit_tests/MacroTranslTests.d
ciechowoj: It is about trying to guess the return type of macro. But I suppose it is unimportant now, as D's `auto` is good enough. I removed the fixme.

dstep/translator/MacroDefinition.d
ciechowoj: Parsing of C-style sizeof is type dependent too (you need to know the names of types). I added that FIXME when translation of types was not fully implemented. But as it is handled quite good now, I've just added a commit that implements parsing and translation of type-dependent `sizeof` operators.

dstep/translator/MacroDefinition.d
jacob-carlborg: Fair enough.

unit_tests/MacroTranslTests.d
jacob-carlborg: 👍

dstep/translator/MacroDefinition.d
jacob-carlborg: Is the FIXME still needed?

ciechowoj: Of course not. Forgot to push.

jacob-carlborg:
👌

jacob-carlborg:
Could you please clean up the commits a bit and then I can merge this.

ciechowoj:
done

jacob-carlborg:
Please cleanup the commit message for the second commit.

ciechowoj:
Actually I intentionally left it that way, but ok : )

jacob-carlborg:
Ideally, the fixes you've done according to my comments should not be visible in the history.

ciechowoj:
Done, It has some sense.

ciechowoj:
Thanks.

jacob-carlborg:
Thanks for the great work 👍.

jacob-carlborg:
Well, I haven't really planned anything 😃. I could at least create a tag but building binaries is kind of a pain in the ass.

jacob-carlborg:
@ciechowoj could you please update the changelog as well, I always forget that. I suspecting it's not completely up to date, could you verify that?

ciechowoj:
I've just added a PR.

PR #68, opened 2016-06-08
Packages

ciechowoj:
Adds CLI switch `--package <package_prefix>` that allows to specify a package. If the switch is present following decl will be added at the top:

`module  <package_prefix>.<base_name_without_extension>`;

Improves support for includes - if multiple files are specified as dstep input and `--package` switch is present, `dstep` will threat the other files as part of the package and will add appropriate import directives.


clang/Util.d
```diff
@@ -165,3 +165,11 @@ string namedTempDir(string prefix)
 
     return path.idup;
 }
+
+string asAbsNormPath(string path)
+{
+    import std.array;
+    import std.path;
+
+    return path.asAbsolutePath.asNormalizedPath.array.idup;
```
jacob-carlborg: I think you're suppose to use `to!(string)` when converting a range to a string.

dstep/driver/Application.d
```diff
@@ -37,20 +38,24 @@ class Application : DStack.Application
 
     private
     {
-        Language language;
+        Options blueprint;
```
jacob-carlborg: Why `blueprint`

ciechowoj: `blueprint` holds default values, then I need to change some values before passing it to subsequent processing. I thought blueprint in the meaning of draft is a good name.

clang/Util.d

dstep/driver/Application.d
jacob-carlborg: Isn't `defaultOptions` a more clear name?

dstep/translator/IncludeHandler.d
```diff
@@ -84,58 +92,119 @@ class IncludeHandler
...
+    this (Options options)
+    {
+        import std.format;
+
+        this.options = options;
+
+        if (options.packageName != "")
+        {
+            auto inputFiles = options.inputFiles.filter!(
+                x => x != options.inputFile)();
```
jacob-carlborg: I prefer to not include the last set of parentheses in cases where there are only template parameters.

jacob-carlborg:
> Improves support for includes - if multiple files are specified as dstep input and --package switch is present, dstep will threat the other files as part of the package and will add appropriate import directives.

Will this translate the actual includes?

dstep/driver/Application.d
ciechowoj: Fixed.

dstep/driver/Application.d
```diff
@@ -37,20 +38,24 @@ class Application : DStack.Application
 
     private
     {
-        Language language;
+        Options defaultOptions;
```
jacob-carlborg: I assume this is does not change, so it could be declared with `enum`.

dstep/translator/IncludeHandler.d

dstep/driver/Application.d
ciechowoj: Could you explain it more, I don't understand.

jacob-carlborg: `enum defaultOptions = Options()`

ciechowoj:
> Will this translate the actual includes?

Yes, if the 'include' includes some other file specified for translation (it uses the dstap ability to translate multiple file at once).

dstep/driver/Application.d
ciechowoj: Well, it is initialized in methods: `handleArguments` and `handleObjectiveC`, so it cannot be enum.

jacob-carlborg:
> Yes, if the 'include' includes some other file specified for translation (it uses the dstap ability to translate multiple file at once).

The thing is that C includes don't translate very well to D imports. There can be files that uses symbols declared in other files without including no files at all. There can be includes that are not necessary.

The current approach, in the few cases when imports are added, works by completely ignoring the includes and instead looking up where a symbol is declared.

dstep/driver/Application.d
jacob-carlborg: Aha, I missed that. Then `defaultOptions` might not be a good name after all.

ciechowoj:
Well, but the old behavior is still maintained if someone doesn't use `--package` switch. The intention of the feature implemented in this PR is to enable automatic translation of the set some well-written header files.

The includes that will be translated needs to be explicitly specified as an input, the rest of include directives is ignored as up to now.

The use case scenario is that there is some library consisting of a few header files (e.g. libclang). The user then can specify all libclang headers as an input, specify the package name and dstep will translate the `#include`s of know files automatically.

jacob-carlborg:
But can't the old behavior be extended instead? It seems like a more robust solution.

dstep/driver/Application.d
ciechowoj: Yes, that was the reason I used `blueprint`, I can change it back or to something else. `defaultOptions` isn't that bad too.

ciechowoj:
So far I do not have a better idea.

ciechowoj:
Beside this I do not see a robust way to infer a package name.

dstep/driver/Application.d
jacob-carlborg: Yeah, I don't know which one is best. `defaultOptions` made more sense when I though it was constant.

ciechowoj:
C libraries are written in so unstandardized way : (

ciechowoj:
I'm not considering this as a solution for handling include files, rather a solution to handle packages.

jacob-carlborg:
The package name is a separate thing. Instead of using the includes as a base for deciding what should be included, use the symbols.

> C libraries are written in so unstandardized way : (

Yeah, I've seen errors when a library is using `#error` to protect itself when includes are not in the correct order. I had to use workarounds like `-include <file>`.

ciechowoj:
> The package name is a separate thing. Instead of using the includes as a base for deciding what should be translated to imports, use the symbols.

Ok, I'll see what I can do with it the way you propose.

ciechowoj:
Another thing:
What do you think about adding libclang bindings as a test case? I would like to make a directory `test_files/libclang` or `test_files/clang-c` and run the tests on it.

jacob-carlborg:
> Ok, I'll see what I can do with it the way you propose.

Unless you've already seen that, this is handled by [IncludeHandler](https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/IncludeHandler.d) which currently keeps a static list of known header files. The `Includehandler` "just" need to be able to work with a dynamic list.

jacob-carlborg:
> What do you think about adding libclang bindings as a test case? I would like to make a directory test_files/libclang or test_files/clang-c and run the tests on it.

Hmm, I would prefer that the things you would like to test are extracted to separate files, like functions, structs and so on. Or do you want a more high level test?

Do you plan to keep those tests static or upgrade them when the bindings DStep uses are upgraded?

ciechowoj:
> Unless you've already seen that, this is handled by IncludeHandler which currently keeps a static list of known header files. The Includehandler "just" need to be able to work with a dynamic list.

Good idea.

> Hmm, I would prefer that the things you would like to test are extracted to separate files, like functions, structs and so on. Or do you want a more high level test?

They would serve as a high level tests.

> Do you plan to keep those tests static or upgrade them when the bindings DStep uses are upgraded?

I'll suppose both options are viable.

jacob-carlborg:
> They would serve as a high level tests.

Would it make sense to reduce the files? I don't think there's a point in having a test like this:

```
void foo(void);
void bar(void);
```

ciechowoj:
> Would it make sense to reduce the files? I don't think there's a point in having a test like this:

I would prefer not to. Doing it by hand is tedious and I do not know about any tool that is able to do that. For me the most appealing part of using libclang for test is that is is already written and covers a lot of cases without effort from the programmer. Another thing is that it is hard to know which things are redundant. The bugs are in very surprising tests sometimes.

ciechowoj:
> The package name is a separate thing. Instead of using the includes as a base for deciding what should be translated to imports, use the symbols.

So, the funny thing is it is written that way. The PR uses the symbols instead of literal include directives to create the import statements. I've written it over a month ago and forgotten. Sorry for confusion.

I've improved the tests to expose the behavior, there is a file "unused.h" that isn't imported despite the fact there is a `#include "unused.h"` directive in main file.

dstep/driver/Application.d
ciechowoj: So I propose to rename it back to `blueprint`. It is hard to say how its end anyway, in the face of Dicebot's changes.

jacob-carlborg:
> I would prefer not to. Doing it by hand is tedious and I do not know about any tool that is able to do that. For me the most appealing part of using libclang for test is that is is already written and covers a lot of cases without effort from the programmer. Another thing is that it is hard to know which things are redundant. The bugs are in very surprising tests sometimes.

I guess a test for libclang can be added.

jacob-carlborg:
> So, the funny thing is it is written that way. The PR uses the symbols instead of literal include directives to create the import statements. I've written it over a month ago and forgotten. Sorry for confusion.

No worries.

> I've improved the tests to expose the behavior, there is a file "unused.h" that isn't imported despite the fact there is a #include "unused.h" directive in main file.

Great 😃.

dstep/driver/Application.d
jacob-carlborg: Sure.

dstep/translator/IncludeHandler.d
```diff
@@ -84,62 +92,120 @@ class IncludeHandler
...
     void addInclude (string include)
     {
-        rawIncludes ~= include;
+        import std.path;
+        import std.file;
+        import std.array;
+
+        auto absolute = include.asAbsNormPath;
+
+        if (absolute != options.inputFile)
```
jacob-carlborg: Is the input file an absolute path?

dstep/translator/IncludeHandler.d
```diff
@@ -84,62 +92,120 @@ class IncludeHandler
...
+            importsBlock(output, unhandled);
 
         output.finalize();
-
-        return output;
     }
 
 private:
 
+    void importsBlock(Output output, string[] imports)
```
jacob-carlborg: Change this to take a range of strings will save you an allocation above.

dstep/translator/IncludeHandler.d
ciechowoj: Yes, it is set in Application.d:69 . [https://github.com/jacob-carlborg/dstep/pull/68/files#diff-906a98df53e7c5cece17f35a0240be48R69](link)

dstep/translator/IncludeHandler.d
ciechowoj: There is sorting inside this function, that prevents me from doing that.

dstep/driver/Application.d
ciechowoj: Done.

ciechowoj:
> I guess a test for libclang can be added.

Great. I'll prepare a separate PR.

test_files/module/main0.d
```diff
@@ -0,0 +1,7 @@
+module modules.main0;
+
+public import modules.include;
```
jacob-carlborg: Is it necessary for imports to be public if we're using the symbols as a base for adding imports instead of the includes?

I'm not sure if we actually want to exactly map the C headers in D. Public imports is usually not idiomatic D, unless it's in a package module.

dstep/translator/IncludeHandler.d
jacob-carlborg: Ok, I see.

dstep/translator/IncludeHandler.d
```diff
@@ -4,18 +4,26 @@
  * Version: Initial created: Jun 15, 2012
  * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
  */
+
```
jacob-carlborg: Please don't add this extra newline.

test_files/module/main0.d
ciechowoj: > Is it necessary for imports to be public if we're using the symbols as a base for adding imports instead of the includes?

Both approaches have its own points. I side with public imports because includes imports symbols as public and I want to avoid confusion from the user of translated library.

It is common pattern in C libraries that there is some main header file and some other internal header files responsible for particular modules (the main header includes sub-headers). And the user is expected to include only the main header file. And he expects that all the symbols from internal header files are available to him through that main header file. I we want to maintain that behavior I believe public imports are better.

However a CLI switch to control how the translated sub-modules are imported (let say `--public-submodules`) can be added to let the user of dstep to decide which behavior he wants.

> I'm not sure if we actually want to exactly map the C headers in D. Public imports is usually not idiomatic D, unless it's in a package module.

What do you mean by 'package module'? A module that gathers all other sub-modules?

dstep/translator/IncludeHandler.d

test_files/module/main0.d
jacob-carlborg: > I want to avoid confusion from the user of translated library

If you look at how Swift imports C/Objective-C libraries, they're trying import the symbols in the most idiomatic way for Swift. For example, many Objective-C method names are renamed to better match Swift names.

> What do you mean by 'package module'? A module that gathers all other sub-modules?

A module with the filename `package.d`, like `std.algorithm` [1].

[1] https://github.com/dlang/phobos/blob/master/std/algorithm/package.d#L183.

jacob-carlborg: > However a CLI switch to control how the translated sub-modules are imported (let say --public-submodules) can be added to let the user of dstep to decide which behavior he wants.

Sounds like a good idea.

ciechowoj: > Sounds like a good idea.

Good, I'll implement it tomorrow.

ciechowoj: Done.

jacob-carlborg:
* Please update the changelog
* Could you please squash the commits. At least the three latest ones.

jacob-carlborg:
I merged Dicebot's PR so there's some conflicts now.

ciechowoj:
Is there something more I can do here? It would be great (for me : ), GSOC related) to have this PR, `issue_46` and `clang-c` merged to mainline before the end of Friday.

jacob-carlborg:
👍

ciechowoj:
Thank you, I'll rebase the rest of PRs.

jacob-carlborg:
Thanks.

> Thank you, I'll rebase the rest of PRs.

Don't add separate commits for that.

> It would be great (for me : ), GSOC related) to have this PR, issue_46 and clang-c merged to mainline before the end of Friday.

I don't think I will be able to merge the others before Friday. I will not be available after today until next week.

ciechowoj:
> I don't think I will be able to merge the others before Friday. I will not be available after today until next week.

I suppose I have to live with it : ).

PR #70, opened 2016-06-25
Stop using tango Application framework

Dicebot:
Step 1 on the way to remove tango based dependencies. This is rather nasty change because command-line handling had to be rewritten completely - @jacob-carlborg would you mind having quick look if logic stays the same before I start cleaning up documentation and coding style?


jacob-carlborg:
What's the rational behind this change?

ZombineDev:
I also gave up on using dstep on windows, because of tango quite some time ago. Please no unnecessary dependencies.

jacob-carlborg:
I suggest you talk to the D core developers instead to increase the quality of the compiler because they have not managed to release a single version of D without introducing regression.

jacob-carlborg:
> I also gave up on using dstep on windows, because of tango quite some time ago. Please no unnecessary dependencies.

DStep has never been verified to work on Windows.

ciechowoj:
> DStep has never been verified to work on Windows.

I'm gonna fix it as part of my GSOC, I suppose removing those dependencies may help a lot.

jacob-carlborg:
@Dicebot I haven't really decided yet. It has to be a significant improvement. Removing the dependencies just to remove them is not a significant improvement in my opinion.

ciechowoj:
Profits of removing dependencies:
- contributors do not have to know additional libraries
- dependencies are undermantained (which sometimes hurts badly)
- easier porting to Windows (and other systems)
- some fixes in DStep require fixing dependencies and it is really painful
- faster build process
- other general benefits of having smaller dependencies, it the face of the fact they do not add much value at the moment (maybe not counting driver code)

What is more the reliance of DStep on the dependencies currently is very small. Not counting the driver code, all the rest can be readily replaced by Phobos equivalents.

jacob-carlborg:
* The behaviour of when an exception is thrown doesn't not look to be the same
* The usage information is not the same

jacob-carlborg:
> Do you want exactly the same output or just the same information to be present there

The same information, ideally not completely different output.

jacob-carlborg:
ciechowo gave some more reasons for removing the dependencies, which I agree with.

The reason for moving some code to separate dependencies is to avoid code duplication. If one only works with one project that uses the dependencies the burden increases. But when working with multiple projects that share the same dependencies the burden decreases. It's pretty annoying to do the same bug fix or workarounds for compiler regressions in multiple projects. Before Dub, at one point, I did duplicate the code (or similar) that is in DStack and Mambo, it was a pain in the ass.

jacob-carlborg:
I think this is an acceptable change. But it needs to retain the current behavior of the application.

joakim-noah:
I think this would be a great change.  I thought about cloning the dstep repo to build it myself a couple years ago but didn't because of the tango dependency.

jacob-carlborg:
In the original code there's a top level try catch that handles exception. Should be in the DStack library.

jacob-carlborg:
Yeah, something like that. I'll do a more thorough review when I have time.

ciechowoj:
If I'm correct this PR removes dependencies only from the 'driver' code. @Dicebot have you done anything about the rest? I could remove Tango from the rest of code, if you do not have time to do it at the moment.

ciechowoj:
Ok, so I'll just wait.

dstep/driver/Application.d
```diff
@@ -26,280 +26,163 @@ import clang.Index;
...
-        arguments.argument("input", "input files").params(1, int.max);
-
-        arguments('o', "output", "Write output to")
-            .params(1)
-            .defaults("");
+        {
+            foreach(string fileName; config.input_files)
+            {
+                import std.exception : enforce;
+                enforce(config.output.length);
```
jacob-carlborg: Please specify an exception type.

dstep/driver/Application.d
```diff
@@ -26,280 +26,163 @@ import clang.Index;
...
-            .params(1)
-            .restrict("c", "c-header", "objective-c", "objective-c-header")
-            .on(&handleLanguage);
+                string outputFilename = Path.buildPath(config.output,
+                    defaultOutputFilename(fileName, false));
 
-        arguments("no-comments", "Disable translation of comments.");
+                string outputDir = Path.dirName(outputFilename);
+                if (!exists(outputDir))
+                    mkdirRecurse(outputDir);
```
jacob-carlborg: Why was this code moved here? It's also duplicated in the `if` and `else` statements.

jacob-carlborg:
The functionality looks fairly ok, but the styling/formatting is not correct.

dstep/driver/Application.d
Dicebot: Where can it possibly be put? This kind of non-trivial path deduction euristic doesn't belong to general argument handling. Code looks rather weird other all the copy-pasting though, I will try to DRY it a bit.

dstep/Configuration.d
```diff
@@ -0,0 +1,27 @@
...
+ * Authors: Jacob Carlborg
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+module dstep.Configuration;
+
+import dstep.translator.Options;
+
+/**
+    Aggregation of global configuration options affecting the program
+ */
```
jacob-carlborg: For single line ddoc comments use `///`. For multiline use:

```d
/**
 *
 */
```

dstep/Configuration.d
```diff
@@ -0,0 +1,27 @@
...
+struct AppConfig
+{
+    /// app version string, generated by the build system
+    enum Version = import("VERSION");
+    /// array of file names to translate to D
```
jacob-carlborg: Please separate with an empty line.

dstep/Configuration.d
```diff
@@ -0,0 +1,27 @@
...
+
+import dstep.translator.Options;
+
+/**
+    Aggregation of global configuration options affecting the program
+ */
+struct AppConfig
+{
+    /// app version string, generated by the build system
+    enum Version = import("VERSION");
```
jacob-carlborg: Please separate with an empty line. That is, after the declaration.

dstep/Configuration.d
```diff
@@ -0,0 +1,27 @@
...
+    /// array of file names to translate to D
+    string[] input_files;
```
jacob-carlborg: Camel case.

dstep/Configuration.d
```diff
@@ -0,0 +1,27 @@
...
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+module dstep.Configuration;
+
+import dstep.translator.Options;
+
+/**
+    Aggregation of global configuration options affecting the program
+ */
+struct AppConfig
```
jacob-carlborg: * Please name this `Configuration`, or `Config`
* When there's a one to one mapping of a declaration and a module please use the same name for the declaration and the module

dstep/driver/Application.d
```diff
@@ -26,280 +26,164 @@ import clang.Index;
...
+        bool singleFileInput = config.input_files.length == 1;
 
-        arguments('x', "language", "Treat subsequent input files as having type <language>")
-            .params(1)
-            .restrict("c", "c-header", "objective-c", "objective-c-header")
-            .on(&handleLanguage);
-
-        arguments("no-comments", "Disable translation of comments.");
+        // parallel generation of D modules for each of input files
+        foreach(string fileName; config.input_files)
```
jacob-carlborg: * Please but a space between `foreach` and the opening parenthesis
* Use type inference, i.e. remove `string`

dstep/driver/Application.d
```diff
@@ -26,280 +26,164 @@ import clang.Index;
...
+        }
     }
 
-    void handleObjectiveC ()
+    static void startParsingFile(const AppConfig config, string fileName,
+        string outputFilename)
     {
-        argsToRestore ~= "-ObjC";
-        language = Language.objC;
+        auto parseFile = ParseFile(config, fileName, outputFilename);
```
jacob-carlborg: No point in having a local variable?

dstep/main.d
```diff
@@ -0,0 +1,174 @@
...
+/**
+    Processes command-line arguments
+
+    Params:
+        args = command-line arguments
+
+    Returns:
+        2-element tuple with first element being aggregate struct of app
+        configuration and second - getopt parse result
+ */
```
jacob-carlborg: Style of ddoc comments as previously mentioned.

dstep/main.d
```diff
@@ -0,0 +1,174 @@
...
+auto parseCLI ( string[] args )
```
jacob-carlborg: Don't add space after the opening and before the closing parentheses.

dstep/main.d
```diff
@@ -0,0 +1,174 @@
...
+{
+    import std.getopt;
+
+    AppConfig config;
+
+    // Parse dstep own paramaters:
+
+    void parseLanguage ( string param, string value )
```
jacob-carlborg: Space here with the spacing.

dstep/main.d
```diff
@@ -0,0 +1,174 @@
...
+        else
+            config.input_files ~= arg;
+    }
+
+    // Post-processing of CLI
+
+    import std.algorithm : canFind;
+
+    if (config.clang_params.canFind("-ObjC"))
+    {
```
jacob-carlborg: Don't use braces for single line if statements.

dstep/main.d
```diff
@@ -0,0 +1,174 @@
...
+    GetoptResult getopt_result = parse_result[1];
+
+    if (getopt_result.helpWanted)
+    {
+        showHelp(config, getopt_result);
+        return 0;
+    }
+
+    import dstep.driver.Application;
+	import std.stdio;
```
jacob-carlborg: Indentation.

dstep/main.d
```diff
@@ -0,0 +1,174 @@
...
+		writeln("An error occurred: ", e);
+		return -1;
+	}
+	catch (Throwable e)
+	{
+		writeln("An unknown error occurred: ", e);
+		throw e;
+	}
+
+	return 0;
```
jacob-carlborg: Indentation.

dstep/main.d
```diff
@@ -0,0 +1,174 @@
...
+        std.getopt.config.passThrough,
+        std.getopt.config.caseSensitive,
+        "output|o", "Write output to", &config.output,
+        "language|x", "Treat subsequent input files as having type <language>", &parseLanguage,
+        "objective-c", "Treat source input file as Objective-C input.", &forceObjectiveC,
+        "no-comments", "Disable translation of comments", &config.no_comments
+    );
+
+    // remove dstep binary name (args[0])
+    args = args[1..$];
```
jacob-carlborg: * Please add spacing around `..`.
* It might be that the name of the binary is required by libclang, i.e. for error reporting

ciechowoj:
What do you think about adding the code styling guide page to Wiki?

joakim-noah:
The best option would be to provide the equivalent [dfmt](https://github.com/Hackerpilot/dfmt) command-line to produce the formatting Jacob wants for Dstep source automatically, with only the formatting dfmt cannot automate listed on the wiki.

jacob-carlborg:
I'll give dfmt a try and see what happens. I would be nice if that dfmt is good enough

dstep/main.d
Dicebot: > It might be that the name of the binary is required by libclang, i.e. for error reporting

I was wondering about that - old code seemed to try passing it through but I couldn't find any place where it is actually used. Do you remember any specific?

dstep/main.d
```diff
@@ -0,0 +1,172 @@
...
+        std.getopt.config.passThrough,
+        std.getopt.config.caseSensitive,
+        "output|o", "Write output to", &config.output,
+        "language|x", "Treat subsequent input files as having type <language>", &parseLanguage,
+        "objective-c", "Treat source input file as Objective-C input.", &forceObjectiveC,
+        "no-comments", "Disable translation of comments", &config.no_comments
+    );
+
+    // remove dstep binary name (args[0])
+    args = args[1 .. $];
```
jacob-carlborg: Could you please check the different behavior of libclang when the name of the binary is included and when it's not. I think it might matter when libclang reports compile errors and warnings.

Dicebot: Tried compiling simple malformed C header, with both old and new build get similar output:
```
File(A267F0, "")/home/dicebot/sample.h:1:1: warning: type specifier missing, defaults to 'int' [-Wimplicit-int]
```

Dicebot: Old code looks like some extra clang output was expected using that name but it seems to never have actually made any impact.

jacob-carlborg: Not pretty, but that's the old behavior.

jacob-carlborg:
This does not actually remove Tango as a dependency. Was that planed for a later PR?

dstep/Configuration.d
```diff
@@ -0,0 +1,32 @@
...
+    enum Version = import("VERSION");
+
+    /// array of file names to translate to D
+    string[] inputFiles;
+
+    /// expected programming language of input files
+    Language language;
+
+    /// array of parameters needed to be forwarded to clang driver
+    string[] clang_params;
```
jacob-carlborg: Camel case.

dstep/Configuration.d
```diff
@@ -0,0 +1,32 @@
...
+
+    /// output file name or folder (in case there are many input files)
+    string output;
+
+    /// strip all comments while translating
+    bool no_comments;
```
jacob-carlborg: Camel case.

dstep/main.d
```diff
@@ -0,0 +1,172 @@
...
+
+/**
+ *  Application entry point, handles CLI/config and forwards to
+ *  dstep.driver.Application to do actual work.
+ */
+int main (string[] args)
+{
+    auto parse_result = parseCLI(args);
+    Configuration config = parse_result[0];
+    GetoptResult getopt_result = parse_result[1];
```
jacob-carlborg: Camel case.

dstep/main.d
```diff
@@ -0,0 +1,172 @@
...
+version (unittest) { }
+else:
+
+/**
+ *  Application entry point, handles CLI/config and forwards to
+ *  dstep.driver.Application to do actual work.
+ */
+int main (string[] args)
+{
+    auto parse_result = parseCLI(args);
```
jacob-carlborg: Camel case.

jacob-carlborg:
LGTM. Could you please squash the commits. If you squash everything into the `Stop using tango Application framework` commit but keep `Auto-generate version string by git tag` and `Add .editorcofig matching current style` as two separate commits. Result would be:

* Stop using tango Application framework
* Auto-generate version string by git tag
* Add .editorcofig matching current style

jacob-carlborg:
👍

leandro-lucarella-sociomantic:
> The reason for moving some code to separate dependencies is to avoid code duplication. If one only works with one project that uses the dependencies the burden increases. But when working with multiple projects that share the same dependencies the burden decreases. It's pretty annoying to do the same bug fix or workarounds for compiler regressions in multiple projects. Before Dub, at one point, I did duplicate the code (or similar) that is in DStack and Mambo, it was a pain in the ass.

I know is irrelevant now, but I really like this article about trade-offs with dependencies: http://yosefk.com/blog/redundancy-vs-dependencies-which-is-worse.html

jacob-carlborg:
I'll add it to my reading list.

Issue #46, opened 2016-01-19
Generating code that will not compile


russel:
This is a one off case but…

The C code:

struct dtv_property {
	__u32 cmd;
	__u32 reserved[3];
	union {
		__u32 data;
		struct dtv_fe_stats st;
		struct {
			__u8 data[32];
			__u32 len;
			__u32 reserved1[3];
			void *reserved2;
		} buffer;
	} u;
	int result;
} __attribute__ ((packed));

results in the D code:

struct dtv_property
{
    __u32 cmd;
    __u32[3] reserved;
    union
    {
        __u32 data;
        struct dtv_fe_stats
        {
            __u8 len;
            dtv_stats[4] stat;
        }
        dtv_fe_stats st;
        struct
        {
            __u8[32] data;
            __u32 len;
            __u32[3] reserved1;
            void* reserved2;
        }
    }
    int result;
}

which is compile error not only because of the __u32 and __u8 types, but because of the field data in the anonymous struct.

I do not have any answers, but I think this marker for the problems needs to be put on record.

jacob-carlborg:
You mean because `buffer` and `u` is removed?

russel:
Indeed. removal of the buffer and u flattens the name space and creates a conflict over the name data that wasn't there before.

Also __u32 and __u8 are not being converted to their D forms.

jacob-carlborg:
What are `__u32` and `__u8`? If they're `#define` than they will not be handled.

PR #71, opened 2016-06-27
Fix 46: Generating code that will not compile.

ciechowoj:
Fixes issue 46 (some of problems from the issue were solved by previous PRs).

Add `dont-reduce-aliases` command line switch. Without the switch aliases for basic arithmetic types are not translated and types are directly translated to appropriate basic type e.g.: `uint8_t -> ubyte`. With the switch the aliased named are kept. Some exceptions are made for idiomatic types like `size_t` or `ptrdiff_t`, which are typedefs but have built-in equivalent in D.

### DIFF
--- a/dstep/translator/IncludeHandler.d
+++ b/dstep/translator/IncludeHandler.d
@@ -42,6 +42,14 @@ class IncludeHandler
             "stdarg" : "core.stdc.stdarg",
             "stddef" : "core.stdc.stddef",
             "stdint" : "core.stdc.stdint",
+            "_int8_t" : "core.stdc.stdint",
+            "_int16_t" : "core.stdc.stdint",
+            "_int32_t" : "core.stdc.stdint",
+            "_int64_t" : "core.stdc.stdint",
+            "_uint8_t" : "core.stdc.stdint",
+            "_uint16_t" : "core.stdc.stdint",
+            "_uint32_t" : "core.stdc.stdint",
+            "_uint64_t" : "core.stdc.stdint",
             "stdio" : "core.stdc.stdio",
             "stdlib" : "core.stdc.stdlib",
             "string" : "core.stdc.string",
@@ -80,6 +88,7 @@ class IncludeHandler
             "sys/socket" : "core.sys.posix.sys.socket",
             "sys/stat" : "core.sys.posix.sys.stat",
             "sys/time" : "core.sys.posix.sys.time",
+            "_time_t" : "core.stdc.time",
             "sys/types" : "core.sys.posix.sys.types",
             "sys/_types" : "core.sys.posix.sys.types",
             "sys/uio" : "core.sys.posix.sys.uio",
@@ -140,26 +149,26 @@ class IncludeHandler
         import std.array : array;
         import std.format : format;
 
-        string[] standard, package_, unhandled;
+        Set!string standard, package_, unhandled;
 
         foreach (entry; includes.byKey)
         {
             if (auto i = isKnownInclude(entry))
-                standard ~= toImport(i);
+                standard.add(toImport(i));
             else if (auto i = isPackageSubmodule(entry))
-                package_ ~= toSubmoduleImport(i);
+                package_.add(toSubmoduleImport(i));
             else
-                unhandled ~= format(`/+ #include "%s" +/`, entry);
+                unhandled.add(format(`/+ #include "%s" +/`, entry));
         }
 
         auto extra = imports.byKey.map!(e => toImport(e)).array;
 
-        importsBlock(output, standard);
-        importsBlock(output, extra);
-        importsBlock(output, package_);
+        importsBlock(output, standard.keys);
+        importsBlock(output, extra.array);
+        importsBlock(output, package_.keys);
 
         if (options.keepUntranslatable)
-            importsBlock(output, unhandled);
+            importsBlock(output, unhandled.keys);
 
         output.finalize();
     }
--- a/dstep/translator/MacroDefinition.d
+++ b/dstep/translator/MacroDefinition.d
@@ -1339,10 +1339,9 @@ bool parseTypedefName(ref TokenRange tokens, ref Type type, Cursor[string] table
     {
         if (auto ptr = (spelling in table))
         {
-            type.kind = CXTypeKind.CXType_Typedef;
-            type.spelling = spelling;
-
+            type = Type.makeTypedef(spelling, ptr.type.canonical);
             tokens = local;
+
             return true;
         }
     }
@@ -1513,6 +1512,7 @@ bool basicSpecifierListToType(ref Type type, Set!string specifiers)
                 return false;
 
             type = Type(CXTypeKind.CXType_LongDouble, "long double");
+
             return true;
         }
 
@@ -1520,6 +1520,7 @@ bool basicSpecifierListToType(ref Type type, Set!string specifiers)
             return false;
 
         type = Type(CXTypeKind.CXType_Double, "double");
+
         return true;
     }
 
--- a/dstep/translator/Type.d
+++ b/dstep/translator/Type.d
@@ -121,32 +121,69 @@ string translateSelector (string str, bool fullName = false, bool translateIdent
     return translateIdentifier ? .translateIdentifier(str) : str;
 }
 
+package string reduceAlias(Type type)
+{
+    auto spelling = type.spelling;
+
+    switch (type.spelling)
+    {
+        case "BOOL": return "bool";
+        case "int8_t": return "byte";
+        case "int16_t": return "short";
+        case "int32_t": return "int";
+        case "int64_t": return "long";
+        case "uint8_t": return "ubyte";
+        case "uint16_t": return "ushort";
+        case "uint32_t": return "uint";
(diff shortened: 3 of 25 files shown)


dstep/translator/Type.d
```diff
@@ -121,32 +121,114 @@ string translateSelector (string str, bool fullName = false, bool translateIdent
...
+string translateLongAlike(Context context, Type type)
+{
+    auto kind = type.canonical.kind;
+
+    if (kind == CXTypeKind.CXType_Long)
+    {
+        switch (type.spelling)
+        {
+            case "int64_t":
+            case "__s64":
```
jacob-carlborg: Is this standard/common typedef? Not sure if I like to add too much special cases.

dstep/translator/Type.d
```diff
@@ -121,32 +121,114 @@ string translateSelector (string str, bool fullName = false, bool translateIdent
...
 private:
 
-string translateTypedef (Context context, Type type)
-in
+bool isIdiomatic(Context context, Type type)
 {
-    assert(type.kind == CXTypeKind.CXType_Typedef);
+    switch (type.spelling)
+    {
+        case "NSInteger":
```
jacob-carlborg: If you're adding `NSInteger` then there are `NSUInteger` and `CGFloat` as well.

ciechowoj: Fixed, I have zero experience with Objective-C.

dstep/translator/Type.d
ciechowoj: Yes, it is standard type in linux system headers.

dstep/translator/Type.d
ciechowoj: Eh, the tests doesn't pass on mac for some strange reason.

ciechowoj: I have issues with `CGFloat`. I do not know in which header it is declared. I tried `<Foundation/Foundation.h>` but then it stops to pass for some another reason.

I can leave it without tests if it ins't a problem.

jacob-carlborg: Is the test compiling as C or Objective-C code?
On OS X `NSInteger` is defined in `<objc/NSObjCRuntime.h>`. Officially [Foundation](https://developer.apple.com/library/ios/documentation/Cocoa/Reference/Foundation/Miscellaneous/Foundation_DataTypes/#//apple_ref/c/tdef/NSInteger) should be included. 
CGFloat is defined in [CoreGraphics](https://developer.apple.com/library/ios/documentation/GraphicsImaging/Reference/CGGeometry/index.html#//apple_ref/c/tdef/CGFloat).

ciechowoj: Is CoreGraphics a part of GNUStep?

ciechowoj: Seems it isn't. I can try to enable the tests for `CGFloat` only for MacOS.

ciechowoj: Done.

unit_tests/IssueUnitTests.d
```diff
@@ -0,0 +1,108 @@
+/**
```
jacob-carlborg: Perhaps this should be but in a specific sub directory and name the file `Issue46`.

ciechowoj: Like `test_files/Issue46.h (/.d)` ?

jacob-carlborg: No, the `unit_test` directory can still be used. Something like `unit_tests/issues/Issue46.d`.

ciechowoj: Ok.

ciechowoj: Done.

test_files/objc/primitives.d
```diff
@@ -1,6 +1,10 @@
+import core.stdc.time;
```
jacob-carlborg: Why is this imported?

dstep/translator/MacroDefinition.d
```diff
@@ -1339,8 +1339,9 @@ bool parseTypedefName(ref TokenRange tokens, ref Type type, Cursor[string] table
     {
         if (auto ptr = (spelling in table))
         {
-            type.kind = CXTypeKind.CXType_Typedef;
-            type.spelling = spelling;
+            type = Type.makeTypedef(
+                spelling,
+                ptr.type.canonical);
```
jacob-carlborg: Doesn't this statement fit on one line?

jacob-carlborg:
If one of the commit messages contain `Fix #46` it will automatically close the issue when merged.

test_files/objc/primitives.d
ciechowoj: It is an old bug in imports translation. A matching import is searched with `endsWith`

```
    string isKnownInclude (string include)
    {
        include = Path.stripExtension(include);

        if (auto r = knownIncludes.find!((k, _) => include.endsWith(k)))
            return r.value;

        return null;
    }
```

it creates an ambiguity when there are two or more matches with the same suffix (e.g. `runtime.h` and `time.h`). And as linear search is used first one is used arbitrarily. 

It needs improvement, but I don't want to fix it in this PR.

jacob-carlborg: Interesting. So this is due to `NSInteger` being defined in `NSObjCRuntime.h`?

ciechowoj: Almost for sure.

ciechowoj:
I squashed everything into one commit and added the `#`.

dstep/translator/IncludeHandler.d
```diff
@@ -140,26 +149,26 @@ class IncludeHandler
...
-                unhandled ~= format(`/+ #include "%s" +/`, entry);
+                unhandled.add(format(`/+ #include "%s" +/`, entry));
         }
 
         auto extra = imports.byKey.map!(e => toImport(e)).array;
 
-        importsBlock(output, standard);
-        importsBlock(output, extra);
-        importsBlock(output, package_);
+        importsBlock(output, standard.byKey.array);
```
jacob-carlborg: You can call `keys` instead of `byKey.array`.

resources/VERSION
```diff
@@ -0,0 +1 @@
+v0.2.2-59-g66fe26a
```
jacob-carlborg: Should this file be committed? @Dicebot ?

jacob-carlborg:
The issue is fixed by adding a new feature, which is controlled by the `dont-reduce-aliases` flag, which I think should be mentioned in the changelog as well.

dstep/translator/IncludeHandler.d

resources/VERSION
ciechowoj: This is a good question : )

ciechowoj:
> The issue is fixed by adding a new feature, which is controlled by the dont-reduce-aliases flag, which I think should be mentioned in the changelog as well.

DONE.

ciechowoj:
> > Yes, which can be extended if needed in the future.
> Ok, let it be. I'll make required changes soon (before the end of week I suppose).

I've just implemented the required changes.

ciechowoj:
> I've just implemented the required changes.
Or not. For some reason GitHub doesn't want to update to them.

ciechowoj:
It should be ok now.

jacob-carlborg:
So how is the new behavior?

ciechowoj:
By default the types are translated as they are. With `reduce-aliases` flag the known types are reduced to it's D equivalents. Currently reduced types follows patterns: `uintxx_t`, `__uxx` and `uxx`.

ciechowoj:
Yes, the ones listed in some big switch-case.

jacob-carlborg:
For me it would be fine to reduce known types by default. The issue I mentioned was for reducing unknown types by default.

ciechowoj:
So, if I negate the switch to `dont-reduce-aliases`, will it be fine?
I thought that reducing unknown types is just a no-no for dstep.

ciechowoj:
The reducing unknown types isn't required to close the issue anyway...

jacob-carlborg:
> I thought that reducing unknown types is just a no-no for dstep.

I'm not sure what's best. Lets start with not supporting that. Negate the switch to `dont-reduce-aliases` and it's fine.

ciechowoj:
Done.

resources/VERSION
```diff
@@ -0,0 +1 @@
+v0.2.2-60-g8ffa97f
```
jacob-carlborg: @Dicebot should this file be included in Git?

jacob-carlborg:
Looking good. I just want to get a reply from @Dicebot about the `VERSION` file.

ciechowoj:
Ok.

resources/VERSION
Dicebot: No, it is auto-generated on each `dub build` call. Can be added to `.gitignore` though.

ciechowoj: I removed the file and added it to `.gitignore`.

jacob-carlborg:
Please squash commits and I can merge this PR.

ciechowoj:
Done.

jacob-carlborg:
Thanks.

ciechowoj:
Thank you.

Issue #10, opened 2013-06-01
Embedded struct not generated.


remy-j-a-moueza:
An embedded struct within another struct does not get generated. 

On the mongoose.h header, `mg_request_info`  containing a ` mg_header`  struct: 
```c
// This structure contains information about the HTTP request.
struct mg_request_info {
  const char *request_method; // "GET", "POST", etc
  const char *uri;            // URL-decoded URI
  const char *http_version;   // E.g. "1.0", "1.1"
  const char *query_string;   // URL part after '?', not including '?', or NULL
  const char *remote_user;    // Authenticated user, or NULL if no auth used
  long remote_ip;             // Client's IP address
  int remote_port;            // Client's port
  int is_ssl;                 // 1 if SSL-ed, 0 if not
  void *user_data;            // User data pointer passed to mg_start()

  int num_headers;            // Number of HTTP headers
  struct mg_header {
    const char *name;         // HTTP header name
    const char *value;        // HTTP header value
  } http_headers[64];         // Maximum 64 headers
};

```

What gets generated is: 
```c
struct mg_request_info
{
	const(char)* request_method;
	const(char)* uri;
	const(char)* http_version;
	const(char)* query_string;
	const(char)* remote_user;
	c_long remote_ip;
	int remote_port;
	int is_ssl;
	void* user_data;
	int num_headers;
	mg_header[64] http_headers;
}
```

`mg_header` is left undefined. 

I would have expected a definition, at least outside of mg_request_info like:
```c
struct mg_header {
    const(char) *name;
    const(char) *value;
}
```

Issue #59, opened 2016-04-07
Shouldn't dstep exit with status code when there is some kind of error?


ciechowoj:
Currently `dstep` always returns `0` exit code, even on the failure (wrong input file/wrong command arguments). I believe it would be better, if it returns something other (e.g. 1) when there is an error. I would make `dstep` easier to integrate with some command line tools (and make testing easier).

jacob-carlborg:
Absolutely. It should exit with a non-zero status code if an exception has been thrown. Although, I'm not sure if an exception is thrown in all the right places.

Issue #38, opened 2015-03-18
spurious generation of variadic args rather than implicit void ie () argument list


Laeeth:
converting the mathgl headers

I got:
double mgl_rnd_ (...);

but it should just be:
double mgl_rnd();

hope this is not something simple I am doing wrong.  sorry, but I have not time right now to find the original header and try to replicate it.

jacob-carlborg:
I'm using quite an old version of Clang now, but I can only get this to compile in C++ mode. DStep does not support C++.

Laeeth:
It's one of those strange cases where there is a pure C API documented in the headers but it is mixed in with C++.  So  dstep did an okay job, but got confused on those small cases.

jacob-carlborg:
I'll need to investigate how AST in libclang looks like.

Issue #21, opened 2014-02-15
wchar_t should be tranlsated to core.stdc.stddef.wchar_t


jacob-carlborg:
Currently `wchar_t` translates to `wchar` or `dchar` depending on its size. While not incorrect it would be better to translate it to `core.stdc.stddef.wchar_t` instead. This would make the bindings more cross-platform.

PR #72, opened 2016-06-27
Issue 10

ciechowoj:
Fixes issue #10 and issue #21 . The commits are independent.


ciechowoj:
The tests are not passing but I fixed the cause in some earlier, still not merged branch (#71 if I remember correctly). I'll fix it after merging that other branch.

clang/Type.d
```diff
@@ -78,6 +78,21 @@ struct Type
             kind == CXTypeKind.CXType_DependentSizedArray;
     }
 
+    @property Type undecorated()
```
jacob-carlborg: What's the purpose of this method?

dstep/translator/Record.d
```diff
@@ -34,6 +34,8 @@ void translatePackedAttribute(Output output, Context context, Cursor cursor)
         output.singleLine("align (1):");
 }
 
+import std.stdio;
```
jacob-carlborg: Leftover from debugging?

unit_tests/IssueUnitTests.d
```diff
@@ -106,3 +105,43 @@ struct property_t
 D");
 
 }
+
+// Fix 10: Embedded struct not generated.
+unittest
```
jacob-carlborg: This should follow the same structure as in #71. A minimal unit test for structs and the full test case for the issue in a separate file.

unit_tests/RecordUnitTests.d
```diff
@@ -122,3 +122,167 @@ struct Foo
...
+    Baz data[32];
+    char len;
+};
+C",
+q"D
+extern (C):
+
+struct Foo
+{
+    int[32] data;
```
jacob-carlborg: I thought aliases were not reduced by default?

dstep/driver/Application.d
```diff
@@ -89,29 +89,8 @@ class Application
...
-        arguments('x', "language", "Treat subsequent input files as having type <language>")
-            .params(1)
-            .restrict("c", "c-header", "objective-c", "objective-c-header")
-            .on(&handleLanguage);
-
-        arguments("no-comments", "Disable translation of comments.");
-        arguments("dont-reduce-aliases", "Disable reduction of primitive type aliases.");
-
-        arguments("objective-c", "Treat source input file as Objective-C input.");
->>>>>>> Fix 46: Generating code that will not compile.
```
jacob-carlborg: Please clean this up.

unit_tests/RecordUnitTests.d
ciechowoj: The switch is `dont-reduce-aliases` (vs `reduce-aliases`).

dstep/translator/Record.d
ciechowoj: Yes, fixed.

unit_tests/IssueUnitTests.d
ciechowoj: Done.

dstep/driver/Application.d
ciechowoj: Cleaned.

clang/Type.d
ciechowoj: I added following documentation:
```
    /**
     * Removes array and pointer modifiers from the type.
     */
```

unit_tests/RecordUnitTests.d
ciechowoj: So, it is enabled by default.

jacob-carlborg: Hmm, I thought it was only for known aliases like `uint8_t`. The issue is if the typedef is different due to some condition, i.e. which platform.

ciechowoj: > The issue is if the typedef is different due to some condition, i.e. which platform.
Good point. However such cases should be quite rare. I propose to make the previous behavior the default and the reduction will be enabled by a switch.

jacob-carlborg: I don't agree. For basic types this seems to be a major use case. Why would one otherwise use a typedef?

ciechowoj: The only viable solution would be to detect that there is another typedef for given type in not take branch of some `#ifdef` (or similar preprocessor directive) .

ciechowoj: > I don't agree. For basic types this seems to be a major use case. Why would one otherwise use a typedef?

But how to know that in advance? The idiom to make own fixed-width integers is quite popular in C. There is no way to list all that types in code.

jacob-carlborg: > But how to know that in advance? The idiom to make own fixed-width integers is quite popular in C. There is no way to list all that types in code.

Exactly. If the alias is reduced you can get different results depending on which platform the binding is generated on.

ciechowoj: > Exactly. If the alias is reduced you can get different results depending on which platform the binding is generated on.

Ok, you convinced me. However I still don't like that only the standard types would be reduced. I think there should be a way for user to specify which aliases he want to reduce and to what type.

I mean time I can change it to reduce only types from `stdint.h` and from linux headers. Maybe with two CLI switches `reduce-stdint-aliases` and `reduce-__sxx-types`.

jacob-carlborg:
> Ok, you convinced me. However I still don't like that only the standard types would be reduced. 

It's only the standard types we can know about.

> I think there should be a way for user to specify which aliases he want to reduce and to what type

Not sure I like that. Seems to get quite complicated now.

> I mean time I can change it to reduce only types from stdint.h and from linux headers. Maybe with two CLI switches reduce-stdint-aliases and reduce-__sxx-aliases.

Do we really need two flags?

ciechowoj:
> Do we really need two flags?

I would like to make it explicit which types are reduced exactly. E.g. make it clear that it will not work for other aliases following that idiom (`GLuint`, `GLshort`).

Or maybe we can make a single flag and write in documentation which aliases from which libraries are reduced.

jacob-carlborg:
> Or maybe we can make a single flag and write in documentation which aliases from which libraries are reduced.

Yes, which can be extended if needed in the future.

ciechowoj:
> Yes, which can be extended if needed in the future.

Ok, let it be. I'll make required changes soon.

jacob-carlborg:
> Ok, let it be. I'll make required changes soon.

👍

ciechowoj:
Anyways, there are some other PRs that do not depend on `issue_46` just waiting for a review. Just saying : ) ...

jacob-carlborg:
> Anyways, there are some other PRs that do not depend on issue_46 waiting for a review. Just saying : )

Oh, I know 😃. Could you specify which are dependent on `issue_46`?

ciechowoj:
Dependent:
#79 Issue 58 
#75 Header comment detection improvement 
#72 Issue 10
#71 Fix 46: Generating code that will not compile. 

Independent.
#84 Fix #83 and #30.
#82 Add CONTIBUTING.md. 
#81 Fix #59 and #28. 
#80 Fix #38: Spurious generation of variadic args. 
#77 Clang c 
#76 Implement translation of ## operator.

jacob-carlborg:
BTW, do you have a schedule for the deadlines for GSoC? Then I can know which PR's I should prioritize.

ciechowoj:
The GSOC deadline is 23 August, which is very close. However, it would be great to have merged as much as possible before 15 August. As the deadline is so close, currently it doesn't really matter which PRs will make their way to mainline first. I see, that is a lot of work to review all that, so I believe you shouldn't worry to much about it : ), it is my problem. If not everything is merged at the end, I would need to find some other solution.

I'm still working on two big features right now:
+ improve support for MS Windows
+ make untranslated entities to appear in dstep output (e.g. not taken branches of preprocessor directives)

I also said in my proposal that I'll "add cross­platform testing environment", I would like to fulfill it by setting up a CI server to test dstep on MS Windows, I thought about AppVeyor. But maybe I'll do it on my own clone of dstep.

jacob-carlborg:
> The GSOC deadline is 23 August, which is very close. However, it would be great to have merged as much as possible before 15 August. As the deadline is so close, currently it doesn't really matter which PRs will make their way to mainline first. I see, that is a lot of work to review all that, so I believe you shouldn't worry to much about it : ), it is my problem.

I guess it's a question of which PR's should be prioritized.

> If not everything is merged at the end, I would need to find some other solution.

What do you mean "find some other solution"?

> I also said in my proposal that I'll "add cross­platform testing environment", I would like to fulfill it by setting up a CI server to test dstep on MS Windows, I thought about AppVeyor. But maybe I'll do it on my own clone of dstep.

That would be very nice to have once DStep works on Windows.

> Actually it would be nice to have libclang tests integrated with mainline before 15 August more that anything else.

Could you please prioritize the PR's in which order I should review.

ciechowoj:
> I guess it's a question of which PR's should be prioritized.

Whatever : )

> What do you mean "find some other solution"?

Provide merged version on my own fork of `dstep` for example. I would need to talk with my mentor about this.

> Could you please prioritize the PR's in which order I should review.

#77 Clang c 
#84 Fix #83 and #30.
#81 Fix #59 and #28. 
#80 Fix #38: Spurious generation of variadic args. 
#82 Add CONTIBUTING.md. 
#76 Implement translation of ## operator.
#71 Fix 46: Generating code that will not compile.
#72 Issue 10
#75 Header comment detection improvement 
#79 Issue 58

jacob-carlborg:
> Provide merged version on my own fork of dstep for example. I would need to talk with my mentor about this.

Provide in what way? What would be the gain of this?

ciechowoj:
> Provide in what way? What would be the gain of this?

I think it would be nice, if after the GSOC deadline all improvements I've made were available somewhere ready for use right away, instead of being scattered among multiple branches. It is only relevant in GSOC context. I'll go to the extra mile to pull everything here, eventually.

jacob-carlborg:
> I think it would be nice, if after the GSOC deadline all improvements I've made were available somewhere ready for use right away, instead of being scattered among multiple branches. It is only relevant in GSOC context

Ok, I see.

> I'll go to the extra mile to pull everything here, eventually.

What wold be great, because you're doing a lot of great work here.

resources/VERSION
```diff
@@ -0,0 +1 @@
+v0.2.2-61-g73df69b
```
jacob-carlborg: Should not be included.

jacob-carlborg:
👌

ciechowoj:
Great : )

PR #75, opened 2016-07-18
Header comment detection improvement

ciechowoj:
Improves detection of the header comment. From now the header comments are comment before include guard (if present).

### DIFF
--- a/clang/Diagnostic.d
+++ b/clang/Diagnostic.d
@@ -125,3 +125,30 @@ struct DiagnosticSet
         return length;
     }
 }
+
+CXDiagnosticSeverity severity(DiagnosticSet diagnostics)
+{
+    import std.algorithm.searching : minPos;
+    import std.algorithm.iteration : map;
+
+    bool less(CXDiagnosticSeverity a, CXDiagnosticSeverity b)
+    {
+        return cast(uint) a > cast(uint) b;
+    }
+
+    if (diagnostics.empty)
+        return CXDiagnosticSeverity.CXDiagnostic_Ignored;
+    else
+        return diagnostics.map!(diagnostic => diagnostic.severity).minPos!less.front;
+}
+
+bool hasError(DiagnosticSet diagnostics)
+{
+    auto severity = diagnostics.severity;
+
+    if (severity == CXDiagnosticSeverity.CXDiagnostic_Error ||
+        severity == CXDiagnosticSeverity.CXDiagnostic_Fatal)
+        return true;
+    else
+        return false;
+}
--- a/clang/Diagnostic.d
+++ b/clang/Diagnostic.d
@@ -131,10 +131,7 @@ CXDiagnosticSeverity severity(DiagnosticSet diagnostics)
     import std.algorithm.searching : minPos;
     import std.algorithm.iteration : map;
 
-    bool less(CXDiagnosticSeverity a, CXDiagnosticSeverity b)
-    {
-        return cast(uint) a > cast(uint) b;
-    }
+    alias less = (a, b) => cast(uint) a > cast(uint) b;
 
     if (diagnostics.empty)
         return CXDiagnosticSeverity.CXDiagnostic_Ignored;
@@ -146,9 +143,6 @@ bool hasError(DiagnosticSet diagnostics)
 {
     auto severity = diagnostics.severity;
 
-    if (severity == CXDiagnosticSeverity.CXDiagnostic_Error ||
-        severity == CXDiagnosticSeverity.CXDiagnostic_Fatal)
-        return true;
-    else
-        return false;
+    return severity == CXDiagnosticSeverity.CXDiagnostic_Error ||
+        severity == CXDiagnosticSeverity.CXDiagnostic_Fatal;
 }
--- a/clang/TranslationUnit.d
+++ b/clang/TranslationUnit.d
@@ -25,6 +25,8 @@ struct TranslationUnit
 {
     mixin CX;
 
+    string sourceString = null;
+
     static TranslationUnit parse (
         Index index,
         string sourceFilename,
@@ -67,6 +69,8 @@ struct TranslationUnit
 
         remove(name);
 
+        translationUnit.sourceString = source;
+
         return translationUnit;
     }
 
@@ -108,6 +112,12 @@ struct TranslationUnit
     @property string spelling ()
     {
         return toD(clang_getTranslationUnitSpelling(cx));
+    }
+
+    @property string source ()
+    {
+        import std.file : readText;
+        return sourceString ? sourceString : readText(spelling);
     }
 
     @property Cursor cursor ()
--- a/clang/TranslationUnit.d
+++ b/clang/TranslationUnit.d
@@ -355,6 +355,11 @@ struct TranslationUnit
         return isFileMultipleIncludeGuarded(spelling);
     }
 
+    TokenRange tokens()
+    {
+        return tokenize(extent(0, cast(uint) source.length));
+    }
+
     string dumpAST(bool skipIncluded = true)
     {
         import std.array : appender;
--- a/clang/TranslationUnit.d
+++ b/clang/TranslationUnit.d
@@ -245,17 +245,104 @@ struct TranslationUnit
         return relativeLocationAccessorImpl(cursor.all);
     }
 
-    TokenRange tokenize(SourceRange extent)
+    private struct TokenRange
     {
-        import std.algorithm.mutation : stripRight;
+        CXTranslationUnit cx;
+        CXToken* tokens;
+        uint numTokens;
+        uint currentToken;
 
-        CXToken* tokens = null;
-        uint numTokens = 0;
-        clang_tokenize(cx, extent.cx, &tokens, &numTokens);
-        auto result = TokenRange(cx, tokens, numTokens);
+        Token makeToken(CXToken token)
+        {
+            return Token(
+                clang_getTokenKind(token).toD,
+                clang_getTokenSpelling(cx, token).toD,
+                SourceRange(clang_getTokenExtent(cx, token)));
+        }
+
+        Token front()
+        {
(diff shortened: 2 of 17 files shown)


clang/TranslationUnit.d
```diff
@@ -245,17 +245,33 @@ struct TranslationUnit
...
 
         CXToken* tokens = null;
         uint numTokens = 0;
         clang_tokenize(cx, extent.cx, &tokens, &numTokens);
-        auto result = TokenRange(cx, tokens, numTokens);
+
+        TokenRange range;
+        range.length = numTokens;
+
+        for (size_t i = 0; i < numTokens; ++i)
```
jacob-carlborg: A `foreach` range loop can be used instead..

jacob-carlborg: Or perhaps `iota` + `map`.

jacob-carlborg:
What's the purpose of changing the implementation of `TokenRange`?

clang/TranslationUnit.d
ciechowoj: Fixed.

ciechowoj:
> What's the purpose of changing the implementation of TokenRange?

The previous implementation was slightly over-complicated because I wanted to avoid allocation of the array. The `Token` in previous implementation held a reference to the whole array (through reference counting), so even if I kept only one `Token` from `TokenRange` the whole range needed to exist in memory. In previous application it wasn't problem, but now I want to do filtering on tokens and I need true value semantic for the `Token`.

jacob-carlborg:
Can't you wrap `clang_tokenize` in a range API? Or is the issue that it needs to be disposed? `tokenizeNoComments` would be a lot simpler if `tokenize` returned a range and `tokenizeNoComments` could use that range and apply a filter. These tokenize functions are pretty similar and sharing some code.

ciechowoj:
Ok, I'll give a try. The reason I haven't used a filter on `tokenizeNoComments` is that comments can pose a significant part of all the tokens and then I can save on memory allocation (I can use smaller array for output). 

However the solution with range seems to have that property.

ciechowoj:
If you are worried about much of repeated code in `tokenizeNoComments` I can create a simple range type (or visitor maybe, does algorithms support `opApply` ? ) only for sake of returning it from `tokenize` and `feeding` to filter and then I'll convert the result for `filter` to array.

jacob-carlborg:
> If you are worried about much of repeated code in tokenizeNoComments I can create a simple range type (or visitor maybe, does algorithms support opApply ? ) only for sake of returning it from tokenize and feeding to filter and then I'll convert the result for filter to array.

It was something like that I was thinking about. (algorithms does not support opApply). Here's an untested version:

```d
private struct TokenRange
{
    CXTranslationUnit cx;
    CXToken* tokens;
    uint numTokens;
    uint currentToken;

    Token front()
    {
        return makeToken(tokens[currentToken]);
    }
    
    bool empty()
    {
        return numTokens == 0 || numTokens == currentToken;
    }
    
    void popFront()
    {
        currentToken++;
    }
    
    void dispose()
    {
        clang_disposeTokens(cx, tokens, numTokens);
    }
}

private TokenRange tokenizeImpl(CXTranslationUnit cx, SourceRange extent)
{
    auto range = TokenRange(cx);
    clang_tokenize(cx, extent.cx, &range.tokens, &range.numTokens);
    return range;
}

Token[] tokenize(CXTranslationUnit cx, SourceRange extent)
{
    auto range = tokenizeImpl(cx, extent);
    auto tokens = range.array;
    range.dispose();

    return tokens;
}

Token[] tokenizeNoComments(CXTranslationUnit cx, SourceRange extent)
{
    auto range = tokenizeImpl(cx, extent);
    auto tokens = range.filter(e => e.kind != CXTokenKind.CXToken_Comment).array;
    range.dispose();

    return tokens;
}
```

ciechowoj:
Ok, I'll change this your way, but why have you used `dispose` instead of a destructor ?

ciechowoj:
Changed.

jacob-carlborg:
> Ok, I'll change this your way, but why have you used dispose instead of a destructor ?

I didn't think of using a destructor. But that would require guaranteeing that the struct is moved out of `tokenizeImpl` otherwise the destructor would be run inside `tokenizeImpl`.

ciechowoj:
> I didn't think of using a destructor. But that would require guaranteeing that the struct is moved out of tokenizeImpl otherwise the destructor would be run inside tokenizeImpl.

I suppose the current solution is enough for a private struct.

clang/Diagnostic.d
```diff
@@ -125,3 +125,30 @@ struct DiagnosticSet
...
+        return CXDiagnosticSeverity.CXDiagnostic_Ignored;
+    else
+        return diagnostics.map!(diagnostic => diagnostic.severity).minPos!less.front;
+}
+
+bool hasError(DiagnosticSet diagnostics)
+{
+    auto severity = diagnostics.severity;
+
+    if (severity == CXDiagnosticSeverity.CXDiagnostic_Error ||
```
jacob-carlborg: You can return the boolean condition immediately, no need for an if-statement.

clang/Diagnostic.d
```diff
@@ -125,3 +125,30 @@ struct DiagnosticSet
...
+
+CXDiagnosticSeverity severity(DiagnosticSet diagnostics)
+{
+    import std.algorithm.searching : minPos;
+    import std.algorithm.iteration : map;
+
+    bool less(CXDiagnosticSeverity a, CXDiagnosticSeverity b)
+    {
+        return cast(uint) a > cast(uint) b;
+    }
```
jacob-carlborg: I think using a nested function will cause an unnecessary closure. You can use static function, or even better an alias instead https://dlang.org/changelog/2.070.0.html#alias-funclit.

clang/TranslationUnit.d
```diff
@@ -245,17 +255,104 @@ struct TranslationUnit
...
+        import std.array : array;
+        import std.algorithm : stripRight;
+
+        auto range = tokenizeImpl(cx, extent);
+        auto tokens = range.array;
+        range.dispose();
 
         // For some reason libclang returns some tokens out of cursors extent.cursor
-        return result.stripRight!(token => !intersects(extent, token.extent));
+        return tokens.stripRight!(token => !intersects(extent, token.extent));
```
jacob-carlborg: Performing this on the range instead of the array might be more efficient.

dstep/translator/MacroIndex.d
```diff
@@ -87,4 +93,43 @@ class MacroIndex
 
         return result.data;
     }
+
+    Tuple!(bool, SourceLocation) includeGuardLocation()
+    {
+        import std.range.primitives : empty;
+
+        bool checkIfndef(ConditionalDirective directives, string identifier)
```
jacob-carlborg: Prefer to use static nested functions if no closure is required.

dstep/translator/Preprocessor.d
```diff
@@ -0,0 +1,935 @@
...
+            case "pragma":
+            case "undef":
+                return true;
+
+            default:
+                return false;
+        }
+    }
+
+    this(Token[] tokens, string source = null)
```
jacob-carlborg: Prefer to put the constructor first of the functions and private functions last.

dstep/translator/Preprocessor.d
```diff
@@ -0,0 +1,935 @@
...
+    return 42;
+}
+D").array;
+
+    assert(x2.length == 0);
+
+
+    auto x3 = tokenizedDirectives(q"D
+#define FOO
+D").array;
```
jacob-carlborg: Please use `q"C` when it's C code that is parsed.

clang/TranslationUnit.d
ciechowoj: It requires a `save` method (and bunch of other) to be added to `TokenRange` as stripRight requires bidirectional range. The problem with `save` is that it requires to copy the state to another `TokenRange` and we do not have a proper resource managements (I mean `dispose` needs to be called explicitly).

It is possible to add save method however the `TokenRange` returned by `save` will not be the owner or internal data. Though, it should work in our case.

jacob-carlborg: It's not strictly necessary. We don't need to over complicate things. I thought it would be possible with an input range.

clang/Diagnostic.d
ciechowoj: Fixed.

clang/Diagnostic.d
ciechowoj: Fixed.

clang/TranslationUnit.d
ciechowoj: I'll leave it as is.

dstep/translator/MacroIndex.d
ciechowoj: Fixed.

dstep/translator/Preprocessor.d
ciechowoj: Fixed.

dstep/translator/Preprocessor.d
ciechowoj: Fixed.

jacob-carlborg:
There's a lot of new parsing for preprocessor directives, but that's not for translating?

jacob-carlborg:
Update the changelog as well, if I've not already mentioned that 😃.

ciechowoj:
> There's a lot of new parsing for preprocessor directives, but that's not for translating?

To detect the header comment you need to find `#pragma once` or `#ifndef #define #endif`. I decided to add a bit more extended framework for this purpose and integrate it a bit with existing preprocessor-parsing functionality. Maybe I'll make an use of this at some point...

ciechowoj:
> Update the changelog as well, if I've not already mentioned that :smiley:.

Done.

jacob-carlborg:
👌

PR #77, opened 2016-07-23
Clang c

ciechowoj:
Adds libclang as integration test.

I eyeballed translated code (of libclang) searching if its correct and found that function pointers aren't translated correctly, so this PR contains a fix for this (the most recent commit).

I've tried to replace existing bindings with the generated ones and dstep seems to compile and run fine.


ciechowoj:
Rebased to master.

dstep/translator/CommentIndex.d
```diff
@@ -68,6 +68,73 @@ class CommentIndex
...
+
+        private static bool isNormalized(string content)
+        {
+            import std.ascii : isWhite;
+
+            for (size_t i = 1; i < content.length; ++i)
+            {
+                if (content[i] == '\n' && content[i - 1].isWhite)
+                    return false;
+            }
```
jacob-carlborg: Looks like this can be replaced with an algorithm, like `canFind`.

ciechowoj: I do not see how. There is three versions of `canFind`. The versions with `needle` cannot be used as there is no fixed sub-string to search for (rather a sequence of white-space+new-line). And the predicate considers only one element at once and I need to consider two subsequent elements.

ciechowoj: However I can always replace it with foreach.

jacob-carlborg: Not sure if it's better but you can:

1. Slice off the first character
2. Use `enumerate` to include the index of the current iteration
3. Use `canFind` with the same predicate as you have now

dstep/translator/CommentIndex.d
```diff
@@ -68,6 +68,73 @@ class CommentIndex
...
+            return result.data;
+        }
+
+        unittest
+        {
+            assert(normalize("") == "");
+            assert(normalize("foo") == "foo");
+            assert(normalize("foo \n") == "foo\n");
+            assert(normalize("foo \n  ") == "foo\n");
+            assert(normalize("foo \n  bar \n  ") == "foo\n  bar\n");
```
jacob-carlborg: Not sure I understand this function. Why is the leading whitespace before `bar` kept?

ciechowoj: It trims extra whitespaces at the end of line. That is after significant source code and before the linefeed character. Whitespace before `bar` are kept because it is `bar`'s indentation.

I used this function to trim superfluous whitespaces when translating comments content.

ciechowoj: Example:

source:
```
/*·This·is·multiline·comment·······
with·some·extra·whitespaces·····
at·the·end·of·each·of·lines·····
*/
```

translated
```
/*·This·is·multiline·comment
with·some·extra·whitespaces
at·the·end·of·each·of·lines
*/
```

ciechowoj: You don't want to trim spaces before `with`...

dstep/translator/CommentIndex.d
ciechowoj: I fixed it with `iota`.

dstep/translator/CommentIndex.d
jacob-carlborg: > It trims extra whitespaces at the end of line. That is after significant source code and before the linefeed character. 

So would this be the same as splitting on newline, stripping trailing whitespace and then joining with newline again? Something like `content.splitter("\n").map!(stripRight).join("\n")`.

> Whitespace before bar are kept because it is bar's indentation.

I see.

ciechowoj: > So would this be the same as splitting on newline, stripping trailing whitespace and then joining with newline again? Something like content.splitter("\n").map!(stripRight).join("\n").

Yes, indeed...

jacob-carlborg: So the exiting implementation can be replaced with the above?

ciechowoj: I've already done this.

jacob-carlborg: Ah, you updated the latest commit instead of the commit that actually introduced the code.

ciechowoj: Yeah...

resources/VERSION
```diff
@@ -0,0 +1 @@
+v0.2.2-59-ga88853a
```
jacob-carlborg: This file should not be included. Is the PR not rebased?

unit_tests/Common.d
```diff
@@ -50,6 +50,220 @@ bool compareString(string a, string b, bool strict)
...
+        import std.conv : to;
+
+        auto result = Appender!string();
+
+        auto l = line - q;
+
+        result.put(prefix);
+        result.put("\n");
+
+        for (size_t i = 0; i < q; ++i)
```
jacob-carlborg: Prefer to use `foreach` range loop where possible.

unit_tests/Common.d
```diff
@@ -50,6 +50,220 @@ bool compareString(string a, string b, bool strict)
...
+    {
+        import std.format;
+
+        auto actual = mismatchRegion(a, b, margin, strict);
+
+        if (expected != actual)
+        {
+            auto templ = "\nExpected:\n%s\nActual:\n%s\n";
+
+            string message = templ.format(expected, actual);
```
jacob-carlborg: Don't call `format` using UFCS.

dstep/translator/IncludeHandler.d
```diff
@@ -187,18 +188,6 @@ private:
...
-        import std.array : empty;
-        import std.algorithm.sorting : sort;
-
-        foreach (entry; imports.sort.filter!(e => !e.empty))
-            output.singleLine(entry);
-
-        if (!output.empty)
-            output.separator();
-    }
-
```
jacob-carlborg: I really don't like that you have one commit in a PR which added code and then another commit which restores that modifies/restores that code.

test_files/clang-c/Index.h
```diff
@@ -5206,7 +5206,7 @@ CINDEX_LINKAGE CXResult clang_findIncludesInFile(CXTranslationUnit TU,
                                               CXCursorAndRangeVisitor visitor);
 
 #ifdef __has_feature
-#  if __has_feature(blocks)
+#  if 0 && __has_feature(blocks)
```
jacob-carlborg: I figured out a way to call Clang blocks without language support using a library solution [1], if we would like to translate Clang blocks.

[1] https://github.com/jacob-carlborg/druntime/blob/550edd0a64f0eb2c4f35d3ec3d88e26b40ac779e/src/core/stdc/clang_block.d

jacob-carlborg:

resources/VERSION
ciechowoj: It is temporarily added in some of intermediate commits, but it isnt present at the end. Please do not tell me to rebase intermediate commits, I had enough fun with that yesterday...

jacob-carlborg: I read the commits individually, because I thought they were independent.

In the future you could try "fixup" commits and see if that helps. That will create a new commit with the intention to modify an existing commit. When a review is done you can do an interactive rebase with the `--autosquash` flag which will automatically set the correct commits to "squash". See this [1] for more information. Using fixup commits might actually help as well on GitHub because I think that it's more likely that the comments will persist.

[1] https://robots.thoughtbot.com/autosquashing-git-commits

jacob-carlborg:
Looking good, waiting for the changelog update.

jacob-carlborg:
BTW, does DStep translate enough that we can use automatically translated bindings for Clang?

ciechowoj:
Yes, this should be possible.

resources/VERSION
ciechowoj: > I read the commits individually, because I thought they were independent.

They were, a few rebases ago...

> In the future you could try "fixup" commits and see if that helps.

I'm using it already.

> When a review is done you can do an interactive rebase with the --autosquash flag which will automatically set the correct commits to "squash".

I didn't knew about this one, I'll surely try this.

dstep/translator/IncludeHandler.d
ciechowoj: Seems fixed.

ciechowoj:
> Looking good, waiting for the changelog update.

Done.

test_files/clang-c/Index.h
ciechowoj: I'll take a look.

ciechowoj:
> Done.

Or not, Travis is failing...

test_files/clang-c/Index.h
jacob-carlborg: Just to be clear, it's not something you have to do now, or at all.

ciechowoj: Sure.

jacob-carlborg:
Looking good. Anything else or is it read to be merged?

ciechowoj:
Ready.

jacob-carlborg:
👌

ciechowoj:
Thanks.

PR #78, opened 2016-07-26
Remove mambo and dstack dependencies

Dicebot:
Makes dstep compilable with only Phobos and libclang.

NB: I haven't tried to preserve special handling of whitespace only
strings implemented in `isBlank` as it is never actually needed for
dstep to work. It was replaced with plan length check.


dstep/translator/IncludeHandler.d
```diff
@@ -102,21 +100,27 @@ class IncludeHandler
...
-        auto imps = imports.map!(e => toImport(e));
+        auto r_imports = imports.map!(e => toImport(e));
 
-        foreach (entry; r.append(imps).filter!(e => e.any).unique)
-            output.singleLine(entry);
+        chain(r_includes, r_imports)
+            .filter!(e => e.length)
+            .array
+            .sort
+            .group
```
Dicebot: This is probably most non-trivial change in this PR - because `unique` can't be implemented for generic input range without extra allocations, I defined it via `.array.sort.group` instead.

ciechowoj: `chain(r_includes, r_imports)`
Camel case, I suppose : ) .

dstep/translator/Output.d
```diff
@@ -1062,14 +1060,13 @@ class ClassData : StructData
     string superclass;
     string[] interfaces;
 
-    HashSet!(string) propertyList;
+    bool[string] propertyList;
```
ciechowoj: In `clang/Util.d:82` there is defined a type `Set` which is already used in some other places.

jacob-carlborg: It even has the same API, make the other changes related to `propertyList` unnecessary 😃,

test.d
```diff
@@ -187,24 +184,26 @@ private:
...
     }
 
     void download (const ref Clang clang)
     {
         auto url = clang.baseUrl ~ clang.filename;
         auto dest = archivePath(clang.filename);
 
-        if (!Path.exists(dest))
-            Http.download(url, dest);
+        import std.net.curl;
```
jacob-carlborg: I would rather not bring in the dependency of curl. That's one reason I use Tango.

Dicebot: Would you be ok with using vibe.http.client?

Dicebot: Alternatively, I could use Adam's simple http module which depends only on std.socket

jacob-carlborg: > Would you be ok with using vibe.http.client

Wasn't the point of this PR to minimze the dependencies 😉.

> Alternatively, I could use Adam's simple http module which depends only on std.socket

Sounds better. Do you have a link?

Dicebot: https://github.com/adamdruppe/arsd/blob/master/http.d (can be simply copied in dstep repo itself) I haven't used it before though so I can't say how mature it is before trying out.

jacob-carlborg: IIRC, Adam (or someone) showed a really simple code snippet in the forums how to download a file over HTTP. I think it was one of those discussions about the dependency on libcurl. I mentioned I use Tango, he showed that snippet. I'm trying to find it...

jacob-carlborg: I haven't been able to find that snippet. I guess Adam's HTTP module is fine to add.

ciechowoj:
@Dicebot , @jacob-carlborg Would there be a possibility to speed things up a little and to merge this PR as a PR that removes Tango and Mambo from the dstep core and then add another PR that does it for `test.d`. The GSOC deadline starts to lurk at the horizon and I have some things to do that seems to depend on this PR.

ciechowoj:
Thanks.

dstep/translator/Output.d
```diff
@@ -1080,7 +1075,7 @@ class ClassData : StructData
         if (!(mangledName in mangledMethods))
         {
             mangledMethods.add(mangledName);
-            name = name.isBlank ? selector : name;
+            name = !name.length ? selector : name;
```
jacob-carlborg: You can use `empty` to avoid the negation.

jacob-carlborg:
Why is the dependencies not removed from `dub.json`. Are they used in `test.d` as well? You could replace the dependencies with a direct dependency on Tango. To make sure no usages of the dependencies are left.

jacob-carlborg:
> I don't think it is worth the trouble because there will be another (last) PR anyway, to update test.d and remove all deps. Purpose of merging this PR is simply to make things easier for ciechowoj

Fair enough.

dstep/translator/Output.d
Dicebot: That would require adding `import std.range` into overload set - in my opinion doing that just to check array `empty` is less readable than simple double negation. Are you sure?

jacob-carlborg: You don't consistently check for emptiness. Sometimes explicitly checking for `length == 0` sometimes not.
You could selectively import `empty`.

Dicebot: As you prefer. Updated.

jacob-carlborg: You only made the change in one place.

Dicebot: Updated in more places.

ciechowoj:
You've missed a mambo import in `IncludeHandler.d`.

jacob-carlborg:
Looking good, please squash commits.

jacob-carlborg:
👌

ciechowoj:
Good work : )

ciechowoj:
@Dicebot 
Do you have any progress with replacing the rest of the dependencies?

ciechowoj:
Ok.

PR #81, opened 2016-07-30
Fix #59 and #28.

### LINKED ISSUE #28: Crashes if fed nonexistent header
At least with the prebuilt dstep 0.1.0-2 on Arch, it crashes and spits out a backtrace if I feed it a non-existent C header, such as if I add a typo in the header's path.  It'd be nice if it issued an error instead, so the user knows what he did wrong.

### LINKED ISSUE #59: Shouldn't dstep exit with status code when there is some kind of error?

ciechowoj:
Fix #59: Shouldn't dstep exit with status code when there is some kind of error?
Fix #28: Crashes if fed nonexistent header.

### DIFF
--- a/dstep/driver/Application.d
+++ b/dstep/driver/Application.d
@@ -42,14 +42,24 @@ class Application
         import std.exception : enforce;
 
         enforce!DStepException(config.inputFiles.length > 0,
-            "Must supply at least one input file");
+            "dstep: error: must supply at least one input file\n");
+
+        enforceInputFilesExist(config);
 
         // when only one input file is supplied, -o argument is
         // interpreted as file path, otherwise as base directory path
         bool singleFileInput = config.inputFiles.length == 1;
 
+        Task!(
+            startParsingFile,
+            Configuration,
+            string,
+            string)*[] conversionTasks;
+
+        conversionTasks.length = config.inputFiles.length;
+
         // parallel generation of D modules for each of input files
-        foreach (fileName; config.inputFiles)
+        foreach (size_t index, fileName; config.inputFiles)
         {
             string outputFilename;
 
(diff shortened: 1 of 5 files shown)


dstep/driver/Application.d
```diff
@@ -76,13 +86,35 @@ class Application
...
+
+    static void enforceInputFilesExist(const Configuration config)
+    {
+        import std.exception : enforce;
+
+        foreach (inputFile; config.inputFiles)
+        {
+            enforce!DStepException(
+                exists(inputFile),
+                "dstep: error: file '" ~ inputFile ~ "' doesn't exist\n");
```
jacob-carlborg: Perhaps using `format` is cleaner.

unit_tests/IntegrationTests.d
```diff
@@ -43,3 +46,30 @@ unittest
...
+}
+
+// DStep should exit with non-zero status when there is a syntax error in the input file.
+unittest
+{
+    auto result = executeShell("dstep test_files/syntax_error.h");
+
+    assert(result.status != 0);
+    assert(result.output.find("syntax_error.h"));
+}
```
jacob-carlborg: Is there a unit test to verify that it exits with status code 0 when it completes successfully?

dstep/driver/Application.d
```diff
@@ -76,13 +86,35 @@ class Application
...
+                config,
+                fileName,
+                outputFilename);
 
-            conversionTask.executeInNewThread();
+            conversionTasks[index].executeInNewThread();
+        }
+
+        foreach (conversionTask; conversionTasks)
+            conversionTask.yieldForce();
```
jacob-carlborg: Is this used to force a task to complete?

jacob-carlborg:

dstep/driver/Application.d
ciechowoj: It is used to rethrow the exceptions originated in the thread of that task.

unit_tests/IntegrationTests.d
ciechowoj: Now, there is.

ciechowoj:
> Please add an entry to the changelog as well.

Done.

ciechowoj:
Is there anything more I can do here?

jacob-carlborg:
👌

ciechowoj:
Thanks.

PR #84, opened 2016-08-02
Fix #83 and #30.

### LINKED ISSUE #30: Single space inserted after function names
It gets inserted between a function name and the opening parenthesis for its parameters and even for function pointer declarations, please remove it.

### LINKED ISSUE #83: New multiline translation
I noticed that dstep now outputs functions multi-line. I do some post processing of the dstep outputs and the new multi-line outputs would make it difficult for me to post-process the the functions outputs in the way I usually do.

ciechowoj:
Fix #83: New multiline translation.
Fix #30: Single space inserted after function names.

I know that there was a discussion (#30) about adding formatting features to dstep with some negative arguments, but it doesn't cost much to add them, they can save some time for some people and apparently there is a need for them from some users (at least two at the moment).

### DIFF
--- a/dstep/main.d
+++ b/dstep/main.d
@@ -64,7 +64,9 @@ auto parseCLI (string[] args)
         "public-submodules", "Use public imports for submodules.", &config.publicSubmodules,
         "package", "Specify package name.", &config.packageName,
         "dont-reduce-aliases", "Disable reduction of primitive type aliases.", &config.dontReduceAliases,
-        "no-portable-wchar_t", "Translate wchar_t to wchar or dchar depending on its size.", &config.noPortableWCharT);
+        "no-portable-wchar_t", "Translate wchar_t to wchar or dchar depending on its size.", &config.noPortableWCharT,
+        "single-line-function-headers", "Do not break function headers to multiple lines.", &config.singleLineFunctionHeaders,
+        "no-space-after-function-name", "Do not put a space after a function name.", &config.noSpaceAfterFunctionName);
 
     // remove dstep binary name (args[0])
     args = args[1 .. $];
--- a/changelog.md
+++ b/changelog.md
@@ -21,10 +21,12 @@ Issue #2: Self alias should be removed bug.
 Issue #10: Embedded struct not generated.
 Issue #21: wchar_t should be tranlsated to core.stdc.stddef.wchar_t.
 Issue #29: Don't name anonymous enums.
+Issue #30: Single space inserted after function names.
 Issue #39: Recognize and translate __attribute__((__packed__)).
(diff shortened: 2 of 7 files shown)


dstep/main.d
```diff
@@ -63,7 +63,9 @@ auto parseCLI (string[] args)
         "no-comments", "Disable translation of comments.", &config.noComments,
         "public-submodules", "Use public imports for submodules.", &config.publicSubmodules,
         "package", "Specify package name.", &config.packageName,
-        "dont-reduce-aliases", "Disable reduction of primitive type aliases.", &config.dontReduceAliases);
+        "dont-reduce-aliases", "Disable reduction of primitive type aliases.", &config.dontReduceAliases,
+        "single-line-function-headers", "Do not break function headers to multiple lines.", &config.singleLineFunctionHeaders,
+        "no-space-after-function-name", "Do not put a space after a function name.", &config.noSpaceAfterFunctionName);
```
jacob-carlborg: Would it be possible to have both a positive and negative argument, like `--space-after-function-name` and `--no-space-after-function-name`. This is quite common for Posix commands, see `git add --help` where there is a `--[no-]ignore-removal` command, for an example.

jacob-carlborg:

dstep/main.d
ciechowoj: It cannot be done automatically with getopt. The only way I see is to add a negation of given option to `Configuration` and check the constraints in code. E.g.

```
 "no-space-after-function-name", "Do not put a space after a function name.", &config.noSpaceAfterFunctionName,
 "space-after-function-name", "Put a space after a function name.", &config.spaceAfterFunctionName);

/+ ... +/

if (config.noSpaceAfterFunctionName && config.spaceAfterFunctionName) {
    // very bad
}
```

ciechowoj:
> Please add an entry to the changelog as well.

Done.

dstep/main.d
jacob-carlborg: Then we'll leave it as is.

jacob-carlborg:
👌

ciechowoj:
Thanks.

dstep/main.d
ciechowoj: I rethought the matter and `getopt` seems to support assigning of a `true` and `false` to the CLI switch. Maybe it would be better to resign from `no-` prefixed switches and require users to specify the desirable value explicitly instead. E.g.

`--reduce-aliases=false`
`--comments=true`

I just want to point out that possibility.

jacob-carlborg: Yeah, I think that looks better than all the `dont...` flags.

PR #86, opened 2016-08-06
[WIP] Windows

ciechowoj:
Build and run dstep on MS Windows.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -21,13 +21,15 @@ The source code is available under the [Boost Software License 1.0](http://www.b
 
 ## Building
 
-### Requirements
+### Linux
+
+#### Requirements
 
 * libclang - [http://clang.llvm.org](http://clang.llvm.org) - 3.4 or later that is binary compatible with 3.4
 * DMD - [http://dlang.org/download.html](http://dlang.org/download.html) - 2.069.2
 * Dub [http://code.dlang.org/download](http://code.dlang.org/download)
 
-### Building
+#### Building
 
 1. Install all requirements, see [above](#requirements)
 2. Clone the repository by running:
@@ -35,6 +37,28 @@ The source code is available under the [Boost Software License 1.0](http://www.b
         $ git clone git://github.com/jacob-carlborg/dstep.git
 
 3. run `dub build`
+
+### Windows
+
+#### Requirements
+
+* LLVM 64-bit - [http://llvm.org/releases/download.html](http://llvm.org/releases/download.html) - pre-built 64-bit binaries for Windows
+* DMD - [http://dlang.org/download.html](http://dlang.org/download.html) - 2.069.2
+* Dub - [http://code.dlang.org/download](http://code.dlang.org/download)
+* Visual Studio - for example Visual Studio Community
+
+#### Building
+
+1. Install all requirements, see [above](#requirements)
+2. Clone the repository by running:
+
+		$ git clone git://github.com/jacob-carlborg/dstep.git
+
+3. run `dub build --arch=x86_64 --build=release`
+
+#### Remarks
+
+Currently only 64-bit release version is supported. Remeber to install 64-bit version of LLVM to the default destination (`Program Files`) and to add its binaries to PATH enviromental variable.
 
 ## Usage
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -32,4 +32,4 @@ build_script:
   - dub build --arch=x86_64
 
 test_script:
-  - dub test --arch=x86_64
+  - dub --config=test --arch=x86_64
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -1,0 +1,35 @@
+
+platform: x64
+
+environment:
+  matrix:
+  - DC: dmd
+    DVersion: 2.071.0
+    arch: x64
+
+skip_tags: true
+
+branches:
+  only:
+    - windows
+
+install:
+  - ps: echo "Downloading dmd...";
+        Invoke-WebRequest "http://downloads.dlang.org/releases/2.x/$env:DVersion/dmd.$env:DVersion.windows.7z" -OutFile c:\dmd.7z;
+        7z x c:\dmd.7z -oc:\ > $null;
+        echo "Downloading dub...";
+        Invoke-WebRequest https://code.dlang.org/files/dub-1.0.0-beta.1-windows-x86.zip -OutFile c:\dub.zip;
+        7z x c:\dub.zip -oc:\dub > $null;
+
+  - set PATH=c:\dub;%PATH%
+  - dub --version
+
+before_build:
+  - set PATH=c:\dmd2\windows\bin;%PATH%
+  - '"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall" amd64'
+
+build_script:
+  - dub build --arch=x86_64
+
+test_script:
+  - dub test --arch=x86_64
--- a/clang/Util.d
+++ b/clang/Util.d
@@ -100,9 +100,48 @@ Set!T SetFromList(T)(T[] list)
     return result;
 }
 
-extern (C) int mkstemps(char*, int);
-extern (C) char* mkdtemp(char*);
-extern (C) int close(int);
+version (Posix)
+{
+    private extern (C) int mkstemps(char*, int);
+    private extern (C) char* mkdtemp(char*);
+    private extern (C) int close(int);
+}
+else
+{
+    struct GUID {
+        uint Data1;
+        ushort Data2;
+        ushort Data3;
+        ubyte[8] Data4;
+    }
+
+    private extern (Windows) uint CoCreateGuid(GUID* pguid);
+
+    private string createGUID()
+    {
+        char toHex(uint x)
+        {
+            if (x < 10)
+                return cast(char) ('0' + x);
+            else
+                return cast(char) ('A' + x - 10);
+        }
+
+        GUID guid;
+        CoCreateGuid(&guid);
+
+        ubyte* data = cast(ubyte*)&guid;
+        char[32] result;
+
+        foreach (i; 0 .. 16)
+        {
+            result[i * 2 + 0] = toHex(data[i] & 0x0fu);
+            result[i * 2 + 1] = toHex(data[i] >> 16);
+        }
+
+        return result.idup;
+    }
+}
 
 class NamedTempFileException : object.Exception
 {
@@ -143,24 +182,33 @@ File namedTempFile(string prefix, string suffix)
     import std.path;
     import std.format;
 
-    string name = format("%sXXXXXXXXXXXXXXXX%s\0", prefix, suffix);
-    char[] path = buildPath(tempDir(), name).dup;
-    const size_t termAnd6XSize = 7;
+    version (Posix)
+    {
+        string name = format("%sXXXXXXXXXXXXXXXX%s\0", prefix, suffix);
+        char[] path = buildPath(tempDir(), name).dup;
+        const size_t termAnd6XSize = 7;
 
-    immutable size_t begin = path.length - name.length + prefix.length;
-    immutable size_t end = path.length - suffix.length - termAnd6XSize;
+        immutable size_t begin = path.length - name.length + prefix.length;
+        immutable size_t end = path.length - suffix.length - termAnd6XSize;
 
-    randstr(path[begin .. end]);
+        randstr(path[begin .. end]);
 
-    int fd = mkstemps(path.ptr, cast(int) suffix.length);
-    scope (exit) close(fd);
+        int fd = mkstemps(path.ptr, cast(int) suffix.length);
+        scope (exit) close(fd);
 
-    path = path[0..$-1];
+        path = path[0 .. $ - 1];
 
-    if (fd == -1)
-        throw new NamedTempFileException(path.idup);
+        if (fd == -1)
+            throw new NamedTempFileException(path.idup);
 
-    return File(path, "wb+");
+        return File(path, "wb+");
+    }
+    else
+    {
+        string name = format("%s%s%s", prefix, createGUID(), suffix);
+        string path = buildPath(tempDir(), name);
+        return File(path, "wb+");
+    }
 }
 
 string namedTempDir(string prefix)
@@ -169,22 +217,37 @@ string namedTempDir(string prefix)
     import std.path;
     import std.format;
 
-    string name = format("%sXXXXXXXXXXXXXXXX\0", prefix);
-    char[] path = buildPath(tempDir(), name).dup;
-    const size_t termAnd6XSize = 7;
+    version (Posix)
+    {
+        string name = format("%sXXXXXXXXXXXXXXXX\0", prefix);
+        char[] path = buildPath(tempDir(), name).dup;
+        const size_t termAnd6XSize = 7;
 
-    immutable size_t begin = path.length - name.length + prefix.length;
+        immutable size_t begin = path.length - name.length + prefix.length;
 
-    randstr(path[begin .. $ - termAnd6XSize]);
+        randstr(path[begin .. $ - termAnd6XSize]);
 
-    char* result = mkdtemp(path.ptr);
+        char* result = mkdtemp(path.ptr);
 
-    path = path[0..$-1];
+        path = path[0..$-1];
 
-    if (result == null)
-        throw new NamedTempDirException(path.idup);
+        if (result == null)
+            throw new NamedTempDirException(path.idup);
 
-    return path.idup;
+        return path.idup;
+    }
+    else
+    {
+        string name = prefix ~ createGUID();
+        string path = buildPath(tempDir(), name);
+
+        try
+            mkdirRecurse(path);
+        catch (FileException)
+            throw new NamedTempDirException(path);
+
+        return path;
+    }
 }
 
 string asAbsNormPath(string path)
--- a/clang/Util.d
+++ b/clang/Util.d
@@ -80,6 +80,11 @@ mixin template CX ()
     }
 }
 
+string clangVersion()
+{
+    return clang_getClangVersion().toD;
+}
+
 alias Set(T) = void[0][T];
 
 void add(T)(ref void[0][T] set, T value) {
--- a/dstep/translator/IncludeHandler.d
+++ b/dstep/translator/IncludeHandler.d
@@ -49,11 +49,13 @@ class IncludeHandler
             "_uint32_t" : "core.stdc.stdint",
             "_uint64_t" : "core.stdc.stdint",
             "stdio" : "core.stdc.stdio",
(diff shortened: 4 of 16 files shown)


ciechowoj:
@jacob-carlborg 

Could you take a look on this PR, when you have some time.

It makes all changes necessary to run dstep and its tests on Windows and adds a config script for AppVeyor. The tests are passing on AppVeyor (I attached AppVeyor to my fork of dstep), however only for the version of libclang supplied with AppVeyor.

As for, now it cannot be merged, because the 'full' testing script (`test.d`) doesn't work, due to dependencies.

unit_tests/IntegrationTests.d
```diff
@@ -59,7 +59,7 @@ unittest
 // DStep should exit with non-zero status when an input file doesn't exist.
 unittest
 {
-    auto result = executeShell("./bin/dstep test_files/nonexistent.h");
+    auto result = executeShell("\"./bin/dstep\" test_files/nonexistent.h");
```
jacob-carlborg: I prefer to use a string literal with backticks to avoid having to escape `"`.

jacob-carlborg:
I assume Windows accepts `/` for path separator?

jacob-carlborg:
> As for, now it cannot be merged, because the 'full' testing script (test.d) doesn't work, due to dependencies.

Due to Tango or?

ciechowoj:
To build it I removed `dstack` from `dub.json` dependencies. So that `test` configuration (the one that runs dstep unit tests for different versions of libclang) cannot be compiled, as it still depends on `dstack` and Tango. The configurations `default` and `unittest` work.

ciechowoj:
> I assume Windows accepts / for path separator?
Yes.

unit_tests/IntegrationTests.d
ciechowoj: Fixed.

jacob-carlborg:
> To build it I removed dstack from dub.json dependencies

But where is the problem, dstack, mambo or Tango? Tango should work because I'm using that in another project which supports Windows. I'm thinking, can you remove dstack and mambo but keep Tango as a dependency? That should hopefully minimize the necessary changes.

ciechowoj:
I added back the Tango dependency and it doesn't compile. Here is the the output of `dub`:

[http://pastebin.com/PFgDC5Rm](http://pastebin.com/PFgDC5Rm)

I don't really see why it fails. Are `Deprecation` warnings fatal? I do not see any errors.

ciechowoj:
Hmm, when I try to compile it for x64 there are the errors.

[http://pastebin.com/5wDwqCif](http://pastebin.com/5wDwqCif)

ciechowoj:
With dicebot changes it passes on linux and macos now, however I have a problem with Windows.
Is there a way to extract something like `http://llvm.org/releases/3.5.0/", "LLVM-3.5.0-win32.exe` from console ?

ciechowoj:
If not one way would be to build the LLVM in CLI from source, however it takes some amount of time...

ciechowoj:
Ok, it seems like 7zip can handle it.

README.markdown
```diff
@@ -36,6 +38,28 @@ The source code is available under the [Boost Software License 1.0](http://www.b
...
+1. Install all requirements, see [above](#requirements)
+2. Clone the repository by running:
+
+		$ git clone git://github.com/jacob-carlborg/dstep.git
+
+3. run `dub build --arch=x86_64 --build=release`
+
+#### Remarks
+
+Currently only 64-bit release version is supported. Remeber to install 64-bit version of LLVM to the default destination (`Program Files`) and to add its binaries to PATH enviromental variable.
```
jacob-carlborg: Not sure how LLVM is packaged on Windows but on Posix it's enough for the library to be placed in standard library search path or the DStep directory.

appveyor.yml
```diff
@@ -0,0 +1,66 @@
+
+platform: x64
+
+environment:
+ matrix:
+  - DC: dmd
```
jacob-carlborg: Looks like only one space of indentation for these two lines.

README.markdown
ciechowoj: The standard library search patch on Windows is PATH. It could be placed in DStep directory of course, but it is additional burden for an user. In the readme I'm considering a scenario in which the user uses standard LLVM installer, and that installer asks explicitly if it should add the binaries to the PATH.

clang/Util.d
```diff
@@ -100,9 +100,48 @@ Set!T SetFromList(T)(T[] list)
...
+    struct GUID {
+        uint Data1;
+        ushort Data2;
+        ushort Data3;
+        ubyte[8] Data4;
+    }
+
+    extern (Windows) uint CoCreateGuid(GUID* pguid);
+
+    string createGUID()
```
jacob-carlborg: Please make all these utility functions private.

dstep/translator/IncludeHandler.d
```diff
@@ -49,11 +49,13 @@ class IncludeHandler
             "_uint32_t" : "core.stdc.stdint",
             "_uint64_t" : "core.stdc.stdint",
             "stdio" : "core.stdc.stdio",
+            "corecrt_wstdio" : "core.stdc.stdio",
             "stdlib" : "core.stdc.stdlib",
             "string" : "core.stdc.string",
             "tgmath" : "core.stdc.tgmath",
             "time" : "core.stdc.time",
             "_time_t" : "core.stdc.time",
+            "corecrt" : "core.stdc.time",
```
jacob-carlborg: Are these two addition covered by tests?

ciechowoj: Rather forced : )

README.markdown
jacob-carlborg: Ok, I see.

appveyor.yml
```diff
@@ -0,0 +1,66 @@
...
+  - ps : if($env:DC -eq "dmd"){
+           $env:PATH += ";C:\dmd2\windows\bin;";
+         }
+         elseif($env:DC -eq "ldc"){
+           $version = $env:DVersion;
+           $env:PATH += ";C:\ldc2-$($version)-win64-msvc\bin";
+           $env:DC = "ldc2";
+         }
+  - ps: $env:compilersetup = "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall";
+  - '"%compilersetup%" %compilersetupargs%'
```
jacob-carlborg: Could you please move these powershell scripts to their own files. Too much code as YAML.

README.markdown
ciechowoj: Here you have details, if curious: [https://msdn.microsoft.com/en-us/library/7d83bc18.aspx](https://msdn.microsoft.com/en-us/library/7d83bc18.aspx)

appveyor.yml
ciechowoj: Fixed.

appveyor.yml
ciechowoj: It seems like it is common practice to pack all that scripts inside `appveyor.yml`. I took that script from [https://wiki.dlang.org/Continuous_Integration](https://wiki.dlang.org/Continuous_Integration). Anyway, I distilled the scripts to bare minimum. Is it good enough or should I move the rest to the separate file.

The problem with separate files is the scripts sets and references environmental variables heavily, and I'm not sure how it would work on the subscript boundary.

clang/Util.d
ciechowoj: Done.

clang/Util.d
```diff
@@ -80,6 +80,11 @@ mixin template CX ()
     }
 }
 
+string clangVersion()
+{
+    return clang_getClangVersion().toD;
```
jacob-carlborg: Is this really working? Last time I tried it, it didn't return different values for different versions I tried. Or I made some mistake.

test.d
```diff
@@ -59,7 +65,11 @@ struct TestRunner
     void activate (const Clang clang)
     {
         auto src = buildNormalizedPath(workingDirectory, clang.versionedLibclang);
-        auto dest = buildNormalizedPath(workingDirectory, clang.libclang);
+
+        version (Posix)
+            auto dest = buildNormalizedPath(workingDirectory, clang.libclang);
+        else
+            auto dest = buildNormalizedPath(workingDirectory, clang.libclang);
```
jacob-carlborg: There's no difference between these two lines and the one that you removed.

clang/Util.d
ciechowoj: I've just been trying this and it doesn't work (it behaves like you said). I'll provide more explanation later.

jacob-carlborg:
It says that only Windows 64bit is supported. I assume that's because of the default COFF format. But what about 32bit COFF using the `-m32mscoff` flag?

ciechowoj:
> It says that only Windows 64bit is supported. I assume that's because of the default COFF format. But what about 32bit COFF using the -m32mscoff flag?

I made it work too, however I haven't figured out yes, how to put it together in a way it is simple to use for end user.

test.d
ciechowoj: fixed.

appveyor.yml
jacob-carlborg: You mean that setting `PATH` in the install section persists to the build section?

ciechowoj:
> Is this really working? Last time I tried it, it didn't return different values for different versions I tried. Or I made some mistake.

It seems like the library 'libclang.lib` file and corresponding `libclang.dll` are somehow linked together and the dll cannot be easily exchanges, without recompiling the dstep against another `.lib`.

I arranged the things on my windows machine in a way that only single `libclang.dll` is visible and manually replaced it with different version. An attempt to run `dstep` in such a environment end with some error "The application was unable to start correctly (0xc000007b), yada yada...". 

Most likely you were receiving the same versions because, only single `dll` were used.

ciechowoj:
So `test.d` currently works with the exception that it doesn't switch the dlls. 
I see two solutions: 

1. Make `test.d` recompile dstep for every clang version.
2. Make libclang bindings dynamic (like in derelict, then switching dlls should work).

What do you think about merging it as is for now. The tests on windows would be run against a LLVM version from appveyor (3.8.0). It would be great, if you can integrate this repo with AppVeyor, so that test for all three platforms are run.

appveyor.yml
ciechowoj: Yes, for example.

clang/Util.d
ciechowoj: > Is this really working? Last time I tried it, it didn't return different values for different versions I tried. Or I made some mistake.

> I've just been trying this and it doesn't work (it behaves like you said). I'll provide more explanation later.

My comment refers to Windows, on Linux it works as it should.

ciechowoj:
> Is this really working? Last time I tried it, it didn't return different values for different versions I tried. Or I made some mistake.

Ok, it seems to work on Windows too. I've just used 32-bit dll with 64-bit exe...

jacob-carlborg:
>  made it work too, however I haven't figured out yes, how to put it together in a way it is simple to use for end user.

If DStep is built as 32bit, then it will just libclang 32bit, if DStep is built as 64bit, it will use libclang 64bit. Seems simple enough?

ciechowoj:
Sure. You are right, but how to configure dub to make it work.

ciechowoj:
I'm saying, I just haven't spent enough time on 32-bit to make it work smoothly.

jacob-carlborg:
Something like `"dflags-x86-windows": ["-m32mscoff"]`. I don't remember the exact syntax, it's in the package format documentation for Dub.

jacob-carlborg:
> I'm saying, I just haven't spent enough time on 32-bit to make it work smoothly.

Fair enough.

ciechowoj:
> Something like "dflags-x86-windows": ["-m32mscoff"]. I don't remember the exact syntax, it's in the package format documentation for Dub.

But, if I'm not wrong it will try to use OPTLINK instead of ms linker (they are both named `link.exe` ; P).

jacob-carlborg:
> But, if I'm not wrong it will try to use OPTLINK instead of ms linker (they are both named link.exe ; P).

Yeah, I think this was mentioned in the forums, or somewhere. Don't remember the conclusion. You could just remove Optlink 😃.

ciechowoj:
Sure, something like:
```
"preBuildCommands": ["rm", "-f", "C:\D\dmd2\windows\bin\link.exe"]
```
Should be enough...

jacob-carlborg:
Found this: https://github.com/dlang/installer/commit/83b3ab428e1bc836554fffb7fd2784686f5624eb

jacob-carlborg:
> `"preBuildCommands": ["rm", "-f", "C:\D\dmd2\windows\bin\link.exe"]`

No no no. It has to only execute in the CI environment.

jacob-carlborg:
When a user is building DStep it's up to the user to make sure that the correct linker is found first in the PATH.

ciechowoj:
> Found this: dlang/installer@83b3ab4

Yes, this is a step in good direction.

ciechowoj:
> No no no. It has to only execute in the CI environment.

Just kidding : )

> When a user is building DStep it's up to the user to make sure that the correct linker is found first in the PATH.

> It says that only Windows 64bit is supported. I assume that's because of the default COFF format. But what about 32bit COFF using the -m32mscoff flag?

I assumed that the aim is to arrange it in the way, the user needs only to invoke `dub build` without any other actions and it should work (no changing path, etc.), if only all required tools are installed.

However, to enable 32-bit on CL shouldn't be that hard.

jacob-carlborg:
> Just kidding : )

I was not, not about doing it in the CI environment.

> I assumed that the aim is to arrange it in the way, the user needs only to invoke dub build without any other actions and it should work (no changing path, etc.), if only all required tools are installed.

Ideally that should be the case. I don't know enough about Windows, perhaps it already works? Otherwise we need to somehow detect the Microsoft linker and select that, somehow. Not sure if that's possible using some environment variable.

ciechowoj:
> I don't know enough about Windows, perhaps it already works?

No, it doesn't.

> Otherwise we need to somehow detect the Microsoft linker and select that, somehow. Not sure if that's possible using some environment variable.

At CI I'm doing something like `"C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall" amd64`, maybe it can be added to dub somehow.

jacob-carlborg:
> At CI I'm doing something like "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall" amd64, maybe it can be added to dub somehow.

I have no idea what that is doing. I was about to comment on that. If someone has a different version of Visual Studio installed, or if the CI environment is upgraded, hardcoding the path like that won't work. It it possible to get that path without hardcoding it? Also, if it's run on a 32bit Windows I think it will be `Program Files` instead of `Program Files (x86)`.

BTW, I think there's a Visual Studio shell. I'm wondering if A. that will set the correct linker, B. is expected that that shell is used by the user when building.

ciechowoj:
> I have no idea what that is doing.

I suppose it setups the environment. I based the appveyor script on `https://wiki.dlang.org/Continuous_Integration`. I checked some other projects, they are doing similar.

> If someone has a different version of Visual Studio installed, or if the CI environment is upgraded, hardcoding the path like that won't work. 

For CI it isn't big issue.

> It it possible to get that path without hardcoding it?

Don't know.

> Also, if it's run on a 32bit Windows I think it will be Program Files instead of Program Files (x86).

On my machine I have both linkers in `Program Files (x86)` (there is single VS installation).

jacob-carlborg:
As far as I know `Program Files (x86)` only exists on 32bit versions of Windows. It doesn't matter if the linkers are linking 32bit or 64bit applications. It's a matter if the linker itself is compiled as 32bit or 64bit. If it's compiled as 32bit it will be placed in `Program Files (x86)` on 64bit Windows. On 32bit Windows, since you can only run 32bit code, there's only `Program Files`.

jacob-carlborg:
I think you'll run into issues in the CI environment when trying to run 32bit, unless you cross-compile.

ciechowoj:
> As far as I know Program Files (x86) only exists on 32bit versions of Windows.

No, it exists on x64. In `Program Files (x86)` there are 32-bit apps, in `Program Files` there are 64-bit apps. On 32-bit windows there is only `Program Files`.

jacob-carlborg:
> No, it exists on x64. In Program Files (x86)

Yeah, I mistyped, I just edited my post 😃-

ciechowoj:
> On 32bit Windows, since you can only run 32bit code, there's only Program Files.

Should we worry about 32-bit windows, after all?

jacob-carlborg:
> Should we worry about 32-bit windows?

Not if it will take up too much time.

> We do not worry about 32bit linux, after all...

I do [1]. We cannot test that with Travis CI but I do that manually when it's time for a release. The last release did not have any 32bit binaries because I had some trouble with the virtual machine I used. My 32bit machine was one release older than the 64bit one and basically nothing works. I hope this can be fixed by using docker, assuming it's possible to cross-compile using docker.

[1] https://github.com/jacob-carlborg/dstep/releases/tag/v0.1.0

ciechowoj:
I see.

jacob-carlborg:
Please rebase, I've merged the PR from Dicebot which removes all dependencies.

jacob-carlborg:
I added DStep to AppVeyor https://ci.appveyor.com/project/jacob-carlborg/dstep, is there something else I need to do to make AppVeyor run the tests?

ciechowoj:
> I added DStep to AppVeyor https://ci.appveyor.com/project/jacob-carlborg/dstep, is there something else I need to do to make AppVeyor run the tests?

It should just work.

ciechowoj:
Rebased.

jacob-carlborg:
It doesn't look like this is being tested by AppVeyor, is that because AppVeyor was not enabled before this PR was created?

ciechowoj:
Yes, I would need to add a commit to trigger the tests...

18.08.2016 08:42 "jacob-carlborg" <notifications@github.com> napisał(a):

jacob-carlborg:
Can I trigger a build manually? I tried doing that but I can no options on what to build, i.e. branch or similar.

ciechowoj:
> Can I trigger a build manually? I tried doing that but I can no options on what to build, i.e. branch or similar.

Yes, there is an option to trigger the build but I haven't found yet the one to select the branch. I'll try to overwrite the most recent commit.

ciechowoj:
Changing commits doesn't work either.

jacob-carlborg:
How about making a new commit, we can remove it later.

ciechowoj:
> How about making a new commit, we can remove it later.

I tried this, doesn't work.

jacob-carlborg:
In the other PR you mentioned that the 64bit build is not work. What is the status?

ciechowoj:
Still doesnt work. Iam trying to fix it.

18.08.2016 16:08 "jacob-carlborg" <notifications@github.com> napisał(a):

PR #87, opened 2016-08-09
Fix #85: dstep not converting `const T x[]` to `const (T)* x`.

### LINKED ISSUE #85: dstep not converting `const T x[]` to `const (T)* x`
I am currently working with dstep version 0.2.2 and found that: `const T x[]` is not currently being converted to `const (T)* x`. It says `<unimplemented>`. I am currently doing a find and replace to `const T *x` in the original script which gives the desired output but it would be nice not to have to do that.

Thanks a lot for this very useful tool.

ciechowoj:
As in the title.


jacob-carlborg:
👌

PR #89, opened 2016-08-10
Fix #20: #define (simplest cases only?).

### LINKED ISSUE #20: #define (simplest cases only?)
I'm sorry if this has already been brought up but I haven't found it in issues or docs: can preprocessor directives like `#define` be handled? It looks like they can't. I'm not too familiar with libclang so this may well be something obvious.

Aside from dealing with compile time flags, most of the C libraries use `#define` to set constants, as in `#define X (1)`, `#define Y ((float)-X)`, which easily transforms to `immutable`. For a large percentage of headers, just handling constants would be sufficient to get them working.

However, C-defines can get ugly, weird and unpredictable and sure there's no way to handle all of them... though some can be parsed into immutable constants, some into functions `#define f(a, b) ((a) + (b))`.

Are there any plans regarding this?

Thanks.

ciechowoj:
Add unit test cases for the issue.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -21,6 +21,7 @@
 ### Bugs fixed
 Issue #2: Self alias should be removed bug.
 Issue #10: Embedded struct not generated.
+Issue #20: #define (simplest cases only?).
 Issue #21: wchar_t should be tranlsated to core.stdc.stddef.wchar_t.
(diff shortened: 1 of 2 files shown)


jacob-carlborg:
👌

PR #90, opened 2016-08-14
Last bits of external dependencies

Dicebot:
Removes external dependencies from test.d and from dub.json

### DIFF
--- a/test.d
+++ b/test.d
@@ -193,10 +193,11 @@ private:
         auto url = clang.baseUrl ~ clang.filename;
         auto dest = archivePath(clang.filename);
 
-        import std.net.curl;
+        import std.file : write;
+        import HttpClient : getBinary;
 
         if (!exists(dest))
-            std.net.curl.download(url, dest);
+            write(dest, getBinary(url));
     }
 
     void extractArchive (const ref Clang clang)
--- a/test.d
+++ b/test.d
@@ -1,11 +1,12 @@
 module test;
 
 import std.process;
-import std.file : rmdirRecurse;
-
-import Path = tango.io.Path;
-
-import mambo.core._;
+import std.stdio;
+import std.file;
+import std.path;
+import std.algorithm;
+import std.string;
+import std.exception;
 
 int main ()
 {
@@ -20,13 +21,13 @@ struct TestRunner
     {
         int result = 0;
         auto matrix = setup();
-        activate(matrix.clangs.first);
+        activate(matrix.clangs[0]);
         build();
 
         foreach (const clang ; matrix.clangs)
         {
             activate(clang);
-            println("Testing with libclang version ", clang.version_);
+            writeln("Testing with libclang version ", clang.version_);
             result += unitTest();
         }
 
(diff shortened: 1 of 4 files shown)


ciechowoj:
Thanks.

test.d
```diff
@@ -428,160 +427,345 @@ version (linux):
...
-{
-    import tango.io.device.File;
-    import tango.io.model.IConduit;
-    import tango.net.device.Socket;
-    import tango.net.http.HttpGet;
-    import tango.net.http.HttpConst;
+/*****************************************************************************/
 
-static:
+// HTTP utils based on https://github.com/adamdruppe/arsd/blob/master/http.d
```
jacob-carlborg: * Please put this in its own module
* I assume it handles redirects?

Dicebot: Which one? `test.d` is currently compiled to exclude all of `dstep` and `clang` packages. Shall I introduce new package or remove the limitation?

jacob-carlborg: I'm not sure where it's best to place it. Perhaps in the `unit_test` directory?

Dicebot: > I assume it handles redirects?

I don't think so though it shouldn't be too hard to add it by checking 3xx status code check here https://github.com/jacob-carlborg/dstep/pull/90/commits/875b5988a64c7e1a6a4c9338545bf06e0e3a470e#diff-0c0cbcae56297a371672c36f64de91f3R13

It is not needed right now though seeing that tests are green.

unit_tests/HttpClient.d
```diff
@@ -0,0 +1,326 @@
...
+		f.send(d);
+	};
+
+	char[4096] readBuffer; // rawRead actually blocks until it can fill up the whole buffer... which is broken as far as http goes so one char at a time i guess. slow lol
+	char[] delegate() read = () {
+		size_t num = f.receive(readBuffer);
+		return readBuffer[0..num];
+	};
+
+	version(with_openssl) {
```
jacob-carlborg: Hmm, so it doesn't support SSL out of the box? But I guess since the tests are passing we don't need that.

unit_tests/HttpClient.d
```diff
@@ -0,0 +1,326 @@
...
+
+import std.socket;
+import std.string;
+import std.conv;
+static import std.algorithm;
+static import std.uri;
+
+ubyte[] getBinary(string url, string[string] cookies = null) {
+	auto hr = httpRequest("GET", url, null, cookies);
+	if(hr.code != 200)
```
jacob-carlborg: Looks like it doesn't handle redirects, but I guess we can live without that until it's actually needed.

Dicebot: Yes, as I have commented above, it should be simply to add once needed - just check 3xx status code here and issue new `httpRequest` in a loop with url taken from `Location` header.

unit_tests/HttpClient.d
Dicebot: It would also require extra link-time dependency (openssl) if added right now.

jacob-carlborg: Good point.

jacob-carlborg:
👌

jacob-carlborg:
Thanks.

PR #91, opened 2016-08-17
Windows32

ciechowoj:
Add CI for 32-bit.

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -21,7 +21,7 @@ The source code is available under the [Boost Software License 1.0](http://www.b
 
 ## Building
 
-### Linux
+### Posix
 
 #### Requirements
 
@@ -36,13 +36,13 @@ The source code is available under the [Boost Software License 1.0](http://www.b
 
         $ git clone git://github.com/jacob-carlborg/dstep.git
 
-3. run `dub build`
+3. Run `dub build`
 
 ### Windows
 
 #### Requirements
 
-* LLVM 64-bit - [http://llvm.org/releases/download.html](http://llvm.org/releases/download.html) - pre-built 64-bit binaries for Windows
+* LLVM - [http://llvm.org/releases/download.html](http://llvm.org/releases/download.html) - pre-built binaries for Windows
 * DMD - [http://dlang.org/download.html](http://dlang.org/download.html) - 2.069.2
 * Dub - [http://code.dlang.org/download](http://code.dlang.org/download)
 * Visual Studio - for example Visual Studio Community
@@ -54,11 +54,12 @@ The source code is available under the [Boost Software License 1.0](http://www.b
 
 		$ git clone git://github.com/jacob-carlborg/dstep.git
 
-3. run `dub build --arch=x86_64 --build=release`
+3. Run `dub build --build=release` to build 32-bit version
+4. Run `dub build --arch=x86_64 --build=release` to build 64-bit version
 
 #### Remarks
 
-Currently only 64-bit release version is supported. Remeber to install 64-bit version of LLVM to the default destination (`Program Files`) and to add its binaries to PATH enviromental variable.
+Building 32-bit version requires a 32-bit variant of the Visual Studio toolchain to be present in `PATH`. The same for 64-bit. Remember *not* to specify `-arch=x86` when building 32-bit version (`dub` will overwrite `-m32mscoff` with unsupported `-m32` switch). If you use libclang versions <= 3.7, you may need to install MinGW for header files. Remember to install LLVM to its default installation path and to add its binaries to the `PATH` environmental variable (otherwise you may need to change `dub.json`).
 
 ## Usage
 
--- a/README.markdown
+++ b/README.markdown
@@ -21,13 +21,15 @@ The source code is available under the [Boost Software License 1.0](http://www.b
 
 ## Building
 
-### Requirements
+### Linux
+
+#### Requirements
 
 * libclang - [http://clang.llvm.org](http://clang.llvm.org) - 3.4 or later that is binary compatible with 3.4
 * DMD - [http://dlang.org/download.html](http://dlang.org/download.html) - 2.069.2
 * Dub [http://code.dlang.org/download](http://code.dlang.org/download)
 
-### Building
+#### Building
 
 1. Install all requirements, see [above](#requirements)
 2. Clone the repository by running:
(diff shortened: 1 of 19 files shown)


ciechowoj:
Tests for x86 works correctly.
Tests for x64 are failing (for some reason standard clang headers are not in the include search path).

ciechowoj:
I can remove x64 from the matrix and IMO it could be ready to be merged. I'll fix problems with x64 in separate PR.

ciechowoj:
For now I don't know what is going on. Tests for libcangs 3.4 - 3.5 are passing, and for more recent versions are failing due to missing standard header files : ( .

jacob-carlborg:
Hmm, on Posix I assume that the user have the SDK installed already. Is it just a matter of setting up some paths so Clang can find them?

ciechowoj:
All green : )

> Hmm, on Posix I assume that the user have the SDK installed already. Is it just a matter of setting up some paths so Clang can find them?

Perhaps the most recent version is looking in a different default directory?

The problem was with version 3.7. The earlier versions seems to use MinGW headers and later versions use Visual Studio headers. Version 3.7 do not detect neither by default (I don't really know why).

jacob-carlborg:
Is this both 32 and 64bit ?

ciechowoj:
Yes.

18.08.2016 22:07 "jacob-carlborg" <notifications@github.com> napisał(a):

clang/Cursor.d
```diff
@@ -281,6 +281,11 @@ struct Cursor
         foreach (cursor; unit.cursor.children)
             result[cursor.spelling] = cursor.kind;
 
+        auto version_ = clangVersion();
+
+        if (version_.major == 3 && version_.minor == 7)
+            result["__int64"] = CXCursorKind.CXCursor_MacroDefinition;
```
jacob-carlborg: What is this for?

README.markdown
```diff
@@ -36,6 +38,28 @@ The source code is available under the [Boost Software License 1.0](http://www.b
...
+1. Install all requirements, see [above](#requirements)
+2. Clone the repository by running:
+
+		$ git clone git://github.com/jacob-carlborg/dstep.git
+
+3. run `dub build --arch=x86_64 --build=release`
+
+#### Remarks
+
+Currently only 64-bit release version is supported. Remeber to install 64-bit version of LLVM to the default destination (`Program Files`) and to add its binaries to PATH enviromental variable.
```
jacob-carlborg: The documentation need to be updated now when Windows 32bit is supported.

appveyor.yml
```diff
@@ -0,0 +1,51 @@
...
+          Invoke-WebRequest http://llvm.org/releases/3.8.1/LLVM-3.8.1-win32.exe -OutFile c:\LLVM.exe;
+          7z x c:\LLVM.exe -oc:\PROGRA~2\LLVM > $null;
+        }
+
+  - set PATH=c:\dub;%PATH%
+  - dub --version
+
+before_build:
+  - set PATH=c:\dmd2\windows\bin;%PATH%
+  - ps: $env:compilersetup = "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcvarsall";
```
jacob-carlborg: Is the virtual machine always running Windows 64bit and we're cross-compiling to 32bit?

README.markdown
```diff
@@ -21,13 +21,15 @@ The source code is available under the [Boost Software License 1.0](http://www.b
 
 ## Building
 
-### Requirements
+### Linux
```
jacob-carlborg: Linux -> Posix.

jacob-carlborg:
It seems like the appveyor.yml file contains some configuration for running the tests. What if one wants to run the tests locally?

jacob-carlborg:
The discussion about moving some content of `appveyor.yml` to script files. You mentioned that it needed to set environment variables that are required in later steps. What is the advantage of using all these different sections, like "install", "build_script" and so on. It would be possible to put everything in one script file and invoke that from one of the sections, right?

At least I don't see any indication of these sections in the UI for a given build.

clang/Cursor.d
ciechowoj: Version 3.7 doesn't recognize __int64 and it is used in MinGW. It adds a macrodefinition for __int64 as long long.

appveyor.yml
ciechowoj: Yes.

ciechowoj:
> It seems like the appveyor.yml file contains some configuration for running the tests. What if one wants to run the tests locally?

Configuration in appveyor.yml it is mostly for CI. `dub --config=test` should run fine most of the time.

> The discussion about moving some content of appveyor.yml to script files. You mentioned that it needed to set environment variables that are required in later steps. What is the advantage of using all these different sections, like "install", "build_script" and so on. It would be possible to put everything in one script file and invoke that from one of the sections, right?

I'll check what can be done about it. 

This PR focuses on making it work for CI. I suppose some things can be improved to increase the probability that the end user builds dstep without problems. But it deserves its own PR.

clang/Cursor.d
jacob-carlborg: I see. Is it only that version that is using MinGW?

ciechowoj: No the versions 3.4, 3.5, 3.6 are using MinGW too, but they work fine (which is a bit strange...).

jacob-carlborg:
> This PR focuses on making it work for CI. I suppose some things can be improved to increase the probability that the end user builds dstep without problems. But it deserves its own PR.

The main reason why I would like to move it to separate file(s) is because it's then possible to take advantage of your editor: syntax highlighting, snippets, syntax validation and so on.

ciechowoj:
> The main reason why I would like to move it to separate file(s) is because it's then possible to take advantage of your editor: syntax highlighting, snippets, syntax validation and so on.

I see.

README.markdown
ciechowoj: Fixed.

README.markdown
ciechowoj: Fixed.

ciechowoj:
It took more time than it was worth,  but the script is in the separate file...

ciechowoj:
You need to go to AppVeyor and cancel some test because their got stuck.

jacob-carlborg:
Looking good, I just need the an update of the changelog to merge.

ciechowoj:
Changelog updated.

jacob-carlborg:
👌

ciechowoj:
Thanks.

jacob-carlborg:
BTW, have you tried using the OMF format? Since libclang is a DLL there's no problem, one just needs an import library in the correct format. This can be generated with `implib`, available here: http://www.digitalmars.com/download/freecompiler.html.

ciechowoj:
No, I haven't tried. I'll check it out when I have some free time.

PR #92, opened 2016-08-23
Improve CLI help.

ciechowoj:
The old help:

```
Usage: dstep [options] <input>
Version: v0.2.2-105-g583b72d

Options:
    -o, --output <file>          Write output to <file>.
    -o, --output <directory>     Write all the files to <directory>, in case of multiple input files.
    -ObjC, --objective-c         Treat source input file as Objective-C input.
    -x, --language <language>    Treat subsequent input files as having type <language>.
    -h, --help                   Show this message and exit.
    --no-comments                Disable translation of comments.

All options that Clang accepts can be used as well.

Use the `-h' flag for help.
```

a new one:
```
Usage: dstep [options] <input>
Version: v0.2.2-106-g1d08a02

Options:
    -o, --output <file>             Write output to <file>.
    -o, --output <directory>        Write all the files to <directory>, in case of multiple input files.
    -ObjC, --objective-c            Treat source input file as Objective-C input.
    -x, --language <language>       Treat subsequent input files as having type <language>.
    --version                       Show dstep version.
    --clang-version                 Show libclang version.
    --package <package>             Use <package> as package name.
    --comments                      Translate comments [default: true].
    --public-submodules             Use public imports for submodules [default: false].
    --reduce-aliases                Reduce primitive type aliases [default: true].
    --portable-wchar_t              Translate wchar_t as core.stdc.stddef.wchar_t [default: true].
    --zero-param-is-vararg          Translate functions with no arguments as variadic functions [default: false].
    --single-line-function-headers  Break function headers to multiple lines [default: false].
    --space-after-function-name     Put a space after a function name [default: true].
    -h, --help                      This help information.

All options that Clang accepts can be used as well. Use the `-h' flag for help. To disable boolean options use false, e.g. --comments=false.
```

### DIFF
--- a/dstep/Configuration.d
+++ b/dstep/Configuration.d
@@ -18,12 +18,15 @@ struct Configuration
     /// array of file names to translate to D
     string[] inputFiles;
 
+    /// expected programming language of input files
     @("language|x", "Treat subsequent input files as having type <language>.")
     Language language;
 
+    /// show dstep version
     @("version", "Show dstep version.")
     bool dstepVersion;
 
+    /// show libclang version
     @("clang-version", "Show libclang version.")
     bool clangVersion;
 
@@ -33,28 +36,36 @@ struct Configuration
     /// output file name or folder (in case there are many input files)
     string output;
 
+    /// package name
     @("package", "Use <package> as package name.")
     string packageName;
 
-    @("comments", "Translate comments [default: true].")
+    /// enable translation of comments
+    @("comments", "Translate comments [default].")
     bool enableComments = true;
 
-    @("public-submodules", "Use public imports for submodules [default: false].")
+    /// use public imports for submodules
+    @("public-submodules", "Use public imports for submodules [default].")
     bool publicSubmodules = false;
 
-    @("reduce-aliases", "Reduce primitive type aliases [default: true].")
+    /// enable reduction of primitive type aliases
+    @("reduce-aliases", "Reduce primitive type aliases [default].")
     bool reduceAliases = true;
 
-    @("portable-wchar_t", "Translate wchar_t as core.stdc.stddef.wchar_t [default: true].")
+    /// translate to wchar_t to core.stdc.stddef.wchar_t
+    @("portable-wchar_t", "Translate wchar_t as core.stdc.stddef.wchar_t [default].")
     bool portableWCharT = true;
 
-    @("zero-param-is-vararg", "Translate functions with no arguments as variadic functions [default: false].")
+    /// translate functions with empty argument list as vararg
+    @("zero-param-is-vararg", "Translate functions with no arguments as variadic functions [default].")
     bool zeroParamIsVararg = false;
 
-    @("single-line-function-headers", "Break function headers to multiple lines [default: false].")
-    bool singleLineFunctionHeaders = false;
+    /// single line function headers
+    @("single-line-function-signatures", "Keep function signatures in a single line [default].")
+    bool singleLineFunctionSignatures = false;
 
-    @("space-after-function-name", "Put a space after a function name [default: true].")
+    /// space after function name
+    @("space-after-function-name", "Put a space after a function name [default].")
     bool spaceAfterFunctionName = true;
 }
 
@@ -73,8 +84,36 @@ template makeGetOptArgs(alias config)
                 return &__traits(getMember, config, spelling);
             }
 
+            auto formatHelp(alias spelling)(string help)
+            {
+                import std.algorithm : canFind;
+                import std.format : format;
+                import std.string : replace;
+
+
+                Configuration config;
+
+                static if (is(typeof(__traits(getMember, config, spelling)) == bool))
+                {
+                    auto default_ = "[default]";
+
+                    if (help.canFind(default_))
+                        return help.replace(
+                            default_,
+                            format("[default: %s]", __traits(getMember, config, spelling)));
+                    else
+                        return help;
+                }
+                else
+                {
+                    return help;
+                }
+            }
+
             alias expand = AliasSeq!(
-                __traits(getAttributes, __traits(getMember, config, spelling)),
+                __traits(getAttributes, __traits(getMember, config, spelling))[0],
+                formatHelp!spelling(
+                    __traits(getAttributes, __traits(getMember, config, spelling))[1]),
                 ptr);
         }
         else
--- a/dstep/Configuration.d
+++ b/dstep/Configuration.d
@@ -18,8 +18,14 @@ struct Configuration
     /// array of file names to translate to D
     string[] inputFiles;
 
-    /// expected programming language of input files
+    @("language|x", "Treat subsequent input files as having type <language>.")
     Language language;
+
+    @("version", "Show dstep version.")
+    bool dstepVersion;
+
+    @("clang-version", "Show libclang version.")
+    bool clangVersion;
 
     /// array of parameters needed to be forwarded to clang driver
     string[] clangParams;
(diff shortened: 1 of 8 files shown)


dstep/Configuration.d
```diff
@@ -18,42 +18,71 @@ struct Configuration
...
     /// array of parameters needed to be forwarded to clang driver
     string[] clangParams;
 
     /// output file name or folder (in case there are many input files)
     string output;
 
-    /// strip all comments while translating
-    bool noComments;
-
-    /// package name
```
jacob-carlborg: Now we've lost the documentation, if we would generate it in the future. Should be possible to have both?

dstep/Configuration.d
```diff
@@ -18,42 +18,71 @@ struct Configuration
...
+    @("package", "Use <package> as package name.")
     string packageName;
 
-    /// use public imports for submodules
-    bool publicSubmodules;
+    @("comments", "Translate comments [default: true].")
```
jacob-carlborg: Is it possible to read the default value from the fields in the struct? Or is it too complicated?

dstep/main.d
```diff
@@ -56,19 +56,10 @@ auto parseCLI (string[] args)
...
-        "public-submodules", "Use public imports for submodules.", &config.publicSubmodules,
-        "package", "Specify package name.", &config.packageName,
-        "dont-reduce-aliases", "Disable reduction of primitive type aliases.", &config.dontReduceAliases,
-        "no-portable-wchar_t", "Translate wchar_t to wchar or dchar depending on its size.", &config.noPortableWCharT,
-        "single-line-function-headers", "Do not break function headers to multiple lines.", &config.singleLineFunctionHeaders,
-        "no-space-after-function-name", "Do not put a space after a function name.", &config.noSpaceAfterFunctionName,
-        "zero-param-is-vararg", "Translate functions with empty argument list as variadic functions.", &config.zeroParamIsVararg);
+        "output|o", &config.output,
+        "language|x", &parseLanguage,
+        "objective-c", &forceObjectiveC,
```
jacob-carlborg: Why are these three separate from the rest of `makeGetOptArgs`?

dstep/main.d
```diff
@@ -172,23 +163,70 @@ int main (string[] args)
...
-    writeln();
-    writeln("Use the `-h' flag for help.");
+
+    foreach (entry; entries)
+        writeln(format("    %-*s %s", cast(int) maxLength + 1, entry.option, entry.help));
+
+    writeln("");
+    writeln("All options that Clang accepts can be used as well. "
+        "Use the `-h' flag for help. "
+        "To disable boolean options use false, e.g. --comments=false.");
```
jacob-carlborg: Since most of this is change anyway, is it better to generate the whole string and call `writeln` just once?

dstep/main.d
ciechowoj: They cannot be generated automatically for different reasons.

`"output|o", &config.output,` has two entries in help.
`"language|x", &parseLanguage,` uses custom handler.
`"objective-c", &forceObjectiveC,` does not write directly to config, it writes to some local variable.

Hmm, actually `forceObjectiveC` does nothing. I suppose it is a bug. What was its original purpose?

dstep/Configuration.d
ciechowoj: I restored it.

dstep/Configuration.d
ciechowoj: I added the implementation that does this, but now it adds defaults to `--version`, `--clang-version`. If that is acceptable then its ok, if not then it cannot be done.

dstep/main.d
ciechowoj: Sure. Fixed.

dstep/main.d
jacob-carlborg: Treat source input file as Objective-C input. It's for the `-ObjC, --objective-c` flags. But I see now that they're parsed differently.

dstep/Configuration.d
jacob-carlborg: Hmm, I would prefer not to have a default value for those. Either revert back to the old implementation. Or add a new UDA for this, or add an additional value to the existing UDA. Not sure if it's best to be opt-in or opt-out.

dstep/main.d
jacob-carlborg: I think that all tests are using the `-ObjC` flag.

dstep/Configuration.d
ciechowoj: Hmm, actually I can do something like this: if user writes string like `[default]` it will be filled with the default value. The advantage over writing full default string manually, like `[default: true]` is the value will be kept in sync with real implementation.

dstep/main.d
ciechowoj: I removed `"objective-c", &forceObjectiveC,`.

dstep/Configuration.d
jacob-carlborg: Sounds good.

dstep/main.d
jacob-carlborg: I would like to keep the `objective-c` flag. It allows to have a consistent looks of the flags. `-ObjC` is not consistent but I added that because that's flag is available in Clang.

dstep/Configuration.d
ciechowoj: Done.

dstep/main.d
ciechowoj: I restored it and made it work.

ciechowoj:
> I suggest to move " To disable boolean options use false, e.g. --comments=false" on a separate line before "All options that Clang accepts" because it relates to main set of options.

Sure.

ciechowoj:
Current help appearance:
```
Usage: dstep [options] <input>
Version: v0.2.2-107-g1395abd

Options:
    -o, --output <file>             Write output to <file>.
    -o, --output <directory>        Write all the files to <directory>, in case of multiple input files.
    -ObjC, --objective-c            Treat source input file as Objective-C input.
    -x, --language <language>       Treat subsequent input files as having type <language>.
    --version                       Show dstep version.
    --clang-version                 Show libclang version.
    --package <package>             Use <package> as package name.
    --comments                      Translate comments [default: true].
    --public-submodules             Use public imports for submodules [default: false].
    --reduce-aliases                Reduce primitive type aliases [default: true].
    --portable-wchar_t              Translate wchar_t as core.stdc.stddef.wchar_t [default: true].
    --zero-param-is-vararg          Translate functions with no arguments as variadic functions [default: false].
    --single-line-function-headers  Break function headers to multiple lines [default: false].
    --space-after-function-name     Put a space after a function name [default: true].
    -h, --help                      This help information.

To disable boolean options use false, e.g. --comments=false.
All options that Clang accepts can be used as well.
Use the `-h' flag for help.
```

jacob-carlborg:
Isn't the description for the `--single-line-function-headers` flag backwards?

ciechowoj:
Yeah : P, I'll fix it in a moment.

dstep/Configuration.d
```diff
@@ -19,41 +19,109 @@ struct Configuration
...
 
     /// translate functions with empty argument list as vararg
-    bool zeroParamIsVararg;
+    @("zero-param-is-vararg", "Translate functions with no arguments as variadic functions [default].")
+    bool zeroParamIsVararg = false;
 
-    /// show dstep version
-    bool dstepVersion;
+    /// single line function headers
+    @("single-line-function-headers", "Break function headers to multiple lines [default].")
```
jacob-carlborg: `headers` -> `signatures`.

ciechowoj:
Updated version:
```
Usage: dstep [options] <input>
Version: v0.2.2-107-ge3b8b4f

Options:
    -o, --output <file>                Write output to <file>.
    -o, --output <directory>           Write all the files to <directory>, in case of multiple input files.
    -ObjC, --objective-c               Treat source input file as Objective-C input.
    -x, --language <language>          Treat subsequent input files as having type <language>.
    --version                          Show dstep version.
    --clang-version                    Show libclang version.
    --package <package>                Use <package> as package name.
    --comments                         Translate comments [default: true].
    --public-submodules                Use public imports for submodules [default: false].
    --reduce-aliases                   Reduce primitive type aliases [default: true].
    --portable-wchar_t                 Translate wchar_t as core.stdc.stddef.wchar_t [default: true].
    --zero-param-is-vararg             Translate functions with no arguments as variadic functions [default: false].
    --single-line-function-signatures  Keep function signatures in a single line. [default: false].
    --space-after-function-name        Put a space after a function name [default: true].
    -h, --help                         This help information.

To disable boolean options use false, e.g. --comments=false.
All options that Clang accepts can be used as well.
Use the `-h' flag for help.
```

dstep/Configuration.d
ciechowoj: Done.

jacob-carlborg:
Looking good. Waiting for the tests to complete.

ciechowoj:
The last commit was updated, because I needed to fix a mistake in `Common.d` I've make.

jacob-carlborg:
👌

jacob-carlborg:
Thanks.

jacob-carlborg:
BTW, I've noticed that when running DStep without any arguments, it doesn't print the help as it used to. Not sure when that was changed.

PR #94, opened 2016-08-28
Allow dub-less build with build.sh

Dicebot:
Makes possible to do a quick all-at-once build with no testing
using a shell script in absence of dub. Mostly relevant because
I remember Debian maintainer having issues with dub in its current
implementation and it would be nice to get dstep into Debian.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,8 +2,10 @@
 
 if [ -s "$HOME/.dvm/scripts/dvm" ] ; then
     . "$HOME/.dvm/scripts/dvm" ;
-    dvm use 2.069.2
+    dvm use 2.071.1
 fi
 
-dub build
-#rdmd --build-only -debug -gc -ofbin/dstep -Idstack/mambo -Idstack -L-L. -L-lclang -L-ltango -L-rpath -L. "$@" dstep/driver/DStep.d
+SOURCES=`find ./dstep ./clang -name \*.d`
+
+DC=${DC:-dmd}
+$DC $DFLAGS -g -O -inline -Jresources -L-lclang -ofbin/dstep $SOURCES


russel:
Dub is now in Debian repository (Sid and Testing) so it may be that packages in Debian can now be built using Dub. Debian has a policy of only using package build toolchains that are themselves in Debian.

jacob-carlborg:
> Debian has a policy of only using package build toolchains that are themselves in Debian

Wouldn't a shell script invoking the compiler be ok?

ximion:
Ideally having some kind of build system is very useful. It usually allows detecting dependencies reliably (allowing to prefer system-wide installations), compiles manual pages and additional data, recognizes global compiler flags set by the distributor, and in the end is able to install the resulting application and all its data to some location.

A `build.sh` script looks okay for this project, since it's light on dependencies (it should probably recognize the `$DFLAGS` environment variable though).
Build systems like Automake and Meson have additional benefits though, like (in case of Meson) faster compile times and reliable dependency-detection (which isn't done in the script).

As for Meson, I recently added support for D, which makes it a good option: http://forum.dlang.org/thread/dpakidboblzgbdtejilz@forum.dlang.org - only support for depfiles is missing, because GDC, LDC and DMD either output no depfiles, no GCC-compliant Makefiles or fail to build them.

Scripts like the `build.sh` thing were frowned upon for a while in Debian (mainly due to them producing a very high load and very high memory usage, which makes them fail on less-capable builders, while using `make` or `ninja`, the amount of running jobs can be adjusted to the system), but personally I think they are okay-ish (the bigger the project, the lesser they are a good idea though).

So, tl;dr:
Make sure the project builds with LDC or GDC, recognize `$DFLAGS` in the script, and that should be fine. If you want to go further, consider Meson/Automake/cmake support.

Also, for getting stuff into Debian, you would need someone to actually package and - more importantly - maintain the pieces. I don't have enough time to take on even more packages, but I would sponsor packages (see also https://wiki.debian.org/DebianMentorsFaq ).
Now that we have a revived D team, getting people to help with packages will be much easier.

russel:
I hesitated suggesting SCons which would fit the conditions, as I knew about the rise of Meson. SCons is clearly superior :-) but Meson is the move from Autotools that is being envisaged in the GNOME arena. I had been intending to add D support to Meson as I did to SCons, but it seems I have been beaten to it. No matter my ties to SCons, if Meson is the system to replace Autotools, then I am behind it. My efforts in GStreamer Bad plugins will be Meson driven, but this is C code. I will though switch my D projects from SCons to Meson now and am happy to join in making Meson work with the Dub repository – I will also echo this for SCons. I guess I should contact @ximion via a different channel to set this up. For this issue the headline is clearly build with LDC as in the Debian repository and use Meson. Obviously there can be a Dub build as well. I guess I am volunteering for the Debian packaging D team.

jacob-carlborg:
I'm not a big fan of adding additional build systems, especially not Automake or cmake, I'm not familiar with Meson.

russel:
@jacob-carlborg Autotools I try to avoid like the plague. CMake I use under sufferance (usually because I want to use CLion as IDE). Meson is shaping up to be the CMake and Autotools replacement, picking the best from CMake, Autotools, Waf and SCons. I shall be using it in the GStreamer context. On the other hand Dub is now in the Debian repository and so it should be possible to ease it into the list of acceptable build tools and so avoid a Meson build.

ximion:
@russel No SCons please ;-) https://wiki.debian.org/UpstreamGuide#SCons
Autotools is fine for Debian, but really annoying to work with for upstream projects, I try to avoid it whereever possible as well.
Meson and D are working quite well together, except for the depfiles issue which sometimes makes it not recompile dependencies (and then run into link errors). I haven't had the time yet to investigate this and file bugs against the compilers (or whatever else one can blame this for).
Any help is welcome! DUB support in Meson would be an amazing thing!

jacob-carlborg:
> DUB support in Meson would be an amazing thing!

What would Meson add that Dub cannot do alone?

ximion:
@jacob-carlborg Fix all the issues which make dub a pain to use for distributors, mainly resolving all the issues I listed at https://gist.github.com/ximion/77dda83a9926f892c9a4fa0074d6bf2b#dub (most of them are also reported as bugs against dub already).
Additionally, it allows you to easily have multiple libraries and binaries in one project, conditionally compile additional resources at build-time (e.g. GRessources, ProtoBuf schemas, rendering icons to different sizes, ...), install stuff and also mix C and D code in one project, which is incredibly useful.
Dub support in Meson would be cool to depend on projects not using Meson easier (although Meson has a system to pull dependencies easily if they are using Meson, or to patch-in Meson support if they don't).

jacob-carlborg:
@Dicebot can you adjust the build script to support the `$DFLAGS` environment variable?

PR #95, opened 2016-08-31
Fix #8: Typedef and anonymous structs.

### LINKED ISSUE #8: typedef and anonymous structs
Something like this:

typedef struct { ... } struct_name;

is a common pattern in c. It's just a way to rename an anonymous struct.

dstep split this in two different declaration

struct anonymous
{
}

and 

alias anonymous struct_name;

Original anonymous struct will be never used and can't be called from C code, so I think this two declaration can be merged in a single named struct.

ciechowoj:
As in the title.

I think the major issue of the issue (: P) was fixed some time ago, however this PR makes DStep to rename `struct Foo` to `Foo_` if there is already a symbol with the same name in the global namespace.

It cleans up the implementation of `MacroIndex` too.

### DIFF
--- a/dstep/Configuration.d
+++ b/dstep/Configuration.d
@@ -77,6 +77,14 @@ struct Configuration
     /// do not translate following symbols
     @("skip", "Skip translation of <symbol>.")
     string[] skipSymbols;
+
+    /// print diagnostic informations
+    @("print-diagnostics", "Print diagnostic informations [default].")
+    bool printDiagnostics = true;
+
+    /// action to take on symbol collision
+    @("collision-action", "Action to take when translated symbol collides with a preexisting symbol [default].")
+    CollisionAction collisionAction = CollisionAction.rename;
 }
 
 template makeGetOptArgs(alias config)
@@ -85,9 +93,11 @@ template makeGetOptArgs(alias config)
 
     template expand(alias spelling)
     {
+        alias member = Alias!(__traits(getMember, config, spelling));
+
         static if (
             __traits(compiles, &__traits(getMember, config, spelling)) &&
-            __traits(getAttributes, __traits(getMember, config, spelling)).length == 2)
+            __traits(getAttributes, member).length == 2)
         {
             auto ptr() @property
             {
@@ -96,28 +106,51 @@ template makeGetOptArgs(alias config)
 
             auto formatHelp(alias spelling)(string help)
             {
-                import std.algorithm : canFind;
-                import std.format : format;
-                import std.string : replace;
-
+                import std.algorithm;
+                import std.format;
+                import std.string;
+                import std.range;
 
                 Configuration config;
 
-                static if (is(typeof(__traits(getMember, config, spelling)) == bool))
+                string suffix;
+
+                static if (is(typeof(member) == bool) || is(typeof(member) == enum))
                 {
                     auto default_ = "[default]";
 
                     if (help.canFind(default_))
-                        return help.replace(
+                    {
+                        help = help.replace(
                             default_,
                             format("[default: %s]", __traits(getMember, config, spelling)));
-                    else
-                        return help;
+
+                        static if (is(typeof(member) == bool))
+                        {
+                            suffix = "=true|false";
+                        }
+                        else
+                        {
+                            suffix = format(
+                                "=%s",
+                                join([ __traits(allMembers, typeof(member)) ], "|"));
+                        }
+                    }
                 }
                 else
                 {
-                    return help;
+                    auto beginning = findSplitBefore(help, "<");
+
+                    if (!beginning[0].empty)
+                    {
+                        auto placeholder = findSplitAfter(beginning[1], ">");
+
+                        if (!placeholder[0].empty)
+                            suffix = format(" %s", placeholder[0]);
+                    }
                 }
+
+                return format("%s!%s", suffix, help);
             }
 
             alias expand = AliasSeq!(
--- a/dstep/translator/Context.d
+++ b/dstep/translator/Context.d
@@ -32,6 +32,7 @@ class Context
     private Translator translator_ = null;
     private Output globalScope_ = null;
     private Cursor[string] typeNames_;
+    private string[Cursor] translatedSpellings;
     public MacroDefinition[string] macroDefinitions;
 
     Options options;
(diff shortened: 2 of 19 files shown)


dstep/Configuration.d
```diff
@@ -77,6 +77,10 @@ struct Configuration
     /// do not translate following symbols
     @("skip", "Skip translation of <symbol>.")
     string[] skipSymbols;
+
+    /// enable printing of warnings
+    @("enable-warnings", "Enable printing of warnings [default].")
+    bool enableWarnings = true;
```
jacob-carlborg: I'm thinking this can be confused with enabling warnings for Clang.

ciechowoj: Good point. Any ideas for replacement?

ciechowoj: Maybe `enable-dstep-warnings` or `dstep-warnings`.

jacob-carlborg: What about a flag indicating how the DStep should behave when it encounters something it cannot directly translate? The values of the flag could be: error, warning and ignore.

ciechowoj: Ok, it can be done (even `getopt` supports it) but there are two issues:

* how the option should be named
* there are more cases than `error`, `warning` and `ignore`: `error`, `warn_and_skip`, `warn_and_fix`, `ignore_and_skip` and `ignore_and_fix`.

My point is, that there are two orthogonal concepts: issuing a message and trying to fix the problem.

jacob-carlborg: > My point is, that there are two orthogonal concepts: issuing a message and trying to fix the problem.

That's true. If we start with making issuing a message configurable?

ciechowoj: Yes, the fixing my be enabled by default.

01.09.2016 13:34 "jacob-carlborg" <notifications@github.com> napisał(a):

ciechowoj: So, what about `--warning-level=warn`,  `--on-name-collision=warn`, `--message-level=warn` or `--on-recoverable-error=abort|warn|ignore`, `--on-recoverable-conflict=...`, `--on-fixable-problem=...` ?
Truth say I cannot find good name ; P

jacob-carlborg: `--diagnostic-level`?

ciechowoj: Yes, I suppose it's the best.

jacob-carlborg: Hmm, we might not need different levels. `--print-diagnostics` would be a boolean option. Then (at a later time) we could add `--diagnostic-action` which could be `ignore`, `abort`, `try-fix` and so on.

ciechowoj: Could you tell me what help message should I put for `diagnostic-action`?

jacob-carlborg: "Action to take when a diagnostic issue is encountered", perhaps?

ciechowoj: I was thinking about something similar, but I do not like the term
'diagnostic issue', it doesn't express the spirit of the problem exactly.
But I've just had an idea. What about ´ambiguity-action´ and "Action to
take when an ambiguity is encountered.". My point it is not the diagnostics
that needs an action, it is some kind of conflict/ambiguity.

02.09.2016 15:44 "jacob-carlborg" <notifications@github.com> napisał(a):

jacob-carlborg: True, but I was thinking that it could be used for cases where DStep currently outputs `<unimplemented>` as well. But perhaps that should be separate flags.

ciechowoj: I added two new options: `--print-diagnostics` and `--collision-action`. A collision action is an enum, so I needed to change the help to accommodate for that. Here is the new help:

```
Usage: dstep [options] <input>
Version: v0.2.2-112-gf88c046

Options:
    -o, --output <file>                           Write output to <file>.
    -o, --output <directory>                      Write all the files to <directory>, in case of multiple input files.
    -ObjC, --objective-c                          Treat source input file as Objective-C input.
    -x, --language                                Treat subsequent input files as having type <language>.
    --version                                     Show dstep version.
    --clang-version                               Show libclang version.
    --package <package>                           Use <package> as package name.
    --comments=true|false                         Translate comments [default: true].
    --public-submodules=true|false                Use public imports for submodules [default: false].
    --reduce-aliases=true|false                   Reduce primitive type aliases [default: true].
    --portable-wchar_t=true|false                 Translate wchar_t as core.stdc.stddef.wchar_t [default: true].
    --zero-param-is-vararg=true|false             Translate functions with no arguments as variadic functions [default: false].
    --single-line-function-signatures=true|false  Keep function signatures in a single line [default: false].
    --space-after-function-name=true|false        Put a space after a function name [default: true].
    --skip-definition <symbol>                    Keep only the signature of <symbol>.
    --skip <symbol>                               Skip translation of <symbol>.
    --print-diagnostics=true|false                Print diagnostic informations [default: true].
    --collision-action=ignore|rename|abort        Action to take when translated symbol collides with a preexisting symbol [default: rename].
    -h, --help                                    This help information.

All options that Clang accepts can be used as well.
Use the `-h' flag for help.
```

Tell me what do you think.

ciechowoj:
I added two new options: `--print-diagnostics` and `--collision-action`. A collision action is an enum, so I needed to change the help to accommodate for that. Here is the new help:

```
Usage: dstep [options] <input>
Version: v0.2.2-112-gf88c046

Options:
    -o, --output <file>                           Write output to <file>.
    -o, --output <directory>                      Write all the files to <directory>, in case of multiple input files.
    -ObjC, --objective-c                          Treat source input file as Objective-C input.
    -x, --language                                Treat subsequent input files as having type <language>.
    --version                                     Show dstep version.
    --clang-version                               Show libclang version.
    --package <package>                           Use <package> as package name.
    --comments=true|false                         Translate comments [default: true].
    --public-submodules=true|false                Use public imports for submodules [default: false].
    --reduce-aliases=true|false                   Reduce primitive type aliases [default: true].
    --portable-wchar_t=true|false                 Translate wchar_t as core.stdc.stddef.wchar_t [default: true].
    --zero-param-is-vararg=true|false             Translate functions with no arguments as variadic functions [default: false].
    --single-line-function-signatures=true|false  Keep function signatures in a single line [default: false].
    --space-after-function-name=true|false        Put a space after a function name [default: true].
    --skip-definition <symbol>                    Keep only the signature of <symbol>.
    --skip <symbol>                               Skip translation of <symbol>.
    --print-diagnostics=true|false                Print diagnostic informations [default: true].
    --collision-action=ignore|rename|abort        Action to take when translated symbol collides with a preexisting symbol [default: rename].
    -h, --help                                    This help information.

All options that Clang accepts can be used as well.
Use the `-h' flag for help.
```

Tell me what do you think.

dstep/translator/Context.d
```diff
@@ -136,16 +137,120 @@ class Context
         return typedefIndex_.typedefParent(cursor);
     }
 
+    private string translateSpellingImpl(in Cursor cursor)
+    {
+        return cursor.spelling == ""
```
jacob-carlborg: Don't you need to check the spelling of the type?

dstep/translator/Context.d
```diff
@@ -136,16 +137,120 @@ class Context
...
+        import std.stdio : writeln;
+
+        if (options.printDiagnostics)
+        {
+            auto cursorLocation = cursor.location;
+            auto collisionLocation = collision.location;
+
+            auto message = format(
+                "%s:%d:%d: warning: a type renamed to '%s' due to the "
+                "collision with the symbol declared in %s:%d:%d",
```
jacob-carlborg: Use explicit concatenation.

dstep/translator/Context.d
```diff
@@ -136,16 +137,120 @@ class Context
...
+            format(
+                "%s:%d:%d: error: a type name '%s' "
+                "collides with the symbol declared in %s:%d:%d",
+                cursorLocation.path,
+                cursorLocation.line,
+                cursorLocation.column,
+                spelling,
+                collisionLocation.path,
+                collisionLocation.line,
+                collisionLocation.column));
```
jacob-carlborg: Perhaps move the formatting of the error message to a separate function since it's the same as for the warning.

dstep/translator/MacroIndex.d
```diff
@@ -20,66 +21,63 @@ import dstep.translator.Preprocessor;
...
-            else if (cursor.kind == CXCursorKind.CXCursor_MacroDefinition)
-            {
-                recent[cursor.spelling] = cursor;
-            }
+            if (!cursor.spelling.empty)
+                globalCursors_[extendedSpelling(cursor)] = cursor;
         }
 
+        lessOp = translUnit.relativeCursorLocationLessOp();
+        expansions = expansionsAppender.data.sort!((a, b) => lessOp(a, b)).array;
```
jacob-carlborg: Can't `lessOp` be passed directly?

dstep/translator/MacroIndex.d
```diff
@@ -20,66 +21,63 @@ import dstep.translator.Preprocessor;
...
     Cursor[] queryExpansion(Cursor cursor) const
     {
         import std.array;
         import std.algorithm.searching;
 
         SourceRange extent = cursor.extent;
 
-        auto equal = expansions.equalRange(cursor);
-        auto greater = expansions.upperBound(cursor);
+        auto expansionsSorted = assumeSorted!((a, b) => lessOp(a, b))(expansions);
```
jacob-carlborg: Please use UFCS.

dstep/translator/MacroIndex.d
ciechowoj: No, it can't. It is private member of MacroIndex and is inaccessible from inside of `sort`.

dstep/translator/Context.d
ciechowoj: To detect collisions?

dstep/translator/MacroIndex.d
ciechowoj: Fixed.

dstep/translator/Context.d
ciechowoj: Fixed.

dstep/translator/Context.d
ciechowoj: I rewrote the functions.

dstep/translator/Context.d
jacob-carlborg: Never mind.

dstep/translator/MacroIndex.d
jacob-carlborg: Ah, I see.

unit_tests/Common.d
```diff
@@ -610,53 +631,105 @@ void assertRunsDStep(
             auto extra = findExtraGNUStepPaths(file, line);
 
             if (extra.empty)
-                return;
+                throw new NoGNUStepException();
```
jacob-carlborg: Why is this now throwing an exception?

ciechowoj: The code was used in a single place before. Now the function is reused in two places. I used exception to communicate the outer functions that the GNUStep is no available, so that thay can return silently (as before).

jacob-carlborg:
👌

jacob-carlborg:
The changelog should have been updated. Please create a new PR for that.

PR #100, opened 2016-11-09
Sort entry array after generating

Dicebot:
Otherwise `assumeSorted` contract will fail on certain inputs. Specifically it was failing during conversion of `libmpdclient` headers:

```
 dstep -I/usr/lib/clang/3.5.2/include/ -o source/c /usr/include/mpd/*
core.exception.AssertError@/usr/include/dlang/dmd/std/range/package.d(7625): Range is not sorted
```

### DIFF
--- a/clang/TranslationUnit.d
+++ b/clang/TranslationUnit.d
@@ -234,6 +234,11 @@ struct TranslationUnit
         foreach (uint index, location; locations)
             map[location.path] ~= Entry(index, location);
 
+        import std.algorithm;
+
+        foreach (key, ref value; map)
+            sort(value);
+
         uint findIndex(SourceLocation a)
         {
             auto entries = map[a.path];


jacob-carlborg:
Could you please add a test.

mihails-strasuns-sociomantic:
Checked this again and I can't get a small repro yet. What I have figured out so far is that problem is related to not being able to resolve sub-includes when translating header directly from `/usr/include` - copying the folder to work dir and translating from there works just fine. With absolute path I had entity array like this:

```
[Entry(1, SourceLocation(file = /usr/include/mpd/entity.h, line = 1, column = 1, offset = 0), 7F75267951C0), Entry(49, SourceLocation(file = /usr/include/mpd/entity.h, line = 42, column = 22, offset = 1793), 7F75267951C0), Entry(51, SourceLocation(file = /usr/include/mpd/entity.h, line = 43, column = 27, offset = 1820), 7F75267951C0), Entry(75, SourceLocation(file = /usr/include/mpd/entity.h, line = 1, column = 1, offset = 0), 7F75267951C0), Entry(113, SourceLocation(file = /usr/include/mpd/entity.h, line = 44, column = 24, offset = 1844), 7F75267951C0)]
```

I will probably not bother reducing it for now as bug is not critical and is easily worked around. Closing.

jacob-carlborg:
Closing as requested.

PR #103, opened 2016-12-18
Fix #102: Regression: usage of typedefs within structs.

### LINKED ISSUE #102: Regression: usage of typedefs within structs
C code:

```C
typedef struct _ProtobufCMethodDescriptor ProtobufCMethodDescriptor;

struct _ProtobufCMethodDescriptor
{
    const char *name;
    const ProtobufCMethodDescriptor *input;
    const ProtobufCMethodDescriptor *output;
};
```

Generated D code with 3e154a6d201dc8494a281ff8c9660a01194a1710 :

```D
extern (C):

struct _ProtobufCMethodDescriptor
{
    alias _ProtobufCMethodDescriptor ProtobufCMethodDescriptor;
    const(ProtobufCMethodDescriptor)* input;
    alias _ProtobufCMethodDescriptor ProtobufCMethodDescriptor;
    const(ProtobufCMethodDescriptor)* output;
}

alias _ProtobufCMethodDescriptor ProtobufCMethodDescriptor;
```

Duplicate injected alias inside a struct causes compilation error. Used to not inject such aliases before, though I haven't tracked regression commit yet.

ciechowoj:
As in the title.

### DIFF
--- a/clang/Cursor.d
+++ b/clang/Cursor.d
@@ -180,7 +180,7 @@ struct Cursor
         return Cursor.empty;
     }
 
-    Cursor findChild(CXCursorKind kind)
+    Cursor findChild(CXCursorKind kind) const
     {
         foreach (child; all)
         {
--- a/clang/SourceRange.d
+++ b/clang/SourceRange.d
@@ -68,3 +68,9 @@ bool intersects(in SourceRange a, in SourceRange b)
         (a.start.offset <= b.start.offset && b.start.offset < a.end.offset) ||
         (a.start.offset < b.end.offset && b.end.offset <= a.end.offset);
 }
+
+bool contains(in SourceRange a, in SourceRange b)
+{
+    return a.path == b.path &&
+        a.start.offset <= b.start.offset && b.end.offset <= a.end.offset;
+}
--- a/dstep/translator/Context.d
+++ b/dstep/translator/Context.d
@@ -9,6 +9,7 @@ module dstep.translator.Context;
 import clang.c.Index;
 import clang.Cursor;
 import clang.TranslationUnit;
+import clang.SourceRange;
 
 import dstep.translator.CommentIndex;
 import dstep.translator.IncludeHandler;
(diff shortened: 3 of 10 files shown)


ciechowoj:
It seems like there are problems with running the tests, and more precisely downloading clang binaries.

```
Downloading clang 3.4.2
object.Exception@unit_tests/HttpClient.d(14): HTTP answered 302 instead of 200 on http://llvm.org/releases/3.4.2/clang+llvm-3.4.2-x86_64-unknown-ubuntu12.04.xz
```

ciechowoj:
> Originally I haven't implemented redirects when removing external http client dependency as those weren't necessary. Looks like they are now. Something as simple as modifying https://github.com/jacob-carlborg/dstep/blob/master/unit_tests/HttpClient.d#L13-L14 to check 302 status and Location header in a loop should do the trick.

Good idea.

dub.json
```diff
@@ -29,7 +29,7 @@
             "sourcePaths": ["dstep", "clang"],
             "importPaths": ["dstep", "clang"],
             "dflags-windows-x86": ["-m32mscoff"],
-            "lflags-posix": ["-lclang", "-rpath", ".", "-rpath", "/usr/lib/llvm-3.7/lib", "-L.", "-L/usr/lib64/llvm", "-L/usr/lib/llvm-3.7/lib"],
+            "lflags-posix": ["-lclang", "-rpath", ".", "-rpath", "/usr/lib/llvm-3.8/lib", "-L.", "-L/usr/lib64/llvm", "-L/usr/lib/llvm-3.8/lib"],
```
jacob-carlborg: Please don't modify this. I'll remove this eventually when I've finished support for static and dynamic linking of libclang.

ciechowoj: Ok.

jacob-carlborg:
👌

jacob-carlborg:
Thanks.

PR #104, opened 2016-12-24
Fix extern variable generation for D1 version

Dicebot:
Trivial regression noticed when testing current dstep master with our code base

### DIFF
--- a/dstep/translator/Translator.d
+++ b/dstep/translator/Translator.d
@@ -217,7 +217,12 @@ class Translator
 
     void translateVarDecl(Output output, Cursor cursor, Cursor parent)
     {
-        variable(output, cursor, "extern __gshared ");
+        version (D1)
+            string storageClass = "extern ";
+        else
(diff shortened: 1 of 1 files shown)


jacob-carlborg:
Would it be too complicated to test the D1 output?

dstep/translator/Translator.d
```diff
@@ -217,7 +217,10 @@ class Translator
 
     void translateVarDecl(Output output, Cursor cursor, Cursor parent)
     {
-        variable(output, cursor, "extern __gshared ");
+        version (D1)
+            variable(output, cursor, "extern ");
+        else
+            variable(output, cursor, "extern __gshared ");
```
jacob-carlborg: Please put the string in a local variable and version that instead of the whole call.

Dicebot: Updated.

jacob-carlborg:
👌

PR #105, opened 2016-12-31
Add support for configuring static/dynamic linking of libclang

jacob-carlborg:
@Dicebot @ciechowoj I've created this pull request to get some feedback on this code. It's tagged as WIP because I've only tested it on macOS so far. Here's the commit message that contains the description how the configure script works:

> This is done by adding a "configure" script. When the "configure"
script is called it will look for libclang in the most common paths
for the given platform and pick the latest version available that is
supported by DStep. When the "configure" script is run it will write a
file, "config.txt", in the root directory of DStep. This file contains
linker flags according to the specified configuration.
>
> When Dub is run, a pre generate command will look if the "config.txt"
file exists, if it doesn't, Dub will automatically invoke the
"configure" script. This file is later passed as a flag to the linker,
which will pickup the linker flags for the specified configuration.
>
> It's possible to manually invoke the "configure" script to change the
default configuration. The following parameters can be configured:
> 
> * Static or dynamic linking of libclang
> 
> * The version/path of libclang to use by specifying the path to the
"llvm-config" binary that is shipped with LLVM
> 
> * The path to the "ncurses" library (only used when static linking
is used)

### DIFF
--- a/configure
+++ b/configure
@@ -1,0 +1,7 @@
+#!/bin/sh
+
+# This wrapper script is necessary due to https://github.com/dlang/dub/issues/910
+
+set -e
+
+dub configure.d "$@"
--- a/configure.d
+++ b/configure.d
@@ -1,0 +1,631 @@
+/+ dub.sdl: name "configure" +/
+/**
+ * This file implements a configuration script that will setup the correct flags
+ * to link with libclang.
+ *
+ * The script will try to automatically detect the location of libclang by
+ * searching through a number of preset library search paths for different
+ * platforms.
+ *
+ * If the script fails to find libclang or fails to find the correct version,
+ * this script provides a flag,`--llvm-config`, which can be used by manually
+ * executing this script (`./configure`) and specifying the path to the LLVM
+ * configuration binary, `llvm-config`. The LLVM configuration binary will then
+ * be used to find the location of libclang.
+ *
+ * The result of invoking this configuration script is a file,
+ * `linker_flags.txt`, which will be created. This file contains the necessary
+ * linker flags which will be read by the linker when building DStep.
+ *
+ * This script is only intended to be used on the Posix platforms.
+ */
+module configure;
+
+import std.algorithm;
+import std.array;
+import std.conv;
+import std.exception;
+import std.file;
+import std.format;
+import std.getopt;
+import std.path;
+import std.process;
+import std.range;
+import std.string;
+import std.uni;
+import std.traits;
+
+version (Posix)
+{
+    version (OSX) {}
+    else version (linux) {}
+    else version (FreeBSD) {}
+    else
+        static assert("The current platform is not supported");
+}
+else
+    static assert("This script should only be run on Posix platforms");
+
+/**
+ * The options of the application.
+ *
+ * When parsing the command line arguments, these fields will be set.
+ */
+struct Options
+{
+    /// Print extra information.
+    bool verbose = true;
+
+    /// Indicates if help/usage information was requested.
+    bool help = false;
+
+    /// The specified path to the LLVM/Clang library path.
+    string llvmLibPath;
+
+    /// The specified path to the location of the `ncurses` library.
+    string ncursesLibPath;
+
+    /// Indicates if libclang should be statically or dynamically linked.
+    bool dynamicClang = true;
+}
+
+/// Default configuration and paths.
+struct DefaultConfig
+{
+static:
+
+    version (D_Ddoc)
+    {
+        /// The name of the Clang dynamic library.
+        enum clangLib = "";
+
+        /**
+         * A list of default paths where to look for the LLVM and Clang
+         * libraries.
+         */
+        immutable string[] llvmLibPaths = [];
+
+        /**
+         * A list of default paths where to look for additional libraries.
+         *
+         * Thes are libraries that are not part of LLVM or Clang which are used
+         * when statically linking libclang.
+         */
+        immutable string[] additionalLibPaths = [];
+
+        /**
+         * The name of the Ncurses static library.
+         *
+         * Used when statically linking libclang.
+         */
+        enum ncursesLib = "";
+
+        /**
+         * The name of the C++ standard library.
+         *
+         * Used when statically linking libclang.
+         */
+        enum cppLib = "c++";
+    }
+
+    else version (OSX)
+    {
+        enum clangLib = "libclang.dylib";
+
+        enum standardPaths = [
+            "/usr/local/lib",
+            "/usr/lib"
+        ];
+
+        immutable llvmLibPaths = [
+            "/opt/local/libexec/llvm-4.0/lib", // MacPorts
+            "/usr/local/opt/llvm40/lib", // Homebrew
+            "/opt/local/libexec/llvm-3.9/lib", // MacPorts
+            "/usr/local/opt/llvm39/lib", // Homebrew
+            "/opt/local/libexec/llvm-3.8/lib", // MacPorts
+            "/usr/local/opt/llvm38/lib", // Homebrew
+            "/opt/local/libexec/llvm-3.7/lib", // MacPorts
+            "/usr/local/opt/llvm37/lib" // Homebrew
+        ] ~ standardPaths;
+
+        immutable additionalLibPaths = [
+            "/opt/local/lib"
+        ] ~ standardPaths;
+
+        enum ncursesLib = "libncurses.a";
+        enum cppLib = "c++";
+    }
+
+    else version (linux)
+    {
+        enum clangLib = "libclang.so";
+
+        enum standardPaths = [
+            "/usr/lib",
+            "/usr/local/lib",
+            "/usr/lib/x86_64-linux-gnu" // Debian
+        ];
+
+        immutable llvmLibPaths = [
+            "/usr/lib/llvm-4.0/lib", // Debian
+            "/usr/lib/llvm-3.9/lib", // Debian
+            "/usr/lib/llvm-3.8/lib", // Debian
+            "/usr/lib/llvm-3.7/lib" // Debian
+        ] ~ standardPaths;
+
+        immutable additionalLibPaths = standardPaths;
+
+        enum ncursesLib = "libncurses.a";
+        enum cppLib = "stdc++";
+    }
+
+    else version (FreeBSD)
+    {
+        enum clangLib = "libclang.so";
+
+        enum standardPaths = [
+            "/usr/lib",
+            "/usr/local/lib"
+        ];
+
+        immutable llvmLibPaths = [
+            "/usr/lib/llvm-4.0/lib",
+            "/usr/lib/llvm-3.9/lib",
+            "/usr/lib/llvm-3.8/lib",
+            "/usr/lib/llvm-3.7/lib"
+        ] ~ standardPaths;
+
+        immutable additionalLibPaths = standardPaths;
+
+        enum ncursesLib = "libncurses.a";
+        enum cppLib = "stdc++";
+    }
+
+    else
+        static assert(false, "Unsupported platform");
+
+    /// The name of the LLVM configure binary.
+    enum llvmConfigExecutable = "llvm-config";
+}
+
+/**
+ * This class represents a path to a file, like a library or an executable.
+ *
+ * It's the abstract base class for the `LibraryPath` and `LLVMConfigPath`
+ * subclasses.
+ */
+class Path
+{
+    private
+    {
+        /**
+         * The name of the file this path represents.
+         *
+         * This is a name for the file that is used in error messages.
+         */
+        string name;
+
+        /**
+         * A set of standard paths to which to search for the file this path
+         * represents.
+         */
+        const(string)[] standardPaths;
+
+        /**
+         * The custom path that was specified when invoking this configuration
+         * script, or `null` if no custom path was specified.
+         */
+        string specifiedPath;
+
+        /// The actual file to look for in `standardPaths` and `specifiedPath`.
+        string fileToCheck;
+
+        /// Local cache for the full path to the file.
+        string path_;
+    }
+
+    alias path this;
+
+    /**
+     * Constructs a new instance of this class.
+     *
+     * Params:
+     *  name = the name of the file this path represents
+     *
+     *  standardPaths = a set of standard paths to which to search for the file
+     *      this path represents
+     *
+     *  specifiedPath = the custom path that was specified when invoking this
+     *      configuration script, or `null` if no custom path was specified
+     *
+     *  fileToCheck = the actual file to look for in `standardPaths` and
+     *      `specifiedPath`
+     */
+    this(string name, const(string)[] standardPaths,
+        string specifiedPath, string fileToCheck)
+    {
+        this.name = name;
+        this.standardPaths = standardPaths;
+        this.specifiedPath = specifiedPath;
+        this.fileToCheck = fileToCheck;
+    }
+
+    /**
+     * Returns the full path to the file this path represents as a string.
+     *
+     * If `specifiedPath` is non-empty, `fileToCheck` will be searched for in
+     * `specifiedPath`. Otherwise `fileToCheck` will be searched for in
+     * `standardPaths`.
+     *
+     * Returns: the full path to the file this path represents
+     */
+    string path()
+    {
+        if (path_.ptr)
+            return path_;
+
+        return path_ = specifiedPath.empty ? standardPath : customPath;
+    }
+
+    override string toString()
+    {
+        return path;
+    }
+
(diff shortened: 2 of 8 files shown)


jacob-carlborg:
Seems that's not possible, I assume because you don't have permission.

configure
```diff
@@ -0,0 +1,459 @@
+#!/usr/bin/env dmd -run configure
```
Dicebot: Using `rdmd` would avoid rebuilds

configure
```diff
@@ -0,0 +1,459 @@
...
+        ];
+    }
+
+    else version (linux)
+    {
+        enum cppLib = "stdc++";
+        enum clangLib = "libclang.so";
+
+        immutable llvmLibPaths = [
+            "/usr/lib/llvm-3.9/lib",
```
Dicebot: Do you intend to manually add new entries when new llvm releases come out?

configure
```diff
@@ -0,0 +1,459 @@
...
+        ];
+
+        immutable systemLibPahts = [
+            "/usr/local/lib",
+            "/usr/lib",
+            "lib"
+        ];
+    }
+
+    enum ncursesLib = "libncurses.a";
```
Dicebot: At least on Arch Linux ncurses package only provides `libcurses.so`

configure
```diff
@@ -0,0 +1,459 @@
...
+class Path
+{
+    immutable string path;
+
+    private
+    {
+        immutable string name;
+        const(string)[] standardPaths;
+        immutable string specifiedPath;
+        immutable string fileToCheck;
```
Dicebot: A bit hard to understand what each of these fields refer to without comments

configure
```diff
@@ -0,0 +1,459 @@
...
+    string standardPath() const
+    {
+        immutable errorMessage = format("Could not find %s in any of the " ~
+            "standard paths for %s: \n%s\nPlease specify a path manually " ~
+            "using './configure --%s-path=<path>'.",
+            fileToCheck, name, standardPaths.join('\n'), name.toLower
+        );
+
+        const result = standardPaths.
+            find!(exists).
```
Dicebot: I think idiomatic D range pipeline style is with dot at the beginning of new line :)

configure
Dicebot: btw why does dstep need ncurses at all?

ciechowoj:
It can be queried at runtime `clang.Util.Version clangVersion()`.

ciechowoj:
`CXString clang_getClangVersion();` is present in libclang at least since version 2.7 which I think is more than enough.

ciechowoj:
Oh, you are right.

jacob-carlborg:
> I don't see much OOP code in D these days, hard to reason about it :)

I tried a different approach first, it was a mess :)

configure
jacob-carlborg: Using `rdmd` would require to invocations of `dmd`. Not sure what's best.

configure
jacob-carlborg: Yes, these are the default paths it will look for. There's also a flag to explicitly specify which LLVM to use. I guess the script could iterate `/usr/lib/` and look for `llvm-x.x`. Seems a bit more complex though.

configure
jacob-carlborg: > At least on Arch Linux ncurses package only provides libcurses.so

That's unfortunate. Hmm, perhaps it's better to make dynamic linking the default. Then I can explicitly specify static linking for the binaries that are released.

> btw why does dstep need ncurses at all?

It's a dependency of libclang. Don't know what it's used for.

configure
jacob-carlborg: Yes, I'll add that.

configure
jacob-carlborg: Yeah, I'm not a big fan of that. I'll see if I change.

jacob-carlborg:
> Ah, also I would recommend to generate not only linker flag file but also D module which sets enum ClangVersion = "3.9" or something like that, so that dstep code can branch accordingly.

I was thinking of that as well, but let's start simple.

configure
Dicebot: Not with the latest rdmd - https://github.com/dlang/tools/pull/194

jacob-carlborg: Ah, interesting. Cool, I'll switch to rdmd then.

jacob-carlborg: I tried that and it doesn't work without a file extension:

```
Error: module configure is in file 'configure.d' which cannot be read
```

Dicebot: Well, you can call it `configure.d` :) I remember there was a long debate over allowing `rdmd` scripts  without file extension but seems it won't happen

jacob-carlborg: Sounds like Andrei is on board. Someone recently had the same issue [1].

[1] http://forum.dlang.org/post/o5sd4e$14n1$1@digitalmars.com

mihails-strasuns-sociomantic:
Is there anything I can do to speed up progress on this functionality?

jacob-carlborg:
> Is there anything I can do to speed up progress on this functionality?

Yeah, that's a good question. The biggest reason this has not progressed is that haven't had time to work on basically anything D related lately. Except from doing the actual work I'm not entirely sure, I would need to have a closer look to see what's missing.

A minor thing would be to collect some default paths for different versions of libclang on different Linux distributions and FreeBSD, which I haven't even started testing on.

There are a few things that I could use some feedback on:

* To support static linking libclang we need to compile libclang our self, since the pre compiled package is missing a static library. I started creating a Docker image for this to use as a cache, and also avoid being tied to whatever Linux distribution and version Travis CI is using. But now I'm not sure if this is the best approach and perhaps overkill. Which leads to the next question

* An alternative to the Docker image would be to use caching in Travis CI, which I'm pretty sure it supports. This would be used to avoid having to rebuild libclang every time. Figuring out how the Travis CI caching works would be helpful

jacob-carlborg:
Another thing I need to try is if the `$DC` environment variable is properly propagated from the main Dub file to the configuration script.

mihails-strasuns-sociomantic:
Would it be acceptable to separate static linking bits from `./configure` part? I have schedule some time to look into dstep next week but pursuing static libclang linking is quite a lot of effort for zero benefit to us, so I am not very excited about doing it :(

jacob-carlborg:
That would be acceptable, as long as it's not making it more difficult to implement later.

mihails-strasuns-sociomantic:
Apologize, I was sick for a while thus getting back to this with a delay. Looking into current state of things now.

mihails-strasuns-sociomantic:
@jacob-carlborg one decision I don't currently understand is that you use `llvm-config --libdir` currently and build linker flags manually instead of relying on `llvm-config --ldflags` with some fallback. That will fail on platforms that don't use plain `libclang.so` as library name. Why complicate things here?

jacob-carlborg:
`llvm-config` seems quite broken. The actual flag to link the necessary libraries are not included when invoking `llvm-config --ldflags`. `llvm-config --libs` only includes the LLVM libraries and not the Clang libraries.

jacob-carlborg:
@mihails-strasuns-sociomantic I just added documentation for everything in the config script and some other fixes.

mihails-strasuns-sociomantic:
Thank you very much!

jacob-carlborg:
@mihails-strasuns-sociomantic I've cleaned up the commits now as well. I *think* the only thing that is left is integrate this with the test runner.

mihails-strasuns-sociomantic:
I have been thinking about an elegant way to fix Mac build and make it more robust in general but it feels like existing system relies on `llvm-config` too much :( It probably makes sense considering your intention to do full static build starting from llvm/clang sources but for a regular dynamic linking approach it becomes a liability as `llvm-config` should not be even present for plain `libclang` installation. Instead, it would make more sense to just look through bunch of OS-specific paths for `libclang*.so` and use that.

I can write down a simple configure script for that of course but it seems in conflict with your main goal.

jacob-carlborg:
> I have been thinking about an elegant way to fix Mac build and make it more robust in general

Currently I'm more concerned of why the tests are passing on Linux at all. The reason why the tests don't work on macOS is because there is no LLVM/libclang/llvm-config in any of the default paths the config script is looking in. The tests need to be updated to invoke `./configure --lvm-config <path>` for each version of Clang that is tested and rebuild DStep.

>  but it feels like existing system relies on llvm-config too much

Currently the only thing that it needs from `llvm-config` is the library path. The config script can be changed to look for that path directly. Also change the flag `--lvm-config` to something like `--lib-path`.

Although, I have been thinking of using it to extract the system dependencies that are currently hardcoded. But if we don't do that, then it's only the library path we need.

mihails-strasuns-sociomantic:
I have a feeling I have lost the direction of your intentions. The way I see it, right now there are some interrelated bits put in `test.d`, `configure.d` and travis CI script and testing complexity explodes because of it.

If that was my call, I'd remove all code responsible for downloading clang versions from `test.d` and put it to `configure.d` making it single script responsible for setting up required environment. It would support just two scenarios:

1) system-wide libraries with dynamic linking - no need to care about `llvm-config`, just check library paths for `libclang*.so` and link to it
2) downloaded libraries with static linking - more or less what you have been going at

After that you can use travis build matrix to build and test dstep with different clang versions using second mode while each individual run will care only about one version. I don't think trying to make configuration generic helps.

If that sounds acceptable, I shall try create an alternative PR implementing it.

jacob-carlborg:
> I have a feeling I have lost the direction of your intentions. The way I see it, right now there are some interrelated bits put in test.d, configure.d and travis CI script and testing complexity explodes because of it.

What's in `test.d` is a bit of a mess now since I pushed unfinished work (backup and sharing across computers). The other parts are under control.

> If that was my call, I'd remove all code responsible for downloading clang versions from test.d and put it to configure.d making it single script responsible for setting up required environment

No, no, no. `configure.d` should not worry about downloading anything. The downloading is only for testing purpose. Otherwise, if the user wants to compile DStep, I expect the user to make sure that libclang is available on the computer, either using system-wide installation or manually downloaded archive.

> system-wide libraries with dynamic linking - no need to care about llvm-config, just check       library paths for libclang*.so and link to it 

This is what the script is doing now, `llvm-config` is gone.

> downloaded libraries with static linking - more or less what you have been going at

This is currently out of scope, to be able to get this PR finished sooner rather than later. I've removed the flags to configure for static linking and dynamic linking is now default. Internally the code is still there for static linking. It does work on macOS, but for some reason not on Linux.

> After that you can use travis build matrix to build and test dstep with different clang versions using second mode while each individual run will care only about one version

This is something I have wanted to do for a quite a while now and is the right direction. I've just never prioritized it.

> If that sounds acceptable, I shall try create an alternative PR implementing it.

I think the configuration script is fine as it is, it just needs to integrated in the test script. I was hoping that it was easier to just integrate the configuration script in the existing test script. In theory it should be enough to just replace the copy operation of libclang by an invocation of the configuration script and rebuild of DStep.

jacob-carlborg:
@mihails-strasuns-sociomantic I've updated the test script and everything is passing now. Is there anything else before I merge?

dub.json
```diff
@@ -21,17 +25,33 @@
         },
     },
 
+    "subPackages": [
+        {
+            "name": "configure",
+            "targetType": "executable",
```
mihails-strasuns-sociomantic: Is there any practical reason to also expose configure script as sub-package? That feels a bit semantically wrong to me as configure script is used as part of build process of other configurations, it is hardly a thing on its own.

dub.json
```diff
@@ -21,17 +25,33 @@
...
     "configurations": [
         {
             "name": "default",
             "mainSourceFile": "dstep/main.d",
             "targetName": "dstep",
             "sourcePaths": ["dstep", "clang"],
             "importPaths": ["dstep", "clang"],
+
+            "preBuildCommands-posix": [
+                "(! [ -s config.txt ] && ./configure) || true"
```
mihails-strasuns-sociomantic: The way it is currently used it makes more sense to call it something like `linker_flags.txt` as format is not reusable for any other kind of configuration.

configure.d
```diff
@@ -0,0 +1,653 @@
...
+
+        enum standardPaths = [
+            "/usr/lib",
+            "/usr/local/lib",
+            "/usr/lib/x86_64-linux-gnu" // Debian
+        ];
+
+        immutable llvmLibPaths = [
+            "/usr/lib/llvm-4.0/lib", // Debian
+            "/usr/lib/llvm-3.9/lib", // Debian
```
mihails-strasuns-sociomantic: Last supported clang version is 3.8 right now, isn't it?

configure.d
```diff
@@ -0,0 +1,653 @@
...
+     * Returns: the full path of `fileToCheck` by searching in `standardPaths`
+     *
+     * Throws: an `Exception` if `fileToCheck` cannot be found in any of the
+     *  paths in `standardPath`
+     */
+    string standardPath()
+    {
+        auto errorMessage = format("Could not find %s in any of the standard " ~
+            "paths for %s: \n%s\nPlease specify a path manually using " ~
+            "'./configure --%s-path=<path>'.",
```
mihails-strasuns-sociomantic: This prints `Please specify a path manually using './configure --llvm-path=<path>'` which is wrong (correct flag is `--llvm-lib-path`)

jacob-carlborg:
> That is probably only real request - would be reasonable to expect things just to compile out of the box with last currently supported libclang on ubuntu.

That's true. I had planed to bump the minimum requirement to 3.9 but that requires another PR so it's better to wait with that.

configure.d
jacob-carlborg: Yes. Well, 3.7 on Posix and 3.8 on Windows.

dub.json
jacob-carlborg: Fair enough.

configure.d
jacob-carlborg: Yes, I'll fix that.

mihails-strasuns-sociomantic:
> That's true. I had planed to bump the minimum requirement to 3.9 but that requires another PR so it's better to wait with that.

Yeah, enhancing configure script to generate D module with clang version so that 3.9 issue can be fixed is obvious next step but better to do it in a separate PR.

dub.json
jacob-carlborg: Is there a better way to do it?

mihails-strasuns-sociomantic: What is bad about plain `./configure` which you use anyway?

jacob-carlborg: It's a shell script that executes the sub package.

mihails-strasuns-sociomantic: Oh, I have missed that. I was `rdmd` script before, wasn't it? Why have you changed that? Alternatively you can move the package definition to `configure.d` itself:

```
#!/usr/bin/env dub
/+ dub.json:
{
    "name": "configure",
    ...
+/

(D code follows)
```

mihails-strasuns-sociomantic: (it will become runnable via `./configure.d` that way)

jacob-carlborg: > Oh, I have missed that. I was rdmd script before, wasn't it? Why have you changed that?

To let Dub automatically handle identifying the compiler, i.e. support both DMD and LDC.

> Alternatively you can move the package definition to configure.d itself

I did that first but run into this problem: https://github.com/dlang/dub/issues/910. Then I though what's the point of having it defined there if I cannot use `./configure` anyway. But as you mention now, it's better to avoid exposing the sub package at all.

jacob-carlborg:
> enhancing configure script to generate D module with clang version so that 3.9 issue can be fixed

Actually, if we bump the minimum requirement to 3.9 we don't need to do that. It's only necessary to support both 3.9+ and 3.9-. But if we only support 3.9+ it shouldn't be necessary.

dub.json
jacob-carlborg: Fixed.

dub.json
jacob-carlborg: Fixed.

configure.d
jacob-carlborg: Fixed.

configure.d
jacob-carlborg: Fixed.

dub.json
mihails-strasuns-sociomantic: Ah, nasty. And I presume you don't want to call it directly like `./configure.d` to keep it more in style with configure-make pattern?

jacob-carlborg: Yes, exactly.

MartinNowak:
Hope you do know that this obviously does not work with `dub fetch` and `dub build`. No git repository is cloned, hence you need some sort of fallback.
Also the path is not changed when building a dub package out of tree, e.g. `--root=path/to/pkg` or `dub build dstep`, so your resources/VERSION file will land in the CWD.

jacob-carlborg:
I did not think of that.

jacob-carlborg:
Pull requests are welcome 😉.

PR #108, opened 2017-01-15
Fix #107: Handle typedef of opaque structs.

### LINKED ISSUE #107: Handle typedef of opaque structs
When an opaque struct is defined in a C header like that:
```c
    typedef struct foo foo;
```
no symbols are generated in the .d file. It only works if the name of the struct is not the same as the typedef. The output should be:
```d
   struct foo;
```

ciechowoj:
As in the header.

### DIFF
--- a/clang/Cursor.d
+++ b/clang/Cursor.d
@@ -93,6 +93,23 @@ struct Cursor
         return Type(clang_getTypedefDeclUnderlyingType(cx));
     }
 
+    @property Cursor underlyingCursor() const
+    {
+        foreach (child; all)
+        {
+            if (child.kind == CXCursorKind.CXCursor_TypeRef)
+                return child.referenced;
+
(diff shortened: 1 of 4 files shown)


jacob-carlborg:
👌

jacob-carlborg:
Thanks.

ciechowoj:
Eh, it seems that my fix the issue, but adds adds another regression. I had forgotten that there is no struct forward declarations in D, so that following code will no compile:

```
struct A;
struct A { int b; }
```

I think the correct implementation should omit the first declaration if the second one is present. I'll fix when I have some spare time.

jacob-carlborg:
Should I revert?

ciechowoj:
Maybe wait a moment, let say 4h from now, I'm trying to fix it.

jacob-carlborg:
Ok, no problem.

PR #109, opened 2017-01-15
Fix #106: Handle complex floating-point types.

### LINKED ISSUE #106: Handle complex floating-point types
Built-in C support for complex numbers, e.g. `float _Complex` or `double _Complex`, is currently not handled by dstep.

Support looks like it should be reasonably straightforward to implement.  Complex numbers are defined by an enum value in `clang.c.Index`:
https://github.com/jacob-carlborg/dstep/blob/561545d48cdc9ab7fe71efc359df456c138e126e/clang/c/Index.d#L2838

and so adding support would probably need to involve adding another special-case to the `switch` statement in `dstep.Type.translateType`:
https://github.com/jacob-carlborg/dstep/blob/561545d48cdc9ab7fe71efc359df456c138e126e/dstep/translator/Type.d#L55

Note, a special-case `translateComplex` would be needed because it will be necessary to extract the base floating-point type of the complex number(s) concerned.

ciechowoj:
As in the title.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -21,6 +21,7 @@
 * Add the `libclang` bindings as a test case
 * Add support for Microsoft Windows
 * Run Windows tests on AppVeyor
+* Handle complex floating-point types.
 
 ### Bugs Fixed
 
--- a/dstep/translator/Type.d
+++ b/dstep/translator/Type.d
@@ -91,6 +91,10 @@ body
                         rewriteIdToObjcObject);
                     break;
 
+                case CXType_Complex:
+                    result = translateComplex(type);
+                    break;
+
                 default: result = translateType(context, type.kind, rewriteIdToObjcObject);
             }
     }
@@ -234,6 +238,17 @@ body
         return translateType(context, declaration, rewriteIdToObjcObject);
     else
         return translateType(context, type.kind, rewriteIdToObjcObject);
+}
+
+string translateComplex (Type type)
+{
+    switch (type.element.kind)
+    {
+        case CXTypeKind.CXType_Float: return "cfloat";
+        case CXTypeKind.CXType_Double: return "cdouble";
+        case CXTypeKind.CXType_LongDouble: return "creal";
+        default: return "<unimplemented>";
+    }
 }
 
 string translateArrayElement(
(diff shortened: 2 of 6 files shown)


jacob-carlborg:
Are there any literals for Complex in C?

ciechowoj:
It doesn't seem so, however there are constants for imaginary unit `_Imaginary_I`, `_Complex_I`,` I`.

jacob-carlborg:
I googled this and found code like `double complex z1 = 1.0 + 3.0 * I` [1]. Will that work in D?

[1] http://stackoverflow.com/questions/6418807/how-to-work-with-complex-numbers-in-c

ciechowoj:
Not without changes, however dstep doesn't translate complex (complicated) expressions at the moment anyway. E.g.

```
assertTranslates(
q"C
static const int x = 1 * 2;

C",
q"D
extern (C):

extern __gshared const int x;
D");
```

No expression.

jacob-carlborg:
Right, that's not needed.

jacob-carlborg:
Could you please update the changelog as well.

dstep/translator/Type.d
```diff
@@ -236,6 +240,16 @@ body
         return translateType(context, type.kind, rewriteIdToObjcObject);
 }
 
+string translateComplex (Type type)
+{
+    switch (type.element.kind) {
```
jacob-carlborg: Curly braces on the next line.

changelog.md
```diff
@@ -40,6 +40,7 @@
 * [Issue 59](https://github.com/jacob-carlborg/dstep/issues/59): Shouldn't dstep exit with status code when there is some kind of error
 * [Issue 83](https://github.com/jacob-carlborg/dstep/issues/83): New multiline translation
 * [Issue 85](https://github.com/jacob-carlborg/dstep/issues/85): dstep not converting `const T x[]` to `const (T)* x`
+* [Issue 106](https://github.com/jacob-carlborg/dstep/issues/106): Handle complex floating-point types.
```
jacob-carlborg: This is a new feature, it should be under `New/Changed Features`.

ciechowoj: Done.

jacob-carlborg:
👌

jacob-carlborg:
Thanks.

PR #112, opened 2017-01-17
Support for custom global attributes

### LINKED ISSUE #111: Add nothrow and @nogc automatically
There could be a way to automatically add `nothrow` and `@nogc` at the beginning of the output:
```d
extern (C) nothrow @nogc:
```
I don't know if it should be the default or set through configuration switches.

biozic:
Fixes #111

### DIFF
--- a/unit_tests/UnitTests.d
+++ b/unit_tests/UnitTests.d
@@ -942,3 +942,21 @@ alias foo = int function(int function() bar);
 D", options);
 
 }
+
+// Test global attributes
+unittest
+{
+    Options options;
+    options.globalAttributes = ["nothrow", "@nogc"];
+
+    assertTranslates(q"C
+void foo();
+C", q"D
+extern (C):
+nothrow:
+@nogc:
+
+void foo ();
+D", options);
+
+}
--- a/changelog.md
+++ b/changelog.md
@@ -23,6 +23,7 @@
 * Run Windows tests on AppVeyor
 * Handle complex floating-point types.
 * Use the new alias syntax (`alias Y = X`) in the output
+* Support for custom global attributes (e.g. `nothrow`, `@nogc`)
 
 ### Bugs Fixed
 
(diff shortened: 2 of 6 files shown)


unit_tests/UnitTests.d
```diff
@@ -905,3 +905,21 @@ void foo ();
...
+{
+    Options options;
+    options.globalAttributes = ["nothrow", "@nogc"];
+
+    assertTranslates(q"C
+void foo();
+C", q"D
+extern (C):
+nothrow:
+@nogc:
```
ciechowoj: Wouldn't it be better to have a syntax from the issue #111 , you've originally proposed?

```
extern (C) nothrow @nogc:
```

biozic: In this case, yes. But then I thought it could also be used for other attributes like `package(...)`, or even `version(...)`, and that would make it less readable, but this might also be feature overuse...

I don't really care. I'm ok to change it.

jacob-carlborg: druntime seems to use one attribute per line in their C bindings.

biozic:
Updated with latest changes.

jacob-carlborg:
Apparently the Windows 64 builds are failing in master but are not properly reported as such. I would like to avoid accepting any pull requests until that is fixed.

biozic:
OK. I'll keep the PR up to date if required.

jacob-carlborg:
We've fixed the Windows 64 builds now, please rebase.

biozic:
Rebased. Still no luck with CI : Travis builds fail now.

jacob-carlborg:
I just pushed a fix for master. The failing build was unrelated to the tests/code. It was `git describe` that failed due to Travis CI's shallow clone of the repository.

biozic:
OK, it's working now.

jacob-carlborg:
LGTM 👌

jacob-carlborg:
If you prefix the commit message with "Fix #111: ..." it will automatically close the issue.

biozic:
Done.

jacob-carlborg:
Thanks.

PR #113, opened 2017-01-19
Updated alias syntax

biozic:
The same alias syntax (the new one) is now used everywhere.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -22,6 +22,7 @@
 * Add support for Microsoft Windows
 * Run Windows tests on AppVeyor
 * Handle complex floating-point types.
+* Use the new alias syntax (`alias Y = X`) in the output
 
 ### Bugs Fixed
 
--- a/dstep/translator/Translator.d
+++ b/dstep/translator/Translator.d
@@ -280,9 +280,9 @@ class Translator
                 underlying.spelling != ""))
             {
                 output.singleLine(
-                    "alias %s %s;",
-                    translateType(context, typedef_, typedef_.type.canonical),
-                    typedef_.spelling);
+                    "alias %s = %s;",
+                    typedef_.spelling,
+                    translateType(context, typedef_, typedef_.type.canonical));
 
                 context.markAsDefined(typedef_);
             }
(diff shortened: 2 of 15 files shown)


jacob-carlborg:
Could you please slightly improve the commit message.

jacob-carlborg:
LGTM. @Dicebot @ciechowoj what do you think?

biozic:
Is "Use new alias syntax everywhere" better?

jacob-carlborg:
> Is "Use new alias syntax everywhere" better?

Yes, but perhaps even better, something like: "Use new alias syntax for output result". I would like to emphasize that new alias syntax is used in the translated code and not the DStep code itself.

ciechowoj:
> LGTM. @Dicebot @ciechowoj what do you think?

Yeah, I always wanted to change this.

biozic:
@Dicebot I wasn't aware that D1 was supposed to be supported. Does this also include [enum aliases](https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Enum.d#L42), which already used the new syntax?

jacob-carlborg:
I know there were some issues with the new alias syntax and `extern (C)`, something like:

```d
extern (C) alias Callback = void function();
```

Has that been fixed? Or perhaps it doesn't matter when `extern (C):` will be at the top?

biozic:
All of these seem to work (2.070.2):
```d
alias Callback = extern (C) void function();
extern (C) alias Callback = void function();
alias extern (C) void function() Callback;
extern (C) alias void function() Callback;
```

jacob-carlborg:
@biozic could you please update the changelog as well.

jacob-carlborg:
👌

jacob-carlborg:
Thanks. Next time, write a comment after a change. I don't get notification just for changes.

jacob-carlborg:
@biozic I just merged this and now the tests are failing: https://travis-ci.org/jacob-carlborg/dstep/jobs/201110273

ciechowoj:
Yes, I've just noticed it also. What is more AppVeyor reports that the tests are passing despite the fact they are failing.

biozic:
The failing tests are tests that were added after my PR, so I guess the CI is not faulty. I will submit a PR to fix the newly-added failing tests.

jacob-carlborg:
Fair enough. Waiting for the tests in the new PR.

Issue #117, opened 2017-01-25
fatal error: 'limits.h' file not found


mihails-strasuns-sociomantic:
With latest dstep master trying to convert https://github.com/protobuf-c/protobuf-c/blob/master/protobuf-c/protobuf-c.h results in this:

```
/usr/include/limits.h:123:16: fatal error: 'limits.h' file not found
```

jacob-carlborg:
Is it the embedded `limits.h` it can't find?

Dicebot:
Seems so, but I am not sure how to debug the lookup. I have noticed `limits.h` is missing in this list https://github.com/jacob-carlborg/dstep/blob/master/clang/Compiler.d#L38 and trying adding it but with no success.

jacob-carlborg:
Hmm, it definitely should be in that list.

jacob-carlborg:
Which version of libclang are you using? How does your `/usr/include/limits.h` look like? I cannot reproduce this on macOS, but I'm pretty sure that `/usr/include/limits.h` is different, I assume you're on Arch Linux?

Dicebot:
Checked on both Arch Linux and Ubuntu, clang versions 3.6 and 3.8

Relevant lines:

```
/* Get the compiler's limits.h, which defines almost all the ISO constants.

    We put this #include_next outside the double inclusion check because
    it should be possible to include this file more than once and still get
    the definitions from gcc's header.  */
#if defined __GNUC__ && !defined _GCC_LIMITS_H_
/* `_GCC_LIMITS_H_' is what GCC's file defines.  */
# include_next <limits.h>
#endif
```

The way I understand it, problem comes from the fact it still looks in `/usr/include/` for `limits.h` despite it is compiled in resources. And finds the one from GCC which causes the trouble.

jacob-carlborg:
I think the issue is the `include_next`. I slightly modified DStep to include the `limits.h` and the virtual include path. My include paths look like this:

```
#include "..." search starts here:
#include <...> search starts here:
 /4421172 # virtual include path added by DStep
 /usr/local/include
 /usr/include
 /System/Library/Frameworks (framework directory)
 /Library/Frameworks (framework directory)
```

It finds `limits.h` in `/usr/include` then `include_next` tells the compiler to look for the header file in the next include path, which is supposed to be the compiler built-in. Although I don't know why it's doesn't find it in the virtual include path `/4421172`.

jacob-carlborg:
[...]

PR #118, opened 2017-01-26
Various bug fixes / tweaks

Dicebot:


dstep/main.d
```diff
@@ -59,8 +59,9 @@ auto parseCLI (string[] args)
         std.getopt.config.passThrough,
         std.getopt.config.caseSensitive,
         "output|o", &config.output,
-        "language|x", &parseLanguage,
         "objective-c", &forceObjectiveC,
+        "language|x", "Treat subsequent input files as having type <language>.",
```
ciechowoj: Shouldn't the help snippet be in `Configuration.d`?

dstep/Configuration.d
```diff
@@ -21,7 +21,6 @@ struct Configuration
     string[] inputFiles;
 
     /// expected programming language of input files
-    @("language|x", "Treat subsequent input files as having type <language>.")
```
jacob-carlborg: I don't understand this commit.

dstep/main.d
Dicebot: It was there, but that means `language` entry gets defined twice which is what causes assertion with DMD 2.073

dstep/Configuration.d
Dicebot: See https://github.com/jacob-carlborg/dstep/pull/118#discussion_r97991745

dstep/main.d
jacob-carlborg: Do you know it's only a problem for the `language` flag?

ciechowoj: Hmm, I think it's ok as a temporary fix, but I would prefer it to be in `Configuration.d`, however I don't see simple solution to do that right now...

ciechowoj: Actually, you could add the help snippet to `customEntries` at `main.cpp:226` instead.

Dicebot: @jacob-carlborg 

> Do you know why it's only a problem for the language flag?

Others are not duplicated (exist either in main list, or auto-generated one)

jacob-carlborg: Aha, I see. Would it be better to move it down to custom entries: https://github.com/Dicebot/dstep/blob/9a720e5e3a20ec4ad6a61437525d96d256892fd8/dstep/main.d#L227. Otherwise this will be spread across places.

dstep/translator/MacroDefinition.d
```diff
@@ -1967,8 +1967,13 @@ void translateConstDirective(
 {
     Set!string params, imports;
 
+    version (D1)
+        string fmt = "const %s = %s;";
+    else
+        string fmt = "enum %s = %s;";
```
jacob-carlborg: Perhaps immutable or enum.

jacob-carlborg:
* What are the test files testing?
* Can the `limits.h` change be tested somehow? A header file which includes it should be enough?

ciechowoj:
We would need to implement translation of well-known preprocessor macros to handle it. In the particular case of `INT_MAX` - `int.max`.

ciechowoj:
No, you need to add something like:

```
    assertTranslatesObjCFile(
        "test_files/objc/classes.d",
        "test_files/objc/classes.h");
```

ciechowoj:
I think it would be good idea to make it automatic process, though.

ciechowoj:
I'm trying to fix this. And I've a problem: Is there a reliable way to get a path to `resources` directory? May I assume that it has fixed location with respect to the directory that contains dstep executable (`../resources`)?

@jacob-carlborg , @Dicebot

ciechowoj:
Hmm, after some more thinking my question is probably no more relevant.

jacob-carlborg:
@Dicebot 18 commits and 35 change files. Is that intentional?

jacob-carlborg:
@leandro-lucarella-sociomantic 18 commits and 35 changed files. Is that intentional?

jacob-carlborg:
@ciechowoj the `resources` directory is only used during compilation. What do you need it for?

jacob-carlborg:
@Dicebot about the blocked issue. Wouldn't it be enough to include `limits.h` as a test?

ciechowoj:
> @ciechowoj the resources directory is only used during compilation. What do you need it for?

Yeah, I've just realized that.

ciechowoj:
What about leaving the translated code as `enum TEST = INT_MAX;` and creating a new issue with the problem?

jacob-carlborg:
Hmm, is a symbol is used from the `limits.h` header file, then a corresponding import should be added in the D module [1]. Not sure if this isn't working because the symbol in this case is a preprocessor symbol.

[1] https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/IncludeHandler.d#L36

jacob-carlborg:
@Dicebot If we cannot fix this now, could you please add a comment in the C file used in the test, indicating somehow, that the translation need to be fixed at some point.

jacob-carlborg:
@Dicebot looking good. Could you please update the changelog as well, for #117.

jacob-carlborg:
Thanks. 👌

PR #119, opened 2017-01-26
Fix #116: Option --space-after-function-name doesn't work with function pointer syntax.

### LINKED ISSUE #116: Option --space-after-function-name doesn't work with function pointer syntax
I think it's a recent change.
```c
int foo(void);

typedef int (*Fun0)(void);
typedef int (*Fun1)(int (*param)(void));

struct Foo {
  int (*bar)(void);
};
```
when run with `--space-after-function-name=false`, gives:
```d
int foo();

alias Fun0 = int function ();
alias Fun1 = int function (int function () param);

struct Foo
{
    int function () bar;
}
```

ciechowoj:
As in the title.

### DIFF
--- a/unit_tests/UnitTests.d
+++ b/unit_tests/UnitTests.d
@@ -905,3 +905,34 @@ void foo ();
 D", options));
 
 }
+
+// Put or don't put a space after the 'function' keyword when
+// --space-after-function-name is or isn't specified.
+unittest
+{
+    Options options;
+    options.spaceAfterFunctionName = true;
+
+    assertTranslates(
+q"C
+typedef int (*foo)(int (*bar)(void));
+C",
+q"D
+extern (C):
+
+alias foo = int function (int function () bar);
+D", options);
+
+    options.spaceAfterFunctionName = false;
+
+    assertTranslates(
+q"C
+typedef int (*foo)(int (*bar)(void));
+C",
+q"D
+extern (C):
+
+alias foo = int function(int function() bar);
+D", options);
+
+}
--- a/unit_tests/issues/Issue116.d
+++ b/unit_tests/issues/Issue116.d
@@ -1,0 +1,43 @@
+/**
+ * Copyright: Copyright (c) 2016 Wojciech Szęszoł. All rights reserved.
+ * Authors: Wojciech Szęszoł
+ * Version: Initial created: Jan 26, 2017
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+
+import Common;
+import dstep.translator.Options;
+
+// Fix 116: Option --space-after-function-name doesn't work with function
+// pointer syntax.
+unittest
+{
+    Options options;
+    options.spaceAfterFunctionName = false;
+
+    assertTranslates(
+q"C
+int foo(void);
+
+typedef int (*Fun0)(void);
+typedef int (*Fun1)(int (*param)(void));
+
+struct Foo {
+  int (*bar)(void);
+};
+C",
+q"D
+extern (C):
+
+int foo();
+
+alias Fun0 = int function();
+alias Fun1 = int function(int function() param);
+
+struct Foo
+{
+    int function() bar;
+}
+D", options);
+
+}
--- a/changelog.md
+++ b/changelog.md
@@ -44,6 +44,7 @@
 * [Issue 85](https://github.com/jacob-carlborg/dstep/issues/85): dstep not converting `const T x[]` to `const (T)* x`
 * [Issue 107](https://github.com/jacob-carlborg/dstep/issues/107): Handle typedef of opaque structs.
 * [Issue 114](https://github.com/jacob-carlborg/dstep/issues/114): Crash on recursive typedef.
+* [Issue 116](https://github.com/jacob-carlborg/dstep/issues/116): Option --space-after-function-name doesn't work with function pointer syntax.
 * [Issue 117](https://github.com/jacob-carlborg/dstep/issues/117): fatal error: 'limits.h' file not found.
 
 ## Version 0.2.1
--- a/dstep/translator/Type.d
+++ b/dstep/translator/Type.d
@@ -430,7 +430,10 @@ string translateFunctionPointerType (Context context, Cursor cursor, FuncType fu
     auto result = translateType(context, cursor, func.resultType);
 
     Output output = new Output();
-    translateFunction(output, result, "function", params, func.isVariadic);
+    auto spacer = context.options.spaceAfterFunctionName ? " " : "";
+    translateFunction(output, result, "function",
+                      params, func.isVariadic, "", spacer);
+
     return output.data();
 }
 


jacob-carlborg:
Update the changelog please.

ciechowoj:
Done. However some tests don't pass due to [this](https://github.com/jacob-carlborg/dstep/pull/118/files#diff-0c0cbcae56297a371672c36f64de91f3R101), so this PR will have to wait until #118 is merged.

ciechowoj:
@jacob-carlborg  Rebased and ready to merge.

unit_tests/UnitTests.d
```diff
@@ -905,3 +905,34 @@ void foo ();
 D", options));
 
 }
+
+// Put or put not space after 'function' keyword when
```
jacob-carlborg: `put not` -> `don't put`.

unit_tests/issues/Issue116.d
```diff
@@ -0,0 +1,43 @@
...
+
+import Common;
+import dstep.translator.Options;
+
+// Fix 116: Option --space-after-function-name doesn't work with function
+// pointer syntax.
+unittest
+{
+    Options options;
+  	options.spaceAfterFunctionName = false;
```
jacob-carlborg: Indentation is not correct.

unit_tests/UnitTests.d
ciechowoj: "Do or do not, there is no try." ~ Yoda : D

ciechowoj: Fixed, anyway.

unit_tests/issues/Issue116.d
ciechowoj: Fixed.

jacob-carlborg:
master is failing, so I'm going to wait with accepting this PR.

ciechowoj:
Sure.

ciechowoj:
I've added here an another commit that is supposed to fix the problem with AppVeyor, I've just mentioned [here](https://github.com/jacob-carlborg/dstep/pull/113#issuecomment-279396135).

jacob-carlborg:
I'm running the tests again.

ciechowoj:
There is something wrong with AppVeyor x86_64 build. I think it has not been passing since quite a long, however it haven't been reported. Check this out [link](https://ci.appveyor.com/project/jacob-carlborg/dstep/build/1.0.169/job/bh60xfdnc8suwdt5). The CI is green despite the fact there were errors.

jacob-carlborg:
Yeah. This seems to be the first time it started to fail: https://ci.appveyor.com/project/jacob-carlborg/dstep/build/1.0.85/job/v1xcjkuk22e5oe83.

Is it necessary to manually set the exist code to have the test fail? Seems pretty bad.

ciechowoj:
> Is it necessary to manually set the exist code to have the test fail? Seems pretty bad.

I see this as mapping of one error codes to some other error codes. But I'm not sure, why it doesn't work. I'll investigate it, when I have some spare time.

jacob-carlborg:
> I see this as mapping of one error codes to some other error codes. But I'm not sure, why it doesn't work. Maybe there is a better solution. I'll investigate it, when I have some spare time.

As long as it properly fails the test I'm satisfied. But now we need to figure out why the 64bit is failing.

ciechowoj:
Ok.

ciechowoj:
I've tried to fix that windows build and there are some problems. The main reason the x64 tests was a single unit test I commented it out because I haven't figured out how to fix it yet. 

However there were issues with `time_t` and adding the `std.time` import. I do not know what had made the handling of time_t to be broken, but handling of includes and imports needs an improvement.

jacob-carlborg:
> I've tried to fix that windows build and there are some problems. The main reason the x64 tests were failing was a single unit test. I commented it out because I haven't figured out how to fix it yet. There is something wrong with throwing an exception in throwCollisionError in Context.d:207...

Ok. Do you expect it to be a logical error, or could it be some kind of bug with the exception handling in the compiler?

How do you recommend we proceed, just comment out that test for now and move on, or? 

> What is more, there were issues with time_t and adding the std.time import. I do not know what had made the handling of time_t to be broken, but handling of includes and imports needs an improvement. I've added a quick and dirty fix but it needs to be changed

I'm not sure if this is the same issue but I've noticed (at least in the header files on macOS) that a lot of the symbols from the standard library (or Posix) are not defined in the headers the documentation says to include. Rather they're defined in some platform/system specific files, which are then included in the header files the documentation says to include.

ciechowoj:
> Ok. Do you expect it to be a logical error, or could it be some kind of bug with the exception handling in the compiler?

Currently, almost no idea. It seems to work with 3.7 and break with 3.8, maybe some .dll incompatibility?

> How do you recommend we proceed, just comment out that test for now and move on, or?

I do not have more spare time to spend on it at the moment. The unit test in question tests the `--collision-action=abort" feature which isn't that critical. I would leave the test commented and open a corresponding issue on github. 

I am going to fix it eventually...

> I'm not sure if this is the same issue but I've noticed (at least in the header files on macOS) that a lot of the symbols from the standard library (or Posix) are not defined in the headers the documentation says to include. Rather they're defined in some platform/system specific files, which are then included in the header files the documentation says to include.

Yeah, that is the exact source of problem.

jacob-carlborg:
> Yeah, that is the exact source of problem.

What if we extend the list of include mappings to also include the name of the type/symbol. Something like:

```d
auto knownIncludes = [Foo("time_t", `sys\types.h`) : "core.stdc.time"];
```

Need to come up with a better name for `Foo` of course.

ciechowoj:
> What if we extend the list of include mappings to also include the name of the type/symbol. Something like:
> Need to come up with a better name for Foo of course.

Yes, that seems like a viable solution for the time being .

jacob-carlborg:
@ciechowoj is there a possibility that you could extract the fix for the failing Windows tests as a separate pull requests? I've been trying for the last couple of days to setup an environment for compiling DStep on Windows but have failed.

ciechowoj:
> @ciechowoj is there a possibility that you could extract the fix for the failing Windows tests as a separate pull requests? 

Yeah, I'll do it tomorrow, if I don't forget. 

> I've been trying for the last couple of days to setup an environment for compiling DStep on Windows but have failed so far.

Hmm, it is quite straightforward for x86_64, at least I created the setup that can compile dstep within 1h on virtual machine. You need git, dmd, dub and LLVM, all of them have windows installers...

jacob-carlborg:
> Hmm, it is quite straightforward for x86_64, at least I created the setup that can compile dstep within 1h on virtual machine. You need git, dmd, dub and LLVM, all of them have windows installers...

Most of the issues have been with Visual Studio. Currently I'm running inside the Visual Studio shell (or whatever it's called) and DStep (libclang) cannot find the C standard library header files. I hvae basically no experience with developing on Windows.

jacob-carlborg:
> Yeah, I'll do it tomorrow, if I don't forget.

Thanks, no rush.

ciechowoj:
> Most of the issues have been with Visual Studio. Currently I'm running inside the Visual Studio shell (or whatever it's called) and DStep (libclang) cannot find the C standard library header files. I hvae basically no experience with developing on Windows.

Oh, I see. Truth say, I haven't made it to that state as well, I only made it to compile, so that I can find out why the tests fail on AppVeyor. They fail due to lack of standard headers on my local VirtualBox installation as well. AppVeyor seems to use headers from its MinGW installation. Soon, I should have access to the computer with normal Windows installation, so I will take a closer look at that issue.

jacob-carlborg:
@ciechowoj BTW, have you seen this [1]? It's possible to remotely connect to a failed build.

[1] https://www.appveyor.com/docs/how-to/rdp-to-build-worker/

ciechowoj:
No, I haven't. I could be useful, indeed.

ciechowoj:
@jacob-carlborg 

> > Yeah, I'll do it tomorrow, if I don't forget.
> Thanks, no rush.

Done.

jacob-carlborg:
👌

jacob-carlborg:
Thanks.

PR #124, opened 2017-02-13
Fix new unittests for #113

biozic:


jacob-carlborg:
Thanks for the quick fix. Seems like there was a backlog in Travis CI for the macOS jobs.

PR #125, opened 2017-02-20
Provide D1 fallback for new alias syntax

Dicebot:
Fixup for new syntax changes


jacob-carlborg:
Unless you've already noticed, the Windows 64 builds are failing in master but are not properly reported as such. I would like to avoid accepting any pull requests until that is fixed.

jacob-carlborg:
We've fixed the Windows 64 builds now, please rebase.

PR #130, opened 2017-03-15
Don't try aliasing D keywords

mihails-strasuns-sociomantic:
Most common issue this prevents is trying to alias `assert` keyword when
translating C macro.

### DIFF
--- a/dstep/translator/MacroDefinition.d
+++ b/dstep/translator/MacroDefinition.d
@@ -2053,6 +2053,18 @@ bool translateFunctAlias(
     {
         Identifier ident = cast(Identifier) expr.expr;
 
+        if (ident.spelling == "assert")
+        {
+            // aliasing `assert` from C "assert.h" is commong enough to warrant
+            // a special case to use function instead of an alias
+            output.singleLine("void %s(T...)(T args)", definition.spelling);
+            output.singleLine("    if (T.length <= 2)");
+            output.singleLine("{");
+            output.singleLine("    assert(args);");
+            output.singleLine("}");
+            return true;
+        }
+
         if (ident !is null &&
             equal(definition.params, expr.args.map!(a => a.translate(context, params, imports))))
         {
--- a/unit_tests/MacroTranslTests.d
+++ b/unit_tests/MacroTranslTests.d
@@ -261,6 +261,17 @@ extern (D) auto STRINGIZE(T0, T1)(auto ref T0 major, auto ref T1 minor)
 }
 D");
 
+    assertTranslates(q"C
+#define PROTOBUF_C_ASSERT(condition) assert(condition)
+C", q"D
+extern (C):
+
+void PROTOBUF_C_ASSERT(T...)(T args)
+    if (T.length <= 2)
+{
+    assert(args);
+}
+D");
 }
 
 // Translate member access operators.


mihails-strasuns-sociomantic:
Green

dstep/core/Traits.d
```diff
@@ -0,0 +1,30 @@
...
+        "private", "protected", "public", "pure", "real", "ref", "return",
+        "scope", "shared", "short", "static", "struct", "super", "switch",
+        "synchronized", "template", "this", "throw", "true", "try", "typedef",
+        "typeid", "typeof", "ubyte", "ucent", "uint", "ulong", "union",
+        "unittest", "ushort", "version", "void", "volatile", "wchar", "while",
+        "with"
+    ];
+ 
+    return DKeywords.canFind(ident);
+}
```
jacob-carlborg: There is already an existing function for this: https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Translator.d#L487

dstep/translator/MacroDefinition.d
```diff
@@ -2053,6 +2053,10 @@ bool translateFunctAlias(
     {
         Identifier ident = cast(Identifier) expr.expr;
 
+        import dstep.core.Traits : isKeyword;
+        if (ident.spelling.isKeyword())
+            return false;
```
jacob-carlborg: In other places in the code, when there's a symbol with is a D keyword that is translate, an underscore is added as a suffix. Which there is a separate function for [1].

[1] https://github.com/jacob-carlborg/dstep/blob/master/dstep/translator/Translator.d#L469

mihails-strasuns-sociomantic: This case is a bit different - you simply can't alias a keyword. For example `alias foo = assert` is not legal D code, no matter what you do.

jacob-carlborg: I guess that would end up as `alias foo = assert_`. But that might be a bit pointless.

mihails-strasuns-sociomantic: Not only pointless, it won't even compile because no one defines `assert_`.

jacob-carlborg: But with this change, it's not possible to handle when it _is_ defined. Take this example from test suite, slightly modified:

```c
#define delegate(major, minor)#major"."#minor
#define STRINGIZE(major, minor)delegate(major, minor)
```

Currently this is translated to this:

```d
extern (D) string delegate(T0, T1)(auto ref T0 major, auto ref T1 minor)
{
    import std.conv : to;

    return to!string(major) ~ "." ~ to!string(minor);
}

alias STRINGIZE = delegate;
```

Which is, of course, a bug. Instead it should be translated with a `_` suffix for identifier `delegate`:

```d
extern (D) string delegate_(T0, T1)(auto ref T0 major, auto ref T1 minor)
{
    import std.conv : to;

    return to!string(major) ~ "." ~ to!string(minor);
}

alias STRINGIZE = delegate_;
```

jacob-carlborg:
There are not tests.

dstep/translator/MacroDefinition.d
mihails-strasuns-sociomantic: I see, there seems to be a bit of a clash between your example and the one I have. To be more, specific, issue I am trying to solve is translation of this macro:

```c
#define PROTOBUF_C_ASSERT(condition) assert(condition)
```

Here `assert` resolves to function in `assert.h` in C header, but exiting translation will try turning it into `alias PROTOBUF_C_ASSERT = asser` which of course can't work.

jacob-carlborg: Looking at this problem in general, I don't see why adding a `_` suffix wouldn't work. Example:

```c
void delegate(bool condition);
#define PROTOBUF_C_DELEGATE(condition) delegate(condition)
```

I don't see why that couldn't be translated to:

```d
pragma(mangle, "delegate") void delegate_(bool condition);
alias PROTOBUF_C_DELEGATE = delegate_;
```

Your assumption in this PR is also that `assert` will always be the symbol defined in `assert.h`, which might not be the case. At least how the code is written now, it will ignore all symbols that are keywords in D, which might not be desirable.

In this particular case with `assert`, I don't think anyone would think it's useful to translate that function/macro to D (a corresponding `assert` module does not exist in druntime). If someone would manually translate some C code using `assert` they would most likely translate the use of `assert` to the D built-in `assert` and not translate the actual macro/function.

I'm not sure what's the best short term solution is. One idea would be to special case `assert` and implement the `assert` macro inline, or completely ignore it.

For a long term solution, I've been thinking for a while about a generic way to specify declarations that should be ignored.

mathias-lang-sociomantic:
Ping @mihails-strasuns-sociomantic

mihails-strasuns-sociomantic:
Updated. After some more thinking about the issue I agree that simply special casing "assert" alone to generate function instead of an alias is most practical approach.

Also added test case.

dstep/translator/MacroDefinition.d
```diff
@@ -2053,6 +2053,15 @@ bool translateFunctAlias(
     {
         Identifier ident = cast(Identifier) expr.expr;
 
+        if (ident.spelling == "assert")
```
jacob-carlborg: Is it worth checking that it's defined in `assert.h` as well?

ciechowoj: We would need to be able to determine if it's reachable 'through' the `assert.h` header...

jacob-carlborg: > Does libclang expose such information? It is not well-defined in C in general because of copy-paste nature of headers

Yes, at least for regular non-macro symbols. It's possible to get the location/path of a cursor [1], not sure if macros are represented as cursors. It's used extensively by the include handler.

[1] https://github.com/jacob-carlborg/dstep/blob/master/clang/Cursor.d#L68

jacob-carlborg: > We would need to be able to determine if it's reachable 'through' the assert.h header

@ciechowoj are you referring to these platform specific header files again? The true location of a symbol doesn't match the documentation/standard. I hate those.

ciechowoj: Yes, that's what I'm referring to.

mathias-lang-sociomantic: > Is it worth checking that it's defined in assert.h as well?

Given the discussion, I would say "no" :)

mihails-strasuns-sociomantic: Yeah I checked how `assert.h` header looks on few systems and it looks like euristic that is doomed to fail. Unless libclang provides some super-magical helpers to do that I think assuming that any `assert` no matter where it comes from acts like D built-in `assert` is reasonable.

jacob-carlborg: Hmm, I see.

mihails-strasuns-sociomantic:
Is there anything else that needs to be done with this PR?

dstep/translator/MacroDefinition.d
```diff
@@ -2053,6 +2053,15 @@ bool translateFunctAlias(
...
+        {
+            // aliasing `assert` from C "assert.h" is commong enough to warrant
+            // a special case to use function instead of an alias
+            output.singleLine("void %s(T...)(T args) { assert(args); }",
```
jacob-carlborg: * It's only possible to pass two arguments two `assert`, a boolean condition and a string
* Please don't use this compact syntax for the generated function

mihails-strasuns-sociomantic:
Updated:
- `if` constraint for amount of `assert` arguments
- multi-line output for the function

jacob-carlborg:
👌

PR #133, opened 2017-05-25
Copy of #105 to simplify CI testing

mihails-strasuns-sociomantic:


jacob-carlborg:
What's different compared to #105?

mihails-strasuns-sociomantic:
One extra commit that fixes CI build on AppVeyor primarily :) I create the PR to test various tweaks on platforms I don't have.

jacob-carlborg:
> One extra commit that fixes CI build on AppVeyor primarily :) I create the PR to test various tweaks on platforms I don't have.

Well, it still doesn't work on Travis CI 😉.

mihails-strasuns-sociomantic:
Yeah and doesn't work on my local system that doesn't create `libclang.so` symlink either :) I am experimenting with alternative approaches to deduce flags.

jacob-carlborg:
So what is the name of the library, something like `libclang.so.x.y`? Doesn't it need to be called `libclang.so` to be able to link with `-lclang`? Doesn't that usually mean that the developer library is missing?

mihails-strasuns-sociomantic:
On ubuntu if you install, for example, `libclang-3.8-dev` package, shared library will be located at `/usr/lib/llvm-3.8/lib/libclang-3.8.so`. Basically you are not allowed to link it as `-lclang`, you always have to use `-lclang-3.8` (or other minor version) explicitly.

jacob-carlborg:
I just checked on Ubuntu 16.10 and 16.04, with the packages `libclang-3.8-dev` and `libclang-3.9-dev`. They all provide a symlink called `libclang.so`.

jacob-carlborg:
BTW, I've just noticed that if you only install `libclang-3.8-dev` it will not install LLVM, meaning `llvm-config` will not be installed :(.

mihails-strasuns-sociomantic:
Right, I have re-checked and it was indeed in the same folder (I was looking for it in `/usr/lib/libclang.so`). `llvm-config` doesn't help much in general because it only controls own LLVM libs, not derived works like libclang.

jacob-carlborg:
> `llvm-config` doesn't help much in general because it only controls own LLVM libs, not derived works like libclang.

Yes, currently it wouldn't matter if we used the path to `llvm-config` or the library path. But in the future they might provide some more useful flags, making it a good abstraction.

PR #144, opened 2017-06-04
Fix #137: struct member expansion.

### LINKED ISSUE #137: struct member expansion
I have a struct:
```c
// FILE object wrapper for vfs access
typedef struct {
    struct DB_vfs_s *vfs;
} DB_FILE;
```

dstep now looks up `struct DB_vfs_s` and expands this inside the parent struct while eating all comments before the definition of `struct DB_vfs_s`. Leading to something like:
```d
// FILE object wrapper for vfs access
struct DB_FILE
{
    // md5 calc control structure (see md5/md5.h)

    // bps must be 32 if this is true

    // lots more comments

    struct DB_vfs_s
    {
        DB_plugin_t plugin;

        // capabilities
        const(char*)* function () get_schemes; // NULL-terminated list of supported schemes, e.g. {"http://", "ftp://", NULL}; can be NULL

```
and so on...

This is a workaround (C code):
```c
typedef struct DB_vfs_s DB_vfs_t;
// FILE object wrapper for vfs access
typedef struct DB_FILE_s {
    DB_vfs_t *vfs;
} DB_FILE;
```
but seems like dstep should cope with using struct notation without expanding it inline.

ciechowoj:
Fixes #137: struct member expansion.

### DIFF
--- a/dstep/translator/Context.d
+++ b/dstep/translator/Context.d
@@ -344,6 +344,14 @@ bool isGlobal(Cursor cursor)
 }
 
 /**
+ * Returns true, if cursor is in the global scope.
+ */
+bool isGlobalLexically(Cursor cursor)
+{
+    return cursor.lexicalParent.kind == CXCursorKind.CXCursor_TranslationUnit;
+}
+
+/**
  * The collectGlobalTypes function scans the whole AST of the translation unit and produces
  * a set of the type names in global scope.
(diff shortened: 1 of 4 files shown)


jacob-carlborg:
👌

jacob-carlborg:
Thanks.

PR #149, opened 2017-06-04
Improve aliasing nested enum members.

ciechowoj:


ciechowoj:
I improved it a little but it stills fails for multiple nesting levels:

```
struct Foo
{
    enum FooEnum
    {
        BAR,
        BAZ,
    };

    struct Bar 
    {
        enum BarEnum
        {
            QUX
        };
    };
};
```

jacob-carlborg:
Do you want to handle multiple levels right away or is this a good first step?

ciechowoj:
No, I don't think I'm going to fix it anytime soon : ) (maybe I'll take a closer look when it starts to affect somebody), so please merge this if you can.

unit_tests/EnumUnitTests.d
```diff
@@ -248,6 +248,64 @@ D", options);
...
+    FOO,
+};
+
+struct Foo
+{
+    enum FooEnum
+    {
+        BAR,
+        BAZ,
+    };
```
jacob-carlborg: Compiling this gives me a warning `warning: declaration does not declare anything`. You need to add a field.

ciechowoj: Is this warning at D side?

ZombineDev: No, structs without fields are essentially undefined behavior in C. They are allowed in C++ and other OOP languages, because they're useful for defining interfaces and abstract base classes, but since there's no such​ concepts in C, they're useless and hence undefined by the C standard, IIRC.

ciechowoj:
Added the members.

jacob-carlborg:
👌

jacob-carlborg:
Thanks.

PR #153, opened 2017-06-19
Tweak assert special case for D1

mihails-strasuns-sociomantic:
Slipped through my attention in previous PR ;)


jacob-carlborg:
👌

PR #162, opened 2017-07-09
Issue154

ciechowoj:
Adds a possibility to specify command line arguments of dstep directly in the `assertTranslates` assertion.
Adds a possibility to enable/disable translation of macro-definitions from command line.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -19,6 +19,7 @@
 * Extend a functionality that automatically replaces aliases to basic types with their D equivalents
 * Add a switch `--dont-reduce-aliases` which disables the above functionality
 * Add a switch `--alias-enum-members` which enables generation of aliases for enum members in the global scope
+* Add a switch `--translate-macros` which can disable the translation of C macro definitions
 * Add the `libclang` bindings as a test case
 * Add support for Microsoft Windows
 * Run Windows tests on AppVeyor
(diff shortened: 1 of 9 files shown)


jacob-carlborg:
It seems like it's possible to pass command line arguments internally to DStep as a string, which will require `splitCommandLine`. What's the purpose of that? Why isn't an array enough?

mihails-strasuns-sociomantic:
Just checked, seems to work for my case. Thanks!

ciechowoj:
> It seems like it's possible to pass command line arguments internally to DStep as a string, which will require splitCommandLine. What's the purpose of that? Why isn't an array enough?

I don't fully understand your question. I hasn't chosen the array as it adds additional unnecessary fuss to the notation (commas, braces). I find the string of arguments more readable. What is more, usually only single argument will be specified (unit test should test only one thing) in which case the string is better as well.

jacob-carlborg:
> I don't fully understand your question. I hasn't chosen the array as it adds additional unnecessary fuss to the notation (commas, braces). I find the string of arguments more readable.

I think it adds unnecessary complexity.

ciechowoj:
What about removing it from the main code and move it the the module where the code for unit tests (assertions) resides?

jacob-carlborg:
It would be better, but I still think it's unnecessary.

ciechowoj:
Right, it isn't strictly necessary, but removes a bit of fuss from our unit
test bocks...

ciechowoj:
Ok, lets merge the other changes right now. I'll revisit the specifying of the options in unit tests on another occasion.

jacob-carlborg:
👌.

PR #163, opened 2017-07-23
Fix the dub warning about -m32mscoff.

ciechowoj:
As in the title.

### DIFF
--- a/appveyor.ps1
+++ b/appveyor.ps1
@@ -2,7 +2,7 @@ echo "Downloading dmd...";
 Invoke-WebRequest "http://downloads.dlang.org/releases/2.x/$env:DVersion/dmd.$env:DVersion.windows.7z" -OutFile c:\dmd.7z;
 7z x c:\dmd.7z -oc:\ > $null;
 echo "Downloading dub...";
-Invoke-WebRequest https://code.dlang.org/files/dub-1.0.0-beta.1-windows-x86.zip -OutFile c:\dub.zip;
+Invoke-WebRequest https://code.dlang.org/files/dub-1.4.0-windows-x86.zip -OutFile c:\dub.zip;
 7z x c:\dub.zip -oc:\dub > $null;
 
 $env:PATH = "c:\dub;$($env:PATH)";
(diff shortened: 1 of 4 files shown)


test.d
```diff
@@ -100,7 +100,7 @@ struct TestRunner
         version (Win64)
             auto result = executeShell("dub test --arch=x86_64");
         else
-            auto result = executeShell("dub test");
+            auto result = executeShell("dub test --arch=x86_mscoff");
```
jacob-carlborg: This will be run for all platforms except Windows 64bit, that's why the Travis CI tests are failing. `x86_mscoff` is only available on Windows, not on the Posix platforms.

appveyor.ps1
```diff
@@ -12,7 +12,7 @@ $env:compilersetup = "C:\Program Files (x86)\Microsoft Visual Studio 14.0\VC\vcv
 
 if ($env:arch -eq "x86") {
   $env:compilersetupargs = "x86";
-  $env:archswitch = "";
+  $env:archswitch = "--arch=x86_mscoff";
```
jacob-carlborg: Is this used?

jacob-carlborg:
With this change, it's not enough to run `dub build` on Windows 32bit, one needs to explicitly run `dub build --arch=x86_mscoff`?

ciechowoj:
With this change, it's not enough to run dub build on Windows 32bit, one needs to explicitly run dub build --arch=x86_mscoff?

Yes, the problem is without `--arch=x86_mscoff` it will try to use OPTLINK linker which we don't support at the moment (if I remember correctly the libclang binaries aren't compatible with this linker).

jacob-carlborg:
Please update the readme to reflect this.

appveyor.ps1
jacob-carlborg: Ping ^.

ciechowoj: Yes.

ciechowoj: In line 45 of the same file. I'm not sure if it is *strictly* necessary at the moment (maybe test.d can be compiled without libclang?) but I think it should be present just not to mix `x86` and `x86_mscoff` architectures.

ciechowoj:
The ideal solution would be to PR a patch witch allows to specify default arch in `dub.json`. I took a look into dub's code and it wouldn't be very complicated albeit not so trivial either. 

The other thing is if the maintainers of dub repo would allow such a extension.

Otherwise I do not think we have other option beside this PR or living with this ugly warning... : (

jacob-carlborg:
👌

ciechowoj:
Thank you.

PR #164, opened 2017-07-30
Enable using dstep as a library

atilaneves:
With this PR it's possible to list `dstep` as a dub dependency and use `subConfiguration "dstep" "library"`


jacob-carlborg:
No changes required to the configuration script?

atilaneves:
Which configuration script? In any case, this is useable as-is with a `subConfiguration "dstep" "library"` in `dub.sdl` for a dependent dub package.

jacob-carlborg:
This configuration script: https://github.com/jacob-carlborg/dstep/blob/master/configure.d.

atilaneves:
Ah, I didn't know `linker_flags.txt` was generated from this. No, there's no need since the `library` configuration won't link to libclang - the final client has to.

dstep/main.d
```diff
@@ -5,125 +5,6 @@
  */
 module dstep.main;
```
jacob-carlborg: Please remove extra empty newline.

dstep/cli.d
```diff
@@ -0,0 +1,193 @@
+module dstep.cli;
+
```
jacob-carlborg: Please remove extra empty newline.

dub.json
```diff
@@ -13,7 +13,7 @@
     "stringImportPaths": ["resources"],
 
     "preGenerateCommands": [
-        "git describe --tags > resources/VERSION"
+        "cd $PACKAGE_DIR && git describe --tags > resources/VERSION"
```
jacob-carlborg: * If I recall correctly, on Windows the `%VAR%` syntax is used to interpolate environment variables
* I don't think this would work if `$PACKAGE_DIR` contains spaces

jacob-carlborg: @MartinNowak will this fix building out-of-tree as you commented here:  https://github.com/jacob-carlborg/dstep/pull/105#issuecomment-318119441?

jacob-carlborg:
I just merged https://github.com/jacob-carlborg/dstep/pull/162, which causes conflicts in this PR.

dstep/main.d
atilaneves: I'm having trouble finding an extra newline here.

dub.json
atilaneves: `$PACKAGE_DIR` isn't an environment variable, it's a special dub variable.

jacob-carlborg: I though DUB would set `$PACKAGE_DIR` as an environment variable when executing the pre generate command. But it looks like it expands the variable before.

dstep/main.d
jacob-carlborg: Hmm, I don't see it anymore.

dub.json
jacob-carlborg: Spaces are still a problem.

jacob-carlborg:
Is this up to date with master? Because now there's a `CommandLine` module and this PR adds a `cli` module.

atilaneves:
I hadn't noticed the new module. I changed things and squashed to one commit.

dstep/driver/CommandLine.d
```diff
@@ -122,3 +122,131 @@ unittest
...
     assert(!config.skipSymbols.find("foo").empty);
 }
+
+/**
+ *  "Real" application entry point, handles CLI/config and forwards to
+ *  dstep.driver.Application to do actual work.
+ *  Called by main.
+ */
+
+int run(string[] args)
```
jacob-carlborg: Pleas remove empty line between the function and the documentation.

dub.json
jacob-carlborg: Spaces are still a problem.

jacob-carlborg:
I forgot to mention this before, but ideally I would like to see a test for this as well. Like a separate DUB package that depends on the DStep package, something similar as I did for DMD [1].

[1] https://github.com/jacob-carlborg/dmd/tree/b5e6ed443db8246e98beec64f55f7717dbe45989/test/dub_package

.travis.yml
```diff
@@ -5,7 +5,9 @@ d:
   - ldc
 
 before_script: git fetch --unshallow
-script: "dub --config=test"
+script:
+  - "dub --config=test"
+  - cd test_package && dub build
```
jacob-carlborg: This will no be run when the tests are running on Windows. Either add it to `test.d` or `appveyor.ps1`. If it's added to `test.d` it will be cross-platform and also be tested when running the tests locally.

atilaneves: I can't run `test.d` locally since Arch Linux isn't supported. Even if it were, I'd rather not download clang when I have it installed. Took a while to figure out `test.d` as well.

jacob-carlborg: > I can't run test.d locally since Arch Linux isn't supported. Even if it were, I'd rather not download clang when I have it installed.

Sorry about that, LLVM doesn't provide pre-compiled releases for Arch Linux. I'm currently working on moving the LLVM matrix from `test.d` to the Travis CI and Appveyor build scripts. Perhaps I can add support for running the tests without downloading LLVM.

jacob-carlborg:
The Windows 32bit tests are failing for some reason. It fails to find `libclang.lib`.

atilaneves:
This is one of the reasons I originally only tested the dependent package on Linux: it's easier and probably not worth it on Windows anyway. I understand wanting to make sure dstep can be used as a library properly but at that point all we're testing is if dub works on Windows - a "real" package would just write `libs "clang"`, make sure that `libclang.lib` was in the right path and be done with it.

I also don't know why it'd fail on Windows 32-bit - I just copied the flags verbatim from dstep's own `dub.json`.

atilaneves:
After many iterations through AppVeyor, it turns out that there is no `C:\Program Files (x86)\LLVM` directory on the machine. After reading the code in fact, that's not even where libclang is to be found, so the main `dub.json` `lflags` for Windows are probably useless but beningly so. The Windows build is ok now.

ciechowoj:
I think the binaries for running test on appveyor are downloaded on demand
by test.d, so that the ones from program files arent used.

The entries in dub.json are for end user, they are supposed to install
appropriate version of libclang manually.

atilaneves:
@ciechowoj I see. Makes sense.

jacob-carlborg:
👌.

jacob-carlborg:
Please squash the commits and ideally commit with an e-mail address matching the one specified in GitHub.

jacob-carlborg:
Please also update the changelog.

changelog.md
```diff
@@ -31,6 +31,7 @@
 * Always test using the latest version of DMD and LDC, including beta and nightly builds
 * Bump the minimum requirement of libclang to 3.9.0
 * Automatically generate the libclang bindings using DStep
+* dstep can now be used as a library
```
jacob-carlborg: `dstep` -> `DStep`.

atilaneves:
I force pushed the change to the capitalisation of dstep in the changelog

jacob-carlborg:
@atilaneves after giving this some though, I'm not sure if it should go in the changelog. I don't think we can guarantee any form of API stability at this point.

ciechowoj:
The stability of the library API?

jacob-carlborg:
> The stability of the library API?

@ciechowoj yes.

jacob-carlborg:
👌.

PR #167, opened 2017-08-15
Fix #166 - take array constant token into account

### LINKED ISSUE #166: macro used in struct declaration not translated properly
C code:

```c
#define __FSID_T_TYPE struct { int __val[2]; }
typedef  __FSID_T_TYPE __fsid_t;
typedef __fsid_t fsid_t;
```

Right now produces this:

```d
extern (C):
 
struct __fsid_t
{
    int[__FSID_T_TYPE] __val;
}
 
alias fsid_t = __fsid_t;
```

The correct code would be:

```d
extern (C):

struct __fsid_t
{
    int[2] __val;
}

alias fsid_t = __fsid_t;
```

atilaneves:


dstep/translator/Type.d
```diff
@@ -350,9 +352,10 @@ body
         {
             if (children[dimension].kind == CXCursorKind.CXCursor_IntegerLiteral)
             {
+                auto token = tokenInsideSquareBrackets(cursor);
                 auto expansions = context.macroIndex.queryExpansion(children[dimension]);
 
-                if (expansions.length == 1)
+                if (expansions.length == 1 && token.kind == TokenKind.identifier)
```
jacob-carlborg: I'm not sure I understand this. Shouldn't the token kind be `literal`?

dstep/translator/Type.d
```diff
@@ -383,6 +386,21 @@ body
...
 
+private Token tokenInsideSquareBrackets(Cursor cursor)
+{
+    import std.algorithm: find;
+    import clang.Token: TokenKind;
+
+    // find the first token after the first opening square bracket
+    auto tokenRange = cursor.tokens
+        .find!(a => a.kind == TokenKind.punctuation && a.spelling == "[");
+    if(tokenRange.empty) return Token.init;
```
jacob-carlborg: If the range is empty there's no way to tell that from the return value. Unfortunately it seems it's not possible to currently have an invalid `Token`. Perhaps we should add a new token kind, `invalid`.

jacob-carlborg:
If currently find it a bit difficult to understand how `__FSID_T_TYPE` ends up being used as the array size.

jacob-carlborg:
@ciechowoj please have a look.

ciechowoj:
> If the range is empty there's no way to tell that from the return value.
Unfortunately it seems it's not possible to currently have an invalid Token.
Perhaps we should add a new token kind, invalid.

Please don'tn then every other function has to check for it as input...

Nullable from typecons?

jacob-carlborg:
> Nullable from typecons?

Ideally it would be `Option` or `Optional` instead of `Nullable`.

dstep/translator/Type.d
atilaneves: No. When it's literal we skip this and do the "normal" thing.

dstep/translator/Type.d
atilaneves: Replaced with `Nullable`.

atilaneves:
@jacob-carlborg `Nullable` _is_ Phobos's option type.

atilaneves:
`__FDID_T_TYPE` ends up getting used for the array size because it matches the macro expansions exactly. It's on the same line as the array declaration so the code expected it to be this:

```c
T var[MACRO];
```

It took me a while to figure out the difference between what the code had seen before, and my example. The difference is that the token in square brackets is a macro (kind: identifier) when you _want_ to go looking for expansions, whereas it was the literal kind in my case.

dstep/translator/Type.d
jacob-carlborg: I did some debugging and it looks like the issue is `expansions[0].spelling`.

jacob-carlborg: If the size of the array is a macro constant this still doesn't work.

atilaneves: If that were true, wouldn't one of the existing unit tests fail?

unit_tests/RecordUnitTests.d
```diff
@@ -524,3 +524,25 @@ D");
...
+unittest {
+    assertTranslates(
+    q{
+
+#define __FSID_T_TYPE struct { int __val[2]; }
+typedef  __FSID_T_TYPE __fsid_t;
+typedef __fsid_t fsid_t;
+
+},
+    q{extern (C):
```
ciechowoj: Use the same format of literals as in other unit tests q"C q"D ...

ciechowoj:
Looks fine for me.

jacob-carlborg:
> Nullable is Phobos's option type

A nullable type is not the same as an optional type.

dstep/translator/Type.d
jacob-carlborg: Not if there are no existing tests that covers this.

jacob-carlborg: With this change the following code:

```c
#define foo 2
#define __FSID_T_TYPE struct { int __val[foo]; }
typedef  __FSID_T_TYPE __fsid_t;
typedef __fsid_t fsid_t;
```

Translates to:

```d
extern (C):

enum foo = 2;

struct __fsid_t
{
    int[__FSID_T_TYPE] __val;
}

alias fsid_t = __fsid_t;
```

Which has the same issue as this PR is trying to solve.

jacob-carlborg: Either I'm completely misunderstanding the Clang AST or there's something wrong with the locations. The cursor that is passed to `context.macroIndex.queryExpansion` is of the kind `CXCursor_IntegerLiteral`. But it's location is, according to Clang, on line 3, column 10 and offset 75. Its spelling is `__FSID_T_TYPE`. How is that an integer literal?

atilaneves:
> A nullable type is not the same as an optional type

Besides the name, what's the difference? Even the documentation says "Practically Nullable!T stores a T and a bool".

unit_tests/RecordUnitTests.d
atilaneves: `q{}` gets highlighted as code, which is what these strings are. `q"C` and `q"D` don't. One might object to using `q{}` for C code but the syntax is so similar that in practice it works (at least it does in Emacs).

ciechowoj:
> q{} gets highlighted as code, which is what these strings are. q"C and q"D don't.
One might object to using q{} for C code but the syntax is so similar that
in practice it works (at least it does in Emacs).

Yeah, but lets keep it consistent with the rest of code. Or change the rest
of code to be consistent with it...

unit_tests/RecordUnitTests.d
jacob-carlborg: It depends on which editor is being used. For example, in TextMate for Ruby, for heredoc strings that uses a symbol which is recognized as a language it will highlight the content of that string. Example:

```ruby
a = <<-C
int b = 3;
C

a = <<-FOO
int b = 3;
FOO
```

It plan to do the same thing with the D bundle for TextMate. That will eventually propagate to GitHub.

The approach of using `q"SYMBOL` for code is much more flexible because it can handle HTML, JSON, CSS and other code as well that is not D code.

jacob-carlborg:
> Yeah, but lets keep it consistent with the rest of code. Or change the rest
of code to be consistent with it...

We should not change the rest of the code. I want my C code to be highlighted as C code and my D code to be highlighted as D code.

jacob-carlborg:
> Besides the name, what's the difference? Even the documentation says "Practically Nullable!T stores a T and a bool".

An optional type usual behaves as container that. If it has a value it is a container of length 1, otherwise it's a container of length 0. In D it should have the range interface. But lets keep the `Nullable` for now since it already exists.

jacob-carlborg:
I still think there's something wrong with the location that Clang gives back for the integer literal cursor.

unit_tests/RecordUnitTests.d
atilaneves: And in Emacs it's not just highlighting but formatting.

It's great that TextMate users will benefit from this, but I'm not one of them and am unlikely to ever be. I _could_ try to do the same for Emacs, but `d-mode` is an offshooot of the hairy mess that is `c-mode` and well, I'd really rather stay away from it.

I'll change it, but I'm not going to like it. `q{}` is part of the D language, `q"D` isn't.

dstep/translator/Type.d
atilaneves: The Clang AST for this is... weird. I'm slightly surprised it gets translated to the right machine code even. The `FieldDecl` ends up having an `IntegerLiteral` child with empty spelling, which on the face of it makes no sense (to me). And yet, this is what libclang gives us.

I'll figure out how to deal with your extended example. It looks like it'll be complicated.

unit_tests/RecordUnitTests.d
jacob-carlborg: >  q{} is part of the D language, q"D isn't.

What do you mean?

jacob-carlborg: You will benefit from this when it propagates to GitHub. GitHub uses the TextMate bundle to highlight D code.

jacob-carlborg: I can give it some though if we can replace `q"D` with `q{}`, but `q"C` should stay. _If_ I want to change it would be for all code and in a separate commit/PR.

dstep/translator/Type.d
jacob-carlborg: The spelling of `token`, which you added, seems to be correct. At least for this case, not sure if the rest of the tests would pass.

jacob-carlborg: I was thinking about trying to create a reduced test case for this in C/C++ and ask on the Clang mailing list if the location is correct.

atilaneves:
@jacob-carlborg Even if libclang is wrong, several versions of it that are wrong would need to be supported. My last version fixes the bug I originally reported and your variant where there are two macros.

tests/unit/RecordUnitTests.d
```diff
@@ -524,3 +524,77 @@ D");
...
+
+});
+
+}
+
+
+@("issue #166 - translate record with name from macro in a literal macro")
+unittest {
+    assertTranslates(
+        q{
```
ciechowoj: q"C [...]"  q"D [...]" keep it consistent.

atilaneves: Right, forgot.

atilaneves:
@ciechowoj Changed it to use consistent quotes.

tests/unit/RecordUnitTests.d
```diff
@@ -524,3 +524,80 @@ D");
 
 }
 
+@("issue #166 - translate record with name from macro")
```
jacob-carlborg: There's a specific directory with tests for GitHub issues [1]. @ciechowoj I'm not sure of the structure of the tests, when to create a separate file for an issue and when not to.

[1] https://github.com/jacob-carlborg/dstep/tree/master/tests/unit/issues

dstep/translator/Type.d
```diff
@@ -388,6 +399,33 @@ body
...
+    import clang.Token: TokenKind;
+
+    auto fromNextBracket(R)(R tokens) {
+        return tokens.find!(a => a.kind == TokenKind.punctuation && a.spelling == "]");
+    }
+
+    auto tokens = cursor.tokens.retro.find!(_ => true);
+
+    // dimension + 1 since dimension is 0-indexed and we need to find at least one
+    foreach(_; 0 .. dimension + 1) {
```
jacob-carlborg: Curly braces on their own lines.

tests/unit/RecordUnitTests.d
ciechowoj: @jacob-carlborg It's you who suggested to put the issues for the tests there, if I remember correctly : ), so maybe for consistency @atilaneves you can put your test there as well...

jacob-carlborg: Hehe, I don't remember anymore 😃.

atilaneves:
@jacob-carlborg I think I've addressed all comments.

jacob-carlborg:
I'll take a look.

jacob-carlborg:
Looking good. Please add an entry in the changelog and squash all commits.

atilaneves:
@jacob-carlborg Done.

jacob-carlborg:
code.dlang.org was done, I've restarted the build.

jacob-carlborg:
👌.

jacob-carlborg:
Thanks.

PR #168, opened 2017-08-16
Enable asking for all typedefs from TypeDefIndex

atilaneves:


jacob-carlborg:
What is this PR doing and that's the purpose?

atilaneves:
If I had your email address I'd write a reply that way. Since I don't I guess I'll have to post here.

I'm working on a project with similar goals to DStep, namely allowing for D programmers to call into C (and hopefully later C++) without manually writing bindings. The approach is different, however.

I thought that instead of going with NIH syndrome that I'd leverage DStep, and my goals in doing so are:

1. Don't reinvent the wheel
2. As much as possible, call into DStep instead of modifying it
3. Contribute back bug fixes / bug reports instead of working in a silo

The TL;DR of this PR is: I need `TypedefIndex` to work this way, and I can't see how I can use the rest of the translation code to do what I want it to do without having to modify it.

DStep has assumptions about what it does; that's fine, it wasn't meant to do what I'm using it for. Unfortunately that means I'll have to modify it, as much as I'm trying to avoid it. Real-world usage has shown that it will be unavoidable.

I understand that perhaps you don't want your codebase to change just because I need flexibility from it; if that's the case, I'll drop the PRs and work on my fork instead. Up to you.

jacob-carlborg:
> If I had your email address I'd write a reply that way. Since I don't I guess I'll have to post here.

The email address is in the Git log 😉.

> I'm working on a project with similar goals to DStep ...

Ok, so that's the high level reason. But I'm still not sure I understand what the PR is doing. Is the original code creating an index with all typedefs in the current file? And the with the PR it includes all typedefs regardless which file?

atilaneves:
> The email address is in the Git log 😉.

I'm an idiot. :P

The original code disconsiders files that aren't the current translation unit when typedefing anonymous structs - they get translated as `_Anonymous_x` intead. If it _is_ in the current translation unit, the translation includes an alias. Given what I'm trying to do, I need all of them to be translated in "the good way". I want:

```c
//foo.h
#include "bar.h"
```
```c
//bar.h
struct foo {
    int dummy;
}
```
to get translated as:

```d
struct foo 
{
    int dummy;
}
```

instead of the current

```d
struct _Anonymous_0
{
   int dummy;
}

```

jacob-carlborg:
Could you please give a complete example. The above code neither contains anonymous structs nor typedefs.

jacob-carlborg:
I'm still not clear on what is the actual problem. I haven't seen an example containing anonymous structs or typedefs.

dstep/translator/TypedefIndex.d
```diff
@@ -12,17 +12,19 @@ import clang.TranslationUnit;
 
 class TypedefIndex
 {
+    import std.typecons: Flag, No;
+
     private Cursor[Cursor] typedefs;
 
-    this(TranslationUnit translUnit)
+    this(TranslationUnit translUnit, Flag!"allTypedefs" allTypedefs = No.allTypedefs)
```
jacob-carlborg: Wouldn't it be better if this was a delegate instead, more of a filter?

atilaneves: Maybe, but without an actual use case that would benefit from it, I'd tend to stick with the simplest thing that works.

atilaneves:
The example I gave isn't great, I was trying to simplify things but went too far. I'll post the real snippet causing problems once I have time.

atilaneves:
Ok, so the real code is this:

```c
//system.h
                      #define __FD_SETSIZE 1024
                      typedef long int __fd_mask;
                      #define __NFDBITS (8 * (int) sizeof (__fd_mask))

                      typedef struct
                      {
                       #ifdef __USE_XOPEN
                          __fd_mask fds_bits[__FD_SETSIZE / __NFDBITS];
                       # define __FDS_BITS(set) ((set)->fds_bits)
                       #else
                          __fd_mask __fds_bits[__FD_SETSIZE / __NFDBITS];
                       # define __FDS_BITS(set) ((set)->__fds_bits)
                       #endif
```

and 

```c
//header.h
#include "system.h"
```

Without this PR, this gets translated wrongly.

ciechowoj:
But if you are translating `header.h` why do you expect anything from `system.h` to be translated in output file? dstep translates only the content of the input file, it doesn't output translation of all dependencies...

ciechowoj:
`dstep system.h` gives me:
```
import core.stdc.config;

extern (C):

enum __FD_SETSIZE = 8;

alias __fd_mask = c_long;
enum __NFDBITS = 8 * cast(int) __fd_mask.sizeof;

struct foo
{
    __fd_mask[0] __fds_bits;
}

extern (D) auto __FDS_BITS(T)(auto ref T set)
{
    return set.__fds_bits;
}
```

`dstep header.h`
```
extern (C):
```
I named the struct foo as your example is clipped.

ciechowoj:
If you are going to use dstep as a library maybe you need something like amalgamation mode where the whole translation unit is translated?

atilaneves:
@ciechowoj:
> But if you are translating header.h why do you expect anything from system.h to be translated in output file?

Because for my use case, I need/want it to.

ciechowoj:
> Because for my use case, I need/want it to.

So, what about instead of making an exception only for typedef translate everything that's included as single flat file?

jacob-carlborg:
@atilaneves I don't want to ague but I still don't understand how your tool is supposed to work. Will you translate everything in the translation unit? Why do you only need this for typedefs?

atilaneves:
I didn't really analyse the implications - running dstep on a header failed, I investigated why and found out that I needed the diff in this PR. I just tried to make the simplest change that would work.

jacob-carlborg:
The reason why I'm continue to ask is that I'm afraid I'm missing some use case that DStep will not work for. Or that there's a bug somewhere.

dstep/translator/TypedefIndex.d
jacob-carlborg: Let me rephrase this a bit. I think the `Flag` pattern is absolutely horrible and I would never accept that in DStep. A plain bool would not be very readable at the call site. A delegate would be more generic and void the problem with the bool.

jacob-carlborg:
I can accept this PR if you do something about the `Flag`.

atilaneves:
I understand why you continue to ask. But I don't know what the answer is, other than I don't think there's a bug.

I don't know how a delegate replaces a bool.

jacob-carlborg:
> I don't know how a delegate replaces a bool.

You replace the if condition with a delegate, that takes the cursor as a parameter and returns a bool.

atilaneves:
@jacob-carlborg Changed to a delegate and the build is green.

jacob-carlborg:
👌.

PR #169, opened 2017-08-24
Pretty enums

ciechowoj:
Adds a functionality to remove superfluous prefixes of the enumeration members.

### DIFF
--- a/unit_tests/RenameEnumMembersUnitTests.d
+++ b/unit_tests/RenameEnumMembersUnitTests.d
@@ -1,0 +1,299 @@
+/**
+ * Copyright: Copyright (c) 2017 Wojciech Szęszoł. All rights reserved.
+ * Authors: Wojciech Szęszoł
+ * Version: Initial created: Aug 08, 2017
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+import std.stdio;
+import Common;
+import dstep.translator.Translator;
+import dstep.translator.Enum;
+
+void assertRenameEnumMembers(
+    string spelling,
+    string[] input,
+    string[] expected,
+    string file = __FILE__,
+    size_t line = __LINE__)
+{
+    import core.exception;
+    import std.algorithm.comparison;
+    import std.format;
+    import std.range;
+
+    auto actual = renameEnumMembers(spelling, input);
+
+    if (!equal(expected, actual))
+    {
+        auto formatString = "\nExpected:\n%s\nActual:\n%s\n";
+        auto message = format(
+            formatString,
+            chain([spelling], expected),
+            chain([spelling], actual));
+
+        throw new AssertError(message, file, line);
+    }
+}
+
+unittest
+{
+    assertRenameEnumMembers(
+        "PrefixFoo",
+        ["PrefixFOO", "PrefixBAR", "PrefixBAZ"],
+        ["FOO", "BAR", "BAZ"]);
+}
+
+unittest
+{
+    assertRenameEnumMembers(
+        "CXFooBar",
+        ["CXFoo_Baz"],
+        ["Baz"]);
+
+    assertRenameEnumMembers(
+        "CXFooBar",
+        ["CXFoo_Baz", "CXFoo_Qux"],
+        ["Baz", "Qux"]);
+}
+
+unittest
+{
+    assertRenameEnumMembers(
+        "CXFooBar",
+        ["CXFoo_Baz"],
+        ["Baz"]);
+
+    assertRenameEnumMembers(
+        "CXFooBar",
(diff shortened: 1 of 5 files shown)


jacob-carlborg:
I just had a quick look so far:

* For the future, it would be nice with with a small example in the PR description, then I don't need to read the tests to figure out what it's doing. Even better, include the example in the commit message as well
* Is this enabled with a flag or always on?
* How does it decide that this renaming will be used?
* There are a few style issues (like extra newlines)
* There's no entry in the changelog (you should know this by now 😃)

ciechowoj:
Find the answers inline.

On Aug 25, 2017 09:00, "jacob-carlborg" <notifications@github.com> wrote:

I just had a quick look so far:

   - For the future, it would be nice with with a small example in the PR
   description, then I don't need to read the tests to figure out what it's
   doing. Even better, include the example in the commit message as well

Acknowledged.


   - Is this enabled with a flag or always on?

It will be off by default.


   - How does it decide that this renaming will be used?

It uses some heuristic (in general it looks at prefixes and tries to guess
the moat suitable one)...


   - There are a few style issues (like extra newlines)

Im workin on it.


   - There's no entry in the changelog (you should know this by now 😃)

Yeah...



—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/jacob-carlborg/dstep/pull/169#issuecomment-324839704>,
or mute the thread

unit_tests/PrettyEnumUnitTests.d
```diff
@@ -0,0 +1,153 @@
...
+        auto message = format(formatString, expected, chain([input.front], actual));
+        throw new AssertError(message, file, line);
+    }
+}
+
+unittest
+{
+    assertRefactorEnum(
+        ["PrefixFoo", "PrefixFOO", "PrefixBAR", "PrefixBAZ"],
+        ["PrefixFoo", "FOO", "BAR", "BAZ"]);
```
jacob-carlborg: Perhaps I don't understand the heuristics yet, but why is `PrefixFoo` renamed?

jacob-carlborg:
Does it rename the members individually or will it rename only if all members can be renamed?

jacob-carlborg:
This is a pattern used by Apple, mostly with Objective-C:

```objc
#include <Foundation/NSObjCRuntime.h>

typedef NS_ENUM(NSUInteger, NSApplicationDelegateReply) {
    NSApplicationDelegateReplySuccess = 0,
    NSApplicationDelegateReplyCancel = 1,
    NSApplicationDelegateReplyFailure = 2
};
```

This is imported to Swift as:

```swift
enum NSApplicationDelegateReply : NSUInteger {
    case success
    case cancel
    case failure
}
```

Will this PR strip the `NSApplicationDelegateReply` prefix in this example?

Would we like to do the lowercasing as well?

ciechowoj:
The first item in the array is the name of the enum itself that is why it
isn't renamed.

ciechowoj:
It looks at all of the members an the name of the enum.

However the case where there is only one member is handled separately.

jacob-carlborg:
> The first item in the array is the name of the enum itself that is why it
isn't renamed.

Aha, and that is part of the heuristics?

ciechowoj:
Yes.

jacob-carlborg:
Cool.

ciechowoj:
> Would we like to do the lowercasing as well?

Dont know. Maybe as an option?

jacob-carlborg:
Would `rename`, instead of `refactor`, be a better name?

ciechowoj:
I'll consider adding it as the default.

ciechowoj:
I'll think about it.

ciechowoj:
Fixed some corner cases.
Added conversion to camelCase.
Added more unit tests.

unit_tests/RenameEnumMembersUnitTests.d
```diff
@@ -0,0 +1,272 @@
...
+        auto message = format(formatString, expected, chain([input.front], actual));
+        throw new AssertError(message, file, line);
+    }
+}
+
+unittest
+{
+    assertRenameEnumMembers(
+        ["PrefixFoo", "PrefixFOO", "PrefixBAR", "PrefixBAZ"],
+        ["PrefixFoo", "FOO", "BAR", "BAZ"]);
```
jacob-carlborg: Wouldn't it be more clear if the enum names (not the members) were passed as separate arguments from the members?

ciechowoj: Oh, whatever.

jacob-carlborg:
I think this looks good. Could you please squash the commits?

ciechowoj:
Rebased.

jacob-carlborg:
👌.

jacob-carlborg:
Thanks.

PR #172, opened 2017-09-23
Fix #123: Missing import for usage of constant.

### LINKED ISSUE #123: Missing import for usage of `#define` constant
Split from https://github.com/jacob-carlborg/dstep/pull/118#issuecomment-279214725

ciechowoj:


ciechowoj:
Fixes issue 123.
Adds IncludeGraph class which allows to query if some file is reachable *through* another include file.

test_files/graph/file.h
```diff
@@ -0,0 +1,20 @@
...
+
+#include "subfile2.h"
+
+
+
+
+#include "subfile3.h"
+
+
+#include "subsubfile2.h"
```
jacob-carlborg: Is there a reason why these includes are spread out across the file like this?

jacob-carlborg:
There's quite a lot of new code here. Could you describe a high level overview of what it does?

ciechowoj:
No.

ciechowoj:
No is for includes. I'll write the description later.

ciechowoj:
The new code is mostly in `dstep/translator/HeaderIndex.d` it computes the graph of the includes (it assumes that the graph is directed acyclic graph) and the computes it's transitive closure. Using the graph you can determine if the given symbol is reachable *through* some include file (e.g. is defined is some other header than the given header). It solves the problem when the symbol from standard library is defined in some internal header which is included in standard header. I currently used it only to solve the problem in the current issue.

ciechowoj:
We may use it to solve the problem with detecting symbols defined in standard library in general case.

test_files/graph/file.h
ciechowoj: No.

jacob-carlborg:
Ok, cool.

test_files/graph/file.h
ciechowoj: Fixed.

dstep/translator/HeaderIndex.d
```diff
@@ -0,0 +1,267 @@
...
+            includedBy[inclusion[0]] ~= inclusion[1];
+
+
+        void visit(size_t header)
+        {
+            if (statuses[header] == Status.visited)
+                return;
+
+            if (statuses[header] == Status.visiting)
+                throw new Exception("The include graph isn't a DAG.");
```
jacob-carlborg: Is this something that could happened or would this indicate a logical error in DStep?

dstep/translator/HeaderIndex.d
```diff
@@ -0,0 +1,267 @@
...
+            "inttypes.h" : 99,
+            "tgmath.h" : 99,
+            "stdint.h" : 99,
+            "stdbool.h" : 99,
+            "stdnoreturn.h" : 11,
+            "threads.h" : 11,
+            "uchar.h" : 11,
+            "stdatomic.h" : 11,
+            "stdalign.h" : 11
+        ];
```
jacob-carlborg: I assume the values are the C standard the header was added to? As far as I can see the value is not used, to you plan to use it in the future or should this be replaced with a set?

dstep/translator/HeaderIndex.d
ciechowoj: This would indicate logical error in dstep of libclang (it would happen if headers made cycle).

dstep/translator/HeaderIndex.d
ciechowoj: I will probably not use it, I did it this wat because there is no literal for set... I would need to create an array and the convert to the set.

jacob-carlborg: Ah, I see. Fair enough.

jacob-carlborg:
Please update changelog.

ciechowoj:
Updated.

jacob-carlborg:
👌.

jacob-carlborg:
Thanks.

PR #176, opened 2017-10-07
Decouple macro parsing from translation.

ciechowoj:
Refactors the code in such a way that the parsing of the macros is decoupled from their translation.
No new functionality added.
The base class Expression has been replaced with the Algebraic type from std.variant. The virtual calls have been replaced with `visit` function.

### DIFF
--- a/dstep/translator/MacroParser.d
+++ b/dstep/translator/MacroParser.d
@@ -1,0 +1,1409 @@
+/**
+ * Copyright: Copyright (c) 2017 Wojciech Szęszoł. All rights reserved.
+ * Authors: Wojciech Szęszoł
+ * Version: Initial created: October 05, 2017
+ * License: $(LINK2 http://www.boost.org/LICENSE_1_0.txt, Boost Software License 1.0)
+ */
+module dstep.translator.MacroParser;
+
+import std.meta;
+import std.traits;
+import std.variant;
+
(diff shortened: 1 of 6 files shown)


dstep/translator/MacroParser.d
```diff
@@ -0,0 +1,1409 @@
...
+}
+
+struct Identifier
+{
+    this (string spelling)
+    {
+        this.spelling = spelling;
+    }
+
+    string spelling;
```
jacob-carlborg: Instance variables go before all methods including constructors.

dstep/translator/Preprocessor.d
```diff
@@ -506,7 +515,7 @@ Directive[] directives(string source)
 
     return directives(TranslationUnit.parseString(index, source));
 }
-
+/+
```
jacob-carlborg: Why are the unit tests out commented?

dstep/translator/MacroParser.d
ciechowoj: Fixed.

dstep/translator/Preprocessor.d
ciechowoj: Sorry, my omission.

jacob-carlborg:
👌.

Issue #141, opened 2017-06-03
Use of _IO, _IOR, etc.


russel:
The libdvbv5 headers connect with the Linux DVB API, which means use of IOCTLs. The standard way of doing this for Linux is with a set of macros, including _IO, _IOR, which are defined in various ioctl.h files. Is it worth handling these within DStep or should it be seen as a "done manually" thing made available via another route?

russel:
I manually converted /usr/include/asm-generic/ioctl.h but this turns:
```
#define _IOC(dir,type,nr,size) \
	(((dir)  << _IOC_DIRSHIFT) | \
	 ((type) << _IOC_TYPESHIFT) | \
	 ((nr)   << _IOC_NRSHIFT) | \
	 ((size) << _IOC_SIZESHIFT))

#define _IOC_TYPECHECK(t) (sizeof(t))

/* used to create numbers */
#define _IO(type,nr)		_IOC(_IOC_NONE,(type),(nr),0)
#define _IOR(type,nr,size)	_IOC(_IOC_READ,(type),(nr),(_IOC_TYPECHECK(size)))
```
into
```
extern (D) auto _IOC(T0, T1, T2, T3)(auto ref T0 dir, auto ref T1 type, auto ref T2 nr, auto ref T3 size)
{
    return (dir << _IOC_DIRSHIFT) | (type << _IOC_TYPESHIFT) | (nr << _IOC_NRSHIFT) | (size << _IOC_SIZESHIFT);
}

extern (D) size_t _IOC_TYPECHECK(T)(auto ref T t)
{
    return t.sizeof;
}

/* used to create numbers */
extern (D) auto _IO(T0, T1)(auto ref T0 type, auto ref T1 nr)
{
    return _IOC(_IOC_NONE, type, nr, 0);
}

extern (D) auto _IOR(T0, T1, T2)(auto ref T0 type, auto ref T1 nr, auto ref T2 size)
{
    return _IOC(_IOC_READ, type, nr, (_IOC_TYPECHECK(size)));
}
```
_IO is fine, however _IOR is a real problem. For example:
```
enum FE_READ_STATUS = _IOR('o', 69, fe_status_t);
enum FE_READ_BER = _IOR('o', 70, __u32);
enum FE_READ_SIGNAL_STRENGTH = _IOR('o', 71, __u16);
enum FE_READ_SNR = _IOR('o', 72, __u16);
[...]

PR #177, opened 2017-10-08
Feature improve handling of known libraries

ciechowoj:
Extends handling of the macro definition dependencies beyond `limits.h`.
Fixes problem with cycles in headers (`throw new Exception("The include graph isn't a DAG.");` has been triggered...).

This PR depends on another PR:
https://github.com/jacob-carlborg/dstep/pull/176

This PR is a step forward to solve:
https://github.com/jacob-carlborg/dstep/issues/141

### DIFF
--- a/dstep/translator/HeaderIndex.d
+++ b/dstep/translator/HeaderIndex.d
@@ -17,10 +17,16 @@ class IncludeGraph
 {
     import std.typecons;
 
+    struct Inclusion
+    {
+        size_t included;
+        size_t including;
+    }
+
     private size_t[string] headers_;
     private string[size_t] reverse_;
-    private Set!(Tuple!(size_t, size_t)) directInclusions;
-    private Set!(Tuple!(size_t, size_t)) indirectInclusions;
+    private Set!Inclusion directInclusions;
+    private Set!Inclusion indirectInclusions;
     private string internalPrefix;
 
     public this(TranslationUnit translationUnit)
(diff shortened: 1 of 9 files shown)


jacob-carlborg:
#176 is merged, please rebase.

ciechowoj:
Rebased.

dstep/translator/HeaderIndex.d
```diff
@@ -170,98 +223,170 @@ class HeaderIndex
 {
     private IncludeGraph includeGraph_;
     private string[string] stdLibPaths;
+    private string[string] knownModules;
     private string mainFilePath;
 
     public this(TranslationUnit translationUnit)
     {
+        immutable string[string] standardModuleMapping = [
```
jacob-carlborg: Shouldn't this be moved out of the constructor as a constant. The same associative array will be created each time the constructor is called, unless the compiler is doing something smart.

ciechowoj: I tried, but D is too dumb for this : (:

```
dstep/translator/HeaderIndex.d(222,50): Error: non-constant expression `["assert.h":"core.stdc.assert_", 
```

jacob-carlborg: Hmm, right. One would have to use a shared module constructor for this.

ciechowoj: Lets keep it simple. It will be called only once per translation unit.

jacob-carlborg:
👌.

jacob-carlborg:
Thanks.

PR #178, opened 2017-10-19
Fix #141: Improve type inference for macros, in particular handle type arguments.

### LINKED ISSUE #141: Use of _IO, _IOR, etc.

ciechowoj:
The commit learns dstep to infer the signature of the macro basing on usage. It is still quite rudimentary implementation but it handles the cases that were wrong before anyway. Example:

```
assertTranslates(q"C
#define BAR(x, y, z) x * y * sizeof(z)

#define BAZ BAR(0, 1, float)
C", q"D
extern (C):

extern (D) auto BAR(z, T0, T1)(auto ref T0 x, auto ref T1 y)
{
    return x * y * z.sizeof;
}

enum BAZ = BAR!float(0, 1);
D");
```

More examples in unit tests.

### DIFF
--- a/dstep/translator/MacroDefinition.d
+++ b/dstep/translator/MacroDefinition.d
@@ -67,7 +67,42 @@ void resolveMacroDependency(Context context, string spelling)
         return context.includeHandler.resolveDependency(*cursor);
 }
 
-string translate(CallExpr expression, Context context, Set!string params, ref Set!string imports)
+struct ExpressionContext
+{
+    Context context;
+    Set!string params;
+    Set!string* imports;
+    Cursor[] scope_;
+    alias context this;
+
+    @disable this();
+
+    private this (int x) { }
+
+    static ExpressionContext make()
+    {
+        struct AA { Set!string aa; }
+        auto result = ExpressionContext(0);
+        result.imports = &(new AA()).aa;
+        return result;
+    }
+
+    static ExpressionContext make(Context context)
+    {
+        auto result = ExpressionContext.make();
+        result.context = context;
+        return result;
+    }
+
+    static ExpressionContext make(Context context, Set!string params)
+    {
+        auto result = ExpressionContext.make(context);
+        result.params = params;
+        return result;
+    }
+}
+
+string translate(CallExpr expression, ExpressionContext context)
 {
     import std.algorithm;
     import std.format;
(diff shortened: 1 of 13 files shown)


dstep/translator/MacroDefinition.d
```diff
@@ -66,7 +67,92 @@ void resolveMacroDependency(Context context, string spelling)
...
+        result.context = context;
+        return result;
+    }
+
+    static ExpressionContext make(Context context, Set!string params)
+    {
+        auto result = ExpressionContext.make(context);
+        result.params = params;
+        return result;
+    }
```
jacob-carlborg: Is there a reason why constructors are not used?

ciechowoj: Yes, you cannot create the associative array in the constructor.

ciechowoj:
@jacob-carlborg  What about making a little review : ) ?

jacob-carlborg:
Yes, sorry. I've been busy with non-D related things.

tests/unit/TypeInferenceTests.d
```diff
@@ -0,0 +1,217 @@
...
+unittest
+{
+    assertTranslates(q"C
+#define FOO(x) BAR(int)
+C", q"D
+extern (C):
+
+extern (D) auto FOO(T)(auto ref T x)
+{
+    return BAR(int);
```
jacob-carlborg: This is translated like this because at this point we don't know what `BAR` is?

ciechowoj: Yes, do you think BAR!int() is better?

jacob-carlborg: Well, `BAR(int);` doesn't compile, regardless if `BAR` is defined in the D code or not. So yes I think `BAR!int()` is better.

ciechowoj: It turns out that it isn't that trivial to change it (more than 1h of work), can we merge it as is for now? I don't have time to fix it right now. Maybe I'll manage after winter...

jacob-carlborg: Sure, can you rebase please.

ciechowoj: Do you remember what did you mean by rebase? There is no merge conflict after all.

jacob-carlborg: No. The only thing I can think of is perhaps it's behind master.

ciechowoj: Rebased.

jacob-carlborg:
Thanks.

PR #182, opened 2017-10-21
Add support for unit-threaded framework.

ciechowoj:
As in the title.


jacob-carlborg:
What is this commit https://github.com/jacob-carlborg/dstep/pull/182/commits/4ce5cae06f5a43fdc374e5b50f08dcbe1fc56767 doing in this PR?

ciechowoj:
You can review them separately and merge at once.

dstep/translator/Record.d
```diff
@@ -34,32 +34,141 @@ void translatePackedAttribute(Output output, Context context, Cursor cursor)
...
     auto spelling = keepUnnamed ? "" : context.translateTagSpelling(cursor);
     spelling = spelling == "" ? spelling : " " ~ spelling;
     auto type = translateRecordTypeKeyword(cursor);
 
     output.subscopeStrong(cursor.extent, "%s%s", type, spelling) in {
+        alias predicate = (a, b) =>
+            a == b ||
+            a.isBitField() &&
+            b.isBitField() &&
+            a.isBitField() == b.isBitField();
```
jacob-carlborg: Isn't this last condition redundant?

dstep/translator/Record.d
```diff
@@ -34,32 +34,141 @@ void translatePackedAttribute(Output output, Context context, Cursor cursor)
...
+    }
+
+    BitField[] bitFields;
+    uint totalWidth = 0;
+
+    foreach (cursor; cursors)
+    {
+        auto width = cursor.bitFieldWidth();
+
+        if (width == 0) {
```
jacob-carlborg: Opening curly brace on the wrong line.

dstep/translator/Record.d
```diff
@@ -34,32 +34,141 @@ void translatePackedAttribute(Output output, Context context, Cursor cursor)
...
+            pad(bitFields, totalWidth);
+            totalWidth = 0;
+        }
+        else {
```
jacob-carlborg: Same here.

dstep/translator/Record.d
```diff
@@ -34,32 +34,141 @@ void translatePackedAttribute(Output output, Context context, Cursor cursor)
...
+    import std.range;
+
+    auto bitFields = translateBitFields(context, cursors);
+
+    if (bitFields.length == 1)
+    {
+        auto bitField = bitFields.front;
+
+        output.singleLine(
+            "mixin(bitfields!(%s, \"%s\", %s));",
```
jacob-carlborg: I prefer to use the backtick string literal, then it's not necessary to escape the quotes.

dstep/translator/Record.d
```diff
@@ -34,32 +34,141 @@ void translatePackedAttribute(Output output, Context context, Cursor cursor)
...
+            bitField.field,
+            bitField.width);
+    }
+    else
+    {
+        output.multiLine("mixin(bitfields!(") in {
+            foreach (index, bitField; enumerate(bitFields))
+            {
+                output.singleLine(
+                    "%s, \"%s\", %s%s",
```
jacob-carlborg: Same here, backtick string literal.

ciechowoj:
It looks like redundant, I need to recheck.

dub.json
```diff
@@ -46,25 +46,35 @@
...
             "sourcePaths": ["dstep", "clang", "tests/unit"],
             "excludedSourceFiles": ["dstep/main.d"],
 
             "preBuildCommands-posix": [
-                "(! [ -s linker_flags.txt ] && ./configure) || true"
+                "(! [ -s linker_flags.txt ] && ./configure) || true",
+                "dub run unit-threaded -c gen_ut_main -- -f bin/ut.d"
+            ],
+
+            "preBuildCommands-windows": [
```
jacob-carlborg: Can the platform suffix be removed to avoid duplicating line [63](https://github.com/jacob-carlborg/dstep/pull/182/files#diff-c440c818ecdf6795a339ff0ad2fba6a0R63) and [67](https://github.com/jacob-carlborg/dstep/pull/182/files#diff-c440c818ecdf6795a339ff0ad2fba6a0R67)?

dstep/translator/Record.d
```diff
@@ -34,32 +34,141 @@ void translatePackedAttribute(Output output, Context context, Cursor cursor)
...
+    string type;
+    string field;
+    uint width;
+}
+
+BitField[] translateBitFields(
+    Context context,
+    Cursor[] cursors)
+{
+    void pad(ref BitField[] bitFields, uint totalWidth)
```
jacob-carlborg: This could perhaps be a static function.

dub.json
ciechowoj: Will we have a guarantee that they will be executed in desired order?

jacob-carlborg: Do you mean the order of the two commands on Posix? I don't think it matters which order those are executed in.

dstep/translator/Record.d
jacob-carlborg: There's nothing wrong with it being nested but since it's not using anything from the surrounding context (like the parameters of `translateBitField`) it can be `static`. That will avoid allocating the extra context pointer but it will still be lexically nested in `translateBitFields`. It's just a minor detail but I think it keeps the code nice and tidy.

ciechowoj: I think any decent compiler should figure it out...

dub.json
ciechowoj: Fixed.

changelog.md
```diff
@@ -33,6 +33,7 @@
 * Automatically generate the libclang bindings using DStep
 * Add a switch `--rename-enum-members`, which enables stripping of superfluous prefixes from enum members.
 * Add a switch `--normalize-modules`, which enables changing the names of modules to the D conforming form.
+* Add support for unit-threaded for running unit tests.
```
jacob-carlborg: Please add an entry for the support for bitfields as well.

jacob-carlborg:
It doesn't look like D1 have `std.bitmanip.bitfields`.

dstep/translator/Record.d
ciechowoj: My point if its nested its clear that its needed only in the scope of parent function (e.g. it doesn't pollute the global namespace), to make it static I would need to extract it and the coupling between the functions becomes less clear.

changelog.md
ciechowoj: Fixed.

ciechowoj:
I removed the commit that adds unit threaded. Not this adds support for bit fields only. I came to the conclusion that unit threaded doesn't solve much problems and its another dependency after all.

dstep/translator/Record.d
ciechowoj: Fixed.

ciechowoj:
> It doesn't look like D1 have std.bitmanip.bitfields. I would be fine if this feature only exists for D2.

I added `version(D1)` in appropriate place.

dstep/translator/Record.d
jacob-carlborg: To make it static, you just add the `static` keyword.

ciechowoj: Heh, the obvious solutions are hardest to see : ) I'll fix it on next PR...

PR #187, opened 2018-03-18
feature-refactor-driver

ciechowoj:
Refactors driver code. Replaces explicit threads with facilities from std.parallelism.

Splits translation into two steps:
1. clang step - creation of translation units
2. dstep step - translation

Thanks to above another step - global analysis - can be added in between.

### DIFF
--- a/dstep/driver/Application.d
+++ b/dstep/driver/Application.d
@@ -40,57 +40,34 @@ class Application
     public void run ()
     {
         import std.exception : enforce;
+        import std.range;
 
         enforce!DStepException(config.inputFiles.length > 0,
             "dstep: error: must supply at least one input file\n");
 
         enforceInputFilesExist(config);
 
-        // when only one input file is supplied, -o argument is
-        // interpreted as file path, otherwise as base directory path
-        bool singleFileInput = config.inputFiles.length == 1;
+        auto translationUnits = makeTranslationUnits(config);
 
-        Task!(
-            startParsingFile,
-            Configuration,
-            string,
-            string)*[] conversionTasks;
+        enforceTranslationUnitsCompiled(translationUnits);
 
-        conversionTasks.length = config.inputFiles.length;
+        auto inputFiles = config.inputFiles;
+        auto outputFiles = makeOutputFiles(config);
 
-        // parallel generation of D modules for each of input files
-        foreach (size_t index, fileName; config.inputFiles)
+        foreach (tuple; zip(inputFiles, outputFiles, translationUnits).parallel(1))
         {
-            string outputFilename;
+            string outputDirectory = Path.dirName(tuple[1]);
 
-            if (singleFileInput)
-            {
-                if (config.output.length)
-                    outputFilename = config.output;
-                else
-                    outputFilename = defaultOutputFilename(fileName, false);
-            }
-            else
-            {
-                outputFilename = Path.buildPath(config.output,
-                    defaultOutputFilename(fileName, false));
-            }
+            if (!exists(outputDirectory))
+                mkdirRecurse(outputDirectory);
 
-            string outputDir = Path.dirName(outputFilename);
+            Options options = this.config.toOptions(tuple[0], tuple[1]);
 
-            if (!exists(outputDir))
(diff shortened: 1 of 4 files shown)


dstep/driver/Application.d
```diff
@@ -40,57 +40,33 @@ class Application
...
-            Configuration,
-            string,
-            string)*[] conversionTasks;
+        enforceTranslationUnitsCompiled(translationUnits);
 
-        conversionTasks.length = config.inputFiles.length;
-
-        // parallel generation of D modules for each of input files
-        foreach (size_t index, fileName; config.inputFiles)
+        foreach (tuple; taskPool.parallel(zip(inputFiles, outputFiles, translationUnits), 1))
```
jacob-carlborg: * I think this would look nicer with UFCS
* I don't think `taskPool` is required, see [1]

[1] https://dlang.org/phobos/std_parallelism.html#.parallel

dstep/driver/Application.d
```diff
@@ -110,115 +86,84 @@ class Application
...
-        import std.exception : enforce;
+        Index translationIndex = Index(false, false);
+        Compiler compiler;
 
-        bool translate = true;
-        auto message = Appender!string();
+        auto translationUnits = new TranslationUnit[config.inputFiles.length];
 
-        foreach (diag ; diagnostics)
+        foreach (index, ref unit; taskPool.parallel(translationUnits))
```
jacob-carlborg: Can this be replaced with a map?

dstep/driver/Application.d
```diff
@@ -40,57 +40,33 @@ class Application
...
         enforce!DStepException(config.inputFiles.length > 0,
             "dstep: error: must supply at least one input file\n");
 
         enforceInputFilesExist(config);
 
-        // when only one input file is supplied, -o argument is
-        // interpreted as file path, otherwise as base directory path
-        bool singleFileInput = config.inputFiles.length == 1;
+        auto inputFiles = config.inputFiles;
+        auto outputFiles = makeOutputFiles(config);
```
jacob-carlborg: Looks like it would be better to perform this step after `enforceTranslationUnitsCompiled`. if `enforceTranslationUnitsCompiled` fails, then `makeOutputFiles` won't be run.

ciechowoj: Yes, that makes sense.

dstep/driver/Application.d
ciechowoj: Do you know why this doesn't compile:

```
    import std.parallelism;

    Index translationIndex = Index(false, false);
    Compiler compiler;

    TranslationUnit delegate(string) functor = inputFile => TranslationUnit.parse(
        translationIndex,
        inputFile,
        config.clangParams ~ compiler.extraIncludeFlags,
        compiler.extraHeaders);

    return taskPool.map!functor(config.inputFiles, 1).array;
```

Error:
`dstep/driver/Application.d(162,24): Error: template instance map!(functor) cannot use local 'functor' as parameter to non-global template map(functions...)`

ciechowoj: ?

ciechowoj: I think this is the reason I haven't used the map in the first place (I created the change quire some time ago, it only the PR which were created yesterday).

https://issues.dlang.org/show_bug.cgi?id=5710

This is (another) stupid limitation of D : /.

dstep/driver/Application.d
ciechowoj: Fixed.

dstep/driver/Application.d
ciechowoj: Won't fix.

dstep/driver/Application.d
ciechowoj: Fixed.

dstep/driver/Application.d
jacob-carlborg: Ah, ok. That issue again 😞.

jacob-carlborg:
Thanks.

ciechowoj:
Thanks : )

PR #196, opened 2018-05-08
Consider types when doing import resolution.

ciechowoj:


jacob-carlborg:
There are no tests for this.

ciechowoj:
> There are no tests for this.
Maybe I'll think about more generic solution later, let's not merge it right now.

ciechowoj:
Closing as duplicate to #206 .

PR #213, opened 2018-10-19
Add missing check for null pointer

jens-mueller-sociomantic:
This patch adds a reduced test case for a C macro defining a function
call `(bar)(a)`, i.e., calling the function `bar` on `a`. The extra
parentheses make the test result in a segmentation fault due to null
pointer dereferencing. To fix the problem the null pointer check is
performed as part of an early return. This makes the code more robust
and allows removing a later null check.

### DIFF
--- a/dstep/translator/MacroDefinition.d
+++ b/dstep/translator/MacroDefinition.d
@@ -521,6 +521,7 @@ bool translateFunctAlias(
     if (expr !is null)
     {
         Identifier* ident = expr.expr.peek!Identifier;
+        if (ident is null) return false;
 
         if (ident.spelling == "assert")
         {
@@ -541,8 +542,7 @@ bool translateFunctAlias(
             return true;
         }
 
-        if (ident !is null &&
-            equal(definition.params, expr.args
+        if (equal(definition.params, expr.args
                 .map!(a => a.translate(expressionContext))))
         {
             version (D1)
--- a/tests/unit/MacroTranslTests.d
+++ b/tests/unit/MacroTranslTests.d
@@ -159,6 +159,21 @@ D");
 
 }
 
+// Translate function call.
+unittest
+{
+    assertTranslates(q"C
+#define FOO(a) (bar)(a)
+C", q"D
+extern (C):
+
+extern (D) auto FOO(T)(auto ref T a)
+{
+    return bar(a);
+}
+D");
+}
+
 // Translate cast operator.
 unittest
 {
@@ -170,7 +185,6 @@ extern (D) auto FOO(T)(auto ref T a)
     return cast(float) a;
 }
 D");
-
 }
 
 // Translate unary operators.


jens-mueller-sociomantic:
Note, the added test is failing. Maybe it needs to be changed but this is what I would have expected. We should clarify this.

dstep/translator/MacroDefinition.d
```diff
@@ -522,7 +522,7 @@ bool translateFunctAlias(
     {
         Identifier* ident = expr.expr.peek!Identifier;
 
-        if (ident.spelling == "assert")
+        if (ident !is null && ident.spelling == "assert")
```
jacob-carlborg: There's a null check further down, I suggest removing that and use an early return instead with the end result of only one null check.

jens-mueller-sociomantic: Good idea. Should be more robust.

jens-mueller-sociomantic:
```diff
diff --git a/dstep/translator/MacroDefinition.d b/dstep/translator/MacroDefinition.d
index f089ccd..baa3f97 100644
--- a/dstep/translator/MacroDefinition.d
+++ b/dstep/translator/MacroDefinition.d
@@ -521,8 +521,9 @@ bool translateFunctAlias(
     if (expr !is null)
     {
         Identifier* ident = expr.expr.peek!Identifier;
+        if (ident is null) return false;
 
-        if (ident !is null && ident.spelling == "assert")
+        if (ident.spelling == "assert")
         {
             // aliasing `assert` from C "assert.h" is commong enough to warrant
             // a special case to use function instead of an alias
@@ -541,8 +542,7 @@ bool translateFunctAlias(
             return true;
         }
 
-        if (ident !is null &&
-            equal(definition.params, expr.args
+        if (equal(definition.params, expr.args
                 .map!(a => a.translate(expressionContext))))
         {
             version (D1)
```

jens-mueller-sociomantic:
@jacob-carlborg What about the test case?

jacob-carlborg:
> Note, the added test is failing. Maybe it needs to be changed but this is what I would have expected. We should clarify this.

It translates the "body" of the macro to a function call, i.e. `bar(a)`, which is correct. The following C example will call the `bar` function:

```c
#include <stdio.h>

#define FOO(a) (bar)(a)

void bar(int b)
{
    printf("bar b=%d\n", b);
}

int main(int c, char** v)
{
    int x = 1;
    FOO(x);
    return 0;
}
```

As far as I know `(bar)(a)` cannot be anything else than a function call in C. Any user defined types in C needs to have `struct` for `enum` prefixed when referring to it, i.e. `struct bar`.

jens-mueller-sociomantic:
What if `bar` is `#define bar int`? I think this was the code I reduced the test case from. Anyway I'm fine with any test.

jens-mueller-sociomantic:
But note `#define FOO(a) (int)(a)` won't segfault.

jens-mueller-sociomantic:
Alright. There no need for this complication. This changed test case also
triggers the seg fault.

```diff
diff --git a/tests/unit/MacroTranslTests.d b/tests/unit/MacroTranslTests.d
index d0428e6..8f75801 100644
--- a/tests/unit/MacroTranslTests.d
+++ b/tests/unit/MacroTranslTests.d
@@ -178,7 +178,7 @@ extern (C):
 
 extern (D) auto FOO(T)(auto ref T a)
 {
-    return cast(bar) a;
+    return bar(a);
 }
 D");
 }
```

jacob-carlborg:
> What if bar is #define bar int? I think this was the code I reduced the test case from. Anyway I'm fine with any test.

That's a good point. @ciechowoj any input?

jens-mueller-sociomantic:
It is important but unrelated to this issue. I think you want to add the test case

```c
#define bar float
#define FOO(a) (bar)(a)
```

which should create something along the lines of

```d
extern (C):
extern (D) auto FOO(T)(auto ref T a)
{
    return cast(float) a;
}
```

But that is another issue which can be addressed elsewhere.

That said I think I need to move the added test case because it is not about translating the cast operator. What do you think?

jacob-carlborg:
> That said I think I need to move the added test case because it is not about translating the cast operator. What do you think?

Yes, please do.

ciechowoj:
> That's a good point. @ciechowoj any input?
I guess it should be sth more like:

```
#define bar float
#define FOO(a) (bar)(a)
```
->
```
extern (C):

alias bar = float;

extern (D) auto FOO(T)(auto ref T a)
{
    return cast(bar) a;
}
```

ciechowoj:
You can add a bug or feature request : )

jacob-carlborg:
```c
#define bar float
#define FOO(a) (bar)(a)
```

Is currently translated to this:

```d
extern (C):

alias bar = float;

extern (D) auto FOO(T)(auto ref T a)
{
    return bar();
}
```

jacob-carlborg:
@jens-mueller-sociomantic still waiting for you to move the test.

jens-mueller-sociomantic:
```diff
diff --git a/tests/unit/MacroTranslTests.d b/tests/unit/MacroTranslTests.d
index 8f75801..8550e23 100644
--- a/tests/unit/MacroTranslTests.d
+++ b/tests/unit/MacroTranslTests.d
@@ -159,6 +159,21 @@ D");
 
 }
 
+// Translate function call.
+unittest
+{
+    assertTranslates(q"C
+#define FOO(a) (bar)(a)
+C", q"D
+extern (C):
+
+extern (D) auto FOO(T)(auto ref T a)
+{
+    return bar(a);
+}
+D");
+}
+
 // Translate cast operator.
 unittest
 {
@@ -169,17 +184,6 @@ extern (D) auto FOO(T)(auto ref T a)
 {
     return cast(float) a;
 }
-D");
-
-    assertTranslates(q"C
-#define FOO(a) (bar)(a)
-C", q"D
-extern (C):
-
-extern (D) auto FOO(T)(auto ref T a)
-{
-    return bar(a);
-}
 D");
 }
```

jacob-carlborg:
Thanks.

PR #219, opened 2019-01-05
Fix #214: Cast inside macro not properly translated

### LINKED ISSUE #214: Cast inside macro not properly translated
```c
#define bar float
#define FOO(a) (bar)(a)
```

Is currently translated to this:

```d
extern (C):

alias bar = float;

extern (D) auto FOO(T)(auto ref T a)
{
    return bar();
}
```

But should be translated to:

```d
extern (C):

alias bar = float;

extern (D) auto FOO(T)(auto ref T a)
{
    return cast(bar) a;
}
```

ciechowoj:


jacob-carlborg:
Thanks.

PR #220, opened 2019-01-05
Fix #216: can't alias to existing D type

### LINKED ISSUE #216: can't alias to existing D type
when translating the header from https://github.com/ggambetta/libz80

dstep emitted the following code (in part)

`alias ushort = ushort;`
`alias byte = ubyte;`

fortunately its not a large header so I could just comment these alias's out and change any byte to ubyte ...

ciechowoj:


dstep/translator/Type.d
```diff
@@ -201,8 +201,13 @@ package string reduceAlias(Type type)
         case "u32": return "uint";
         case "u64": return "ulong";
 
-        default: return null;
+        default: break;
     }
+
+    if (type.spelling == "byte" && type.canonical.kind == CXTypeKind.uChar)
```
jacob-carlborg: I think it would be better to have this go through the `translateType`:

https://github.com/jacob-carlborg/dstep/blob/0bba701beb5413037d93a03a45b4b58207a04e91/dstep/translator/Type.d#L552

ciechowoj: Fixed, unfortunately it required passing context is some more places.

dstep/translator/Type.d
```diff
@@ -201,13 +201,18 @@ package string reduceAlias(Type type)
         case "u32": return "uint";
         case "u64": return "ulong";
 
-        default: return null;
+        default: break;
     }
+
+    if (type.spelling == "byte" && translateType(context, type.canonical.kind) == "ubyte")
```
jacob-carlborg: Why not just return what `translateType` returns?

ciechowoj: Because then dstep would incorrectly translated rubbish as `typedef unsigned int byte` to `ubyte`.

ciechowoj: Sorry, to `uint`.

jacob-carlborg: Hmm, right.

jacob-carlborg: Technically the above switch statement should check the type kind as well. I'm wondering if we should replace the whole implementation of this function with an associative array, mapping a tuple of the type kind and the spelling to the result.

jacob-carlborg: Something like:

```d
enum aliasMapping = [
    tuple("byte", CXTypeKind.uChar): "ubyte",
    .
    .
    .
]

return aliasMapping[tuple(type.spelling, type.canonical.kind)];
```

ciechowoj: Do you have in minds other reductions than byte -> ubyte. Any other type (short, int, long) cannot be used as this as it's keyword in C and cannot be used as typedef.

ciechowoj: I thought about this but it looks like byte is the only special case.

jacob-carlborg: All the typedefs in the above switch statement. Currently it just checks the spelling without checking the underlying type.

tests/functional/objc/primitives.d
```diff
@@ -3,6 +3,6 @@ extern (Objective-C):
 extern __gshared ObjcObject x9;
 extern __gshared Class x10;
 extern __gshared SEL x11;
-extern __gshared bool x12;
+extern __gshared BOOL x12;
```
jacob-carlborg: `BOOL` is `signed char`.

ciechowoj: Hmm, is `bool` guaranteed to be binary compatible with `signed char` in D?

dstep/translator/Type.d
ciechowoj: I have some problems with typedefs to `long` and `unsigned long` as they can be 32/64 bit... Windows tests are failing : |

tests/functional/objc/primitives.d
ciechowoj: I have some problems with typedefs to long and unsigned long as they can be 32/64 bit... Windows tests are failing : |

jacob-carlborg: > Hmm, is bool guaranteed to be binary compatible with signed char in D

They're both the same size. But it's not guaranteed since the size of `char` (in C) is not guaranteed.

> I have some problems with typedefs to long and unsigned long as they can be 32/64 bit... Windows tests are failing : |

Of course :(.

ciechowoj: > But it's not guaranteed since the size of char (in C) is not guaranteed.

But it's de facto 1 byte, so I guess we can add:

tuple("BOOL", CXTypeKind.uChar): "bool",
tuple("BOOL", CXTypeKind.sChar): "bool",

to the table? Or not?

ciechowoj:
Could it be merged?

tests/unit/issues/Issue46.d
```diff
@@ -18,10 +18,14 @@ unittest
     assertTranslates(
 q"C
 
+// This is not strictly correct as __s64 is defined as `typedef __signed__ long __s64;`,
+// however it breaks the test in Windows if left in original form (the types do not exist on Windows).
+// An alternative is to disable this test on Windows completely.
```
jacob-carlborg: I rather have the test being completely correct and disabled on Windows.

jacob-carlborg:
Thanks.

tests/unit/issues/Issue46.d
```diff
@@ -10,12 +10,14 @@ import Common;
 import dstep.translator.Options;
 
 // Fix 46: Generating code that will not compile.
-unittest
+version (Posix)
```
jacob-carlborg: Could have used the `version (Posix):` syntax to avoid the extra indentation. But that's a minor issue.

PR #222, opened 2019-03-29
Replace rdmd usage with dmd to decrease dependencies

Marenz:
Well, the true reason is that I just didn't manage to install rdmd on archlinux.
But using dmd directly still is better than yet another binary, right?

### DIFF
--- a/configure
+++ b/configure
@@ -2,4 +2,15 @@
 
 set -e
 
-rdmd configure.d "$@"
+if test -z "$DC"; then
+    if test -x "$(which dmd)"; then
+        DC=dmd
+    elif test -x "$(which ldc2)"; then
+        DC=ldc2
+    else
+        echo "Neither DMD nor LDC found to compile configure.d"
+        exit -1
+    fi
+fi
+
+$DC -run configure.d "$@"


Marenz:
added support for LDC

configure
```diff
@@ -2,4 +2,13 @@
...
+if test -x "$(which dmd)"; then
+	DCOMPILER=dmd
+elif test -x "$(which ldc)"; then
+	DCOMPILEr=ldc
+else
+	echo "Neither DMD nor LDC found to compile configure.d"
+	exit -1
+fi
+
+$DCOMPILER -i -run configure.d "$@"
```
jacob-carlborg: Looks like the indentation is done using tabs. Please use spaces, 4.

jacob-carlborg: The `-i` flag is not necessary. It's only one file.

jacob-carlborg:
It looks like the tests are failing because LDC is not in the path.

Marenz:
hmm can I rely on $DC being set in that environment?

jacob-carlborg:
Yes, only in the CI environment.

Marenz:
Updated

configure
```diff
@@ -2,4 +2,15 @@
 
 set -e
 
-rdmd configure.d "$@"
+if test -z "$DC"; then
+    if test -x "$(which dmd)"; then
+        DC=dmd
+    elif test -x "$(which ldc)"; then
+        DC=ldc
```
jacob-carlborg: The actual binary is named `ldc2`.

Marenz:
Renamed ldc to ldc2

jacob-carlborg:
Thanks.

PR #223, opened 2019-04-22
Improve Windows build script

GoaLitiuM:
The old `generate_version.bat` script would leave the current directory in the dstep directory after finishing the build command, and in some cases the `cd` command will not change the directory when current directory is in different drive.

This change fixes these issues and also disables echoing of the commands back to terminal.

### DIFF
--- a/tools/generate_version.bat
+++ b/tools/generate_version.bat
@@ -1,7 +1,11 @@
-cd "%DUB_PACKAGE_DIR%"
+@echo off
+
+pushd "%DUB_PACKAGE_DIR%"
 
 if exist .git (
   git describe --tags > resources\VERSION
 ) else (
   echo unknown > resources\VERSION
 )
+
+popd


tools/generate_version.bat
```diff
@@ -1,7 +1,12 @@
-cd "%DUB_PACKAGE_DIR%"
+@echo off
+
+pushd %CD%
+cd /D "%DUB_PACKAGE_DIR%"
```
jacob-carlborg: Why is both `pushd` and `cd` necessary?

GoaLitiuM: I actually didn't know `pushd` also changes the directory, I only thought it would store the given path to a stack. So in this case, `pushd "%DUB_PACKAGE_DIR%"` should be enough.

jacob-carlborg:
Thanks.

PR #241, opened 2019-07-07
Update FreeBSD configuration

myfreeweb:
- `ncursesLib` → `additionalLib` was missed
- `stdc++` hasn't been the default c++ library for *ages*
- I don't recall LLVM paths ever being `/usr/lib/llvm-0.0`, FreeBSD has always installed packages to `/usr/local` and they were `/usr/local/llvm00`..
- (you might want to update llvm paths on all platforms, it mentions really really old versions)

### DIFF
--- a/configure.d
+++ b/configure.d
@@ -203,16 +203,16 @@ static:
         ];
 
         immutable llvmLibPaths = [
-            "/usr/lib/llvm-4.0/lib",
-            "/usr/lib/llvm-3.9/lib",
-            "/usr/lib/llvm-3.8/lib",
-            "/usr/lib/llvm-3.7/lib"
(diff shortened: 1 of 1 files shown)


jacob-carlborg:
Thanks.

jacob-carlborg:
> I don't recall LLVM paths ever being /usr/lib/llvm-0.0, FreeBSD has always installed packages to /usr/local and they were /usr/local/llvm00..

I might have googled that.

> (you might want to update llvm paths on all platforms, it mentions really really old versions)

Yes. But there's the `--llvm-path` flag that can be used when the default configuration is not suitable. This is used to build the pre-compiled binaries.

PR #244, opened 2019-10-02
Use llvm-config to get the library directory

BorisCarvajal:
llvm-config will look for the default LLVM installation, you can pass the --llvm-path to force linking to other version.

The old behavior is still there if llvm-config method fails.


jacob-carlborg:
I’ve tried to use `llvm-config` before, it didn’t work that well. I don’t remember the exact details.

PR #245, opened 2019-12-08
Add experimental meson.build file

danyspin97:


meson.build
```diff
@@ -0,0 +1,78 @@
+project('dstep', 'd', version: '1.0.1')
+
+if get_option('buildtype') == 'release'
+    add_project_arguments('-O3', language: 'd')
```
FFY00: This is wrong. A project should never make this kind of decisions if not required.

FFY00 (changes requested):

meson.build
DanySpin97: It has been removed.

FFY00 (approved):
Seems fine. Perhaps there could be a better way to do the clang stuff but I don't have time to investigate.

DanySpin97:
The current method uses llvm-config, which can also be configured when cross compiling according to the system.

jacob-carlborg:
I don't think I want to have an additional build system. What's the reason for adding another one?

FFY00:
Dub is notoriously bad, especially when it comes to software distribution. Having a build system that actually supports pkgconf properly is very important IMO.

PR #248, opened 2020-01-16
Input/Output dir options

### LINKED ISSUE #243: Please add dedicated output directory argument
Something like `-O`

Current smart universal `-o` is not very comfortable, especially for usage with execution from `find`, with results what wanted to be placed into one specified output directory - dstep complains that the directory is specified, but dstep wants file name in this case.

denizzzka:
Fix #243

### DIFF
--- a/dstep/Configuration.d
+++ b/dstep/Configuration.d
@@ -34,8 +34,11 @@ struct Configuration
     /// array of parameters needed to be forwarded to clang driver
     string[] clangParams;
 
-    /// output file name or folder (in case there are many input files)
-    string output;
+    /// output file name or folder
+    string outputPath;
+
+    /// in case there are many input files or dir output option provided
+    bool isOutputToDir;
 
     /// package name
     @("package", "Use <package> as package name.")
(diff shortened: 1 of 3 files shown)


dstep/Configuration.d
```diff
@@ -34,8 +37,11 @@ struct Configuration
     /// array of parameters needed to be forwarded to clang driver
     string[] clangParams;
 
-    /// output file name or folder (in case there are many input files)
-    string output;
+    /// output file name or folder
+    string outputPath;
+
+    /// in case there are many input files or dir output option provided
+    bool isOutputToDir;
```
jacob-carlborg: Is the intention that both `-O` and `-o` are specified since this is a bool?

jacob-carlborg (changes requested):
Please include tests and a changelog entry.

dstep/Configuration.d
denizzzka: Yes

denizzzka:
> Please include tests

I could not do this so that there were a minimum number of changes.

denizzzka:
I run it by this way:
```
dstep -I -O -o ./dst_dir/ ./src_dir/
```

jacob-carlborg:
> I could not do this so that there were a minimum number of changes.

I’m not sure I follow. You can have a look here at how to to implement a test:

https://github.com/jacob-carlborg/dstep/blob/master/tests/functional/Tests.d

PR #272, opened 2021-07-30
Replace opIn with opBinary operator

ibuclaw:
`opIn` is due to be removed in https://github.com/dlang/dmd/pull/12902.


jacob-carlborg (approved):

jacob-carlborg:
Thanks.

PR #292, opened 2025-03-13
libclang 19.1.7

Inkrementator:
Fix anonymous composites in clang 19.1.7

In some new clang versions, the name (cursor.spelling) of anonymous structs,
unions and enums was set to an informative message instead of an empty
string. I.e. "enum (unnamed at [declaration_file.h:48]"

This fixes the errors that happen due to that in a backwards compatible
way, since it still checks for empty spellings.


jacob-carlborg:
Could you please rebase your branch. I've fixed the failing CI pipeline.

jacob-carlborg (approved):

jacob-carlborg:
Thanks.

PR #297, opened 2026-03-18
Fix incorrect spelling collision for struct typedef

davispuh:
Consider a case where we have 2 files:
`first.h`
```c
typedef struct {
  int a;
} B;
```
And
`second.h`
```c
#include "first.h";

typedef B C;
```

Then currently trying to convert `second.h` we get warning:
```
$ dstep second.h
first.h:2:9: warning: a type renamed to 'B_' due to the collision with the symbol declared in first.h:4:3
```

This is not correct, because `typedef B C` means `C` is alias to `B` and it's not a collision.
This PR fixes this issue.

### DIFF
--- a/dstep/translator/Context.d
+++ b/dstep/translator/Context.d
@@ -286,6 +286,12 @@ class Context
                 if (options.collisionAction != CollisionAction.ignore)
                 {
                     auto collision = spelling in macroIndex.globalCursors;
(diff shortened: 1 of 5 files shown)


jacob-carlborg:
I think the code review from Copilot is reasonable.

davispuh:
Improved PR and added tests.

PR #299, opened 2026-03-18
Fix integers with LL suffix

davispuh:
Consider `a.h` with:
```c
#define LARGE 3LL
```
Currently `dstep` would produce
```d
extern (C):

enum LARGE = 3LL;
```
Which is not valid D code because D language doesn't have `LL` suffix.

```
$ ldc -c a.d
a.d(3): Error: repeated integer suffix `L`
enum LARGE = 3LL;
             ^
```
This PR fixes that so in such case result will be:
```d
extern (C):

enum LARGE = 3L;
```

### DIFF
--- a/dstep/translator/MacroDefinition.d
+++ b/dstep/translator/MacroDefinition.d
@@ -10,6 +10,7 @@ import std.array : Appender;
 import std.traits;
 import std.meta;
 import std.variant;
+import std.string;
 
 import clang.c.Index;
 import clang.Cursor;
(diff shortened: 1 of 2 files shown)


dstep/translator/MacroDefinition.d
```diff
@@ -92,19 +92,28 @@ string translate(Literal literal, ExpressionContext context)
...
     auto uinteger = integer.stripLeft!(x => x == '+' || x == '-');
+    auto suffix = literal.spelling[integer.length .. $];
 
     if (uinteger.length > 1 &&
         uinteger.all!isDigit &&
         uinteger.front == '0')
     {
         (*context.imports).add("std.conv : octal");
-        auto suffix = literal.spelling[integer.length .. $];
+        if (suffix.length >= 2 && suffix[$-2..$] == "LL")
```
jacob-carlborg: There's a function in the standard library that checks if a string ends with another string, can't that be used? https://dlang.org/phobos/std_algorithm_searching.html#.endsWith

jacob-carlborg (changes requested):
Comments by Copilot are reasonable, especially about missing tests.

dstep/translator/MacroDefinition.d
davispuh: Yeah it could be used I just didn't think of it.

davispuh:
I rewrote this PR and added tests. I also implemented support for decimals (eg. `1.dd`).
I looked into supporting complex numbers (eg. `1.0i`, `2.0J`) aswell but that would add additional *complex* code so dropped it.

davispuh:
Found bug, updated PR with fix.

PR #300, opened 2026-03-18
Fix `extern (C)` header case

davispuh:
Currently if you have `a.h` like:
```c
/**
 * comment
 *
**/

#pragma once

#define A 1
```
And you run `dstep --package a --global-import b a.h`
Then result will be broken code (note split `extern`):
```d
/**
 * comment
 *
**/

extmodule a.a;

import b;

ern (C):

enum A = 1;
```

This PR fixes that so it will be correct code:
```d
/**
 * comment
 *
**/

module a.a;

import b;



extern (C):

enum A = 1;
```

### DIFF
--- a/dstep/translator/IncludeHandler.d
+++ b/dstep/translator/IncludeHandler.d
@@ -197,8 +197,6 @@ class IncludeHandler
 
         if (options.keepUntranslatable)
             importsBlock(output, unhandled.keys);
-
-        output.finalize();
     }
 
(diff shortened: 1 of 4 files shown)


jacob-carlborg (changes requested):
There's no test case.

davispuh:
Rewrote this PR and added tests. First commit fixes original issue. 2nd is additional improvement.

PR #301, opened 2026-03-19
Fix module name with reserved keywords (eg. `in`)

davispuh:
This PR fixes module name when it's using reserved keyword.
For example consider empty `in.h` file that is being converted
```
$ dstep --package a in.h
```
Without this PR that would produce:
```d
module a.in;

```
Which doesn't compile:
```
in.d(1): Error: identifier expected following `package`
module a.in;
         ^
in.d(1): Error: declaration expected, not `in`
module a.in;
```

With this PR it will be:
```
module a.in_;

```
Which does compile.


jacob-carlborg:
I haven't looked carefully but the comments by Copilot seems reasonable.

davispuh:
I updated PR by incorporating Copilot suggestions.

PR #302, opened 2026-03-19
Implement partial support for #define objects

davispuh:
Consider common C pattern like:
```c

struct S
{
  int a;
  int b;
};

#define S_INIT { 1, 2 }

int main()
{
   struct S s = S_INIT;
}
```

Currently `dstep` will ignore `S_INIT` which loses this "init" information.
This PR implements partial support so that
```
enum S_INIT = null; // FIXME: {1,2};
```
Will be produced, which allows to create simple shell script with simple `sed` command to afterwards fix up these case with something like this:
```sh
$ sed -iE 's|null; // FIXME: {\([^}]*\)}|S(\1)|g' file.d
```
Which will make final version to be:
```
enum S_INIT = S(1,2);
```
And that will work correctly when `S_INIT` is used.
Unfortunately this 2nd step with `sed`  or another processing tool is needed because implementing usage-aware processing inside `dstep` is not trivial and would require significant effort to also parse macro usages so that's why I'm settling on this partial support.

### DIFF
--- a/dstep/translator/MacroDefinition.d
+++ b/dstep/translator/MacroDefinition.d
@@ -188,6 +188,11 @@ string translate(CallExpr expression, ExpressionContext context)
         expression.args.map!fmap.join(", "));
 }
 
+string translate(BracedExpr expr, ExpressionContext context)
+{
+    return "null /* FIXME: " ~ expr.expr ~ " */";
+}
+
 string translate(Expression expression, ExpressionContext context)
 {
     import std.format : format;
@@ -350,6 +355,10 @@ string translate(Expression expression, ExpressionContext context)
                 condExpr.expr.translate(context),
                 condExpr.left.translate(context),
                 condExpr.right.translate(context));
+        },
+        delegate string(BracedExpr expr)
+        {
+            return expr.translate(context);
         });
 }
 
--- a/dstep/translator/MacroParser.d
+++ b/dstep/translator/MacroParser.d
@@ -503,6 +503,23 @@ class CondExpr
     }
 }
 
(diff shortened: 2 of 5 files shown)


jacob-carlborg:
Is it possible to determine that it's an initializer and have DStep do the translation?

davispuh:
> Is it possible to determine that it's an initializer and have DStep do the translation?

Unfortunately not really possible because:
1. At initializer `#define` time it's impossible to know what exactly it is and to which type it would map to. That can only be known at usage place.
1. Currently DStep parses `.h` one by one but it could be used in some later file or even not included as part of given input files.
2. Usually initializers are used in `.c` files so you might not even see any usage case.  Usage might be only described in documentation.

Eg. consider `.h` file:
```c
struct s1 {
	int *p1;
	int *p2;
};

struct s2 {
	struct s1 a;
	int	b;
	int	c;
};

#define INITIALIZER(x, y) { {{NULL}, {NULL}}, x, y }
// Use this like struct s2 d = INITIALIZER(1, 2);
```
(none of files have actual usage so you can't know what type it maps to)
Only if you know how it's supposed to be used then you know that it maps to
```d
extern (D) auto INITIALIZER(T1, T2)(auto ref T1 x, auto ref T2 y)
{
    return s2(s1.init, x, y);
}
```
You have to know that it's `s2` and `s1`.
The only way how I think this could be automated is if dstep accepts mapping file like:
```yaml
INITIALIZER: s2
```
So you would create this mapping file manually and pass it to dstep so it would know to map it to this type.

Also FYI I improved this PR and added tests.

PR #304, opened 2026-03-19
Comment out compiler builtins

davispuh:
Consider C header like:
```c
#define va_copy __builtin_va_copy
#define ALIGNOF(TYPE)  _Alignof (TYPE)
#define __unused  __attribute__((__unused__))
#define __noreturn  __attribute__((__noreturn__))
#define STATIC_ASSERT  _Static_assert
#define CONST  const
#define STATIC  static
#define __volatile  volatile
```

Currently `dsep` would produce:
```d
enum va_copy = __builtin_va_copy;
```
Which fails to compile:
```
Error: undefined identifier `__builtin_va_copy`
enum va_copy = __builtin_va_copy;
```

When this PR is applied together with #303
Then result will be:
```d
// FIXME: enum va_copy = __builtin_va_copy;
// FIXME: alias ALIGNOF = _Alignof;
// FIXME: enum __unused = __attribute__(__unused__);
// FIXME: enum __noreturn = __attribute__(__noreturn__);
// FIXME: enum STATIC_ASSERT = _Static_assert;
// FIXME: enum CONST = const;
// FIXME: enum STATIC = static;
```

### DIFF
--- a/dstep/translator/MacroDefinition.d
+++ b/dstep/translator/MacroDefinition.d
@@ -522,6 +522,7 @@ bool translateFunctAlias(
 {
     import std.algorithm.comparison : equal;
     import std.algorithm.iteration : map;
+    import std.algorithm.searching : startsWith;
 
     CallExpr* expr = definition.expr.peek!CallExpr;
     auto expressionContext = ExpressionContext.make(context);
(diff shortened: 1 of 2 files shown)


jacob-carlborg (changes requested):
There's no test added.

davispuh:
I added a test.

PR #305, opened 2026-03-19
Implement partial include support

davispuh:
Currently `#include` headers are dropped and not converted to D import statements.
This PR implements partial support so that include headers will be converted from:
```
#include <b.h>
```
to:
```d
// FIXME: import b;
```

This allows to later process and fix them (for example with `sed`).
This is important when headers contain complicated inter-dependencies and without this information it's difficult to know which other files would need to be imported.

### DIFF
--- a/dstep/Configuration.d
+++ b/dstep/Configuration.d
@@ -113,6 +113,9 @@ struct Configuration
     @("public-global-import", "Add <import> as a public global import.")
     string[] publicGlobalImports;
 
+    /// needed for converting absolute includes to relative
+    string[] includePaths = ["/usr/local/include", "/usr/include"];
+
     Options toOptions(string inputFile, string outputFile) const
     {
         Options options = toOptions();
@@ -150,6 +153,7 @@ struct Configuration
         options.globalAttributes = globalAttributes;
         options.globalImports = globalImports;
         options.publicGlobalImports = publicGlobalImports;
+        options.includePaths = includePaths;
 
         return options;
     }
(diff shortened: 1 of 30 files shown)


jacob-carlborg:
Unfortunately there's no one-to-one mapping of header files and D modules, that's why handling of includes is incomplete. I'm not sure this is the right approach. I've been thinking about supporting generating bindings for a whole directory. Then it would be possible to look up where a declaration is located and add imports based on that instead of the includes. 

There's also no new test case added.

davispuh:
> Unfortunately there's no one-to-one mapping of header files and D modules, that's why handling of includes is incomplete. I'm not sure this is the right approach. I've been thinking about supporting generating bindings for a whole directory. Then it would be possible to look up where a declaration is located and add imports based on that instead of the includes.

I looked into that and sadly it's not really possible to do it that way as show in #309. And even then how would you handle includes that are outside of specified dir so still would need something like this.
I think proper support can be added with [dfix](https://github.com/dlang-community/dfix). That is after all `.d` files have been created, point `dfix` to that directory to automatically cleanup imports => remove unused ones and add missing ones by searching across files. Currently it doesn't have such feature but to me that seems like more realistic solution.

Anyway I rewrote this PR and added tests.
Note I included da7bdb9dccda1ac08eb8c1f4f8ee6cfe45833c93 commit from #297 here aswell.

davispuh:
Found bug. Updated PR with fix and a test.

PR #306, opened 2026-03-19
Implement support for bitfields wider than 64 bits

davispuh:
C allows to have structs like this:
```c
#include <stdint.h>

typedef union {
  struct {
    uint32_t    a  : 16;
    uint32_t    b  : 16;
    uint32_t    c  : 8;
    uint32_t    d  : 8;
    uint32_t    e  : 16;
    uint32_t    f  : 32;
    uint32_t    g  : 32;
  } Bits;
  struct {
    uint64_t    Uint64;
    uint64_t    Uint64_1;
  } Uint128;
} S;
```
Currently it's converted like this:
```d
union S
{
    struct _Anonymous_0
    {
        import std.bitmanip : bitfields;

        mixin(bitfields!(
            uint, "a", 16,
            uint, "b", 16,
            uint, "c", 8,
            uint, "d", 8,
            uint, "e", 16,
            uint, "f", 32,
            uint, "g", 32));
    }

    _Anonymous_0 Bits;

    struct _Anonymous_1
    {
        ulong Uint64;
        ulong Uint64_1;
    }

    _Anonymous_1 Uint128;
}
```

But unfortunately D doesn't support bitfields wider than 64-bits:
```
/usr/include/dlang/ldc/std/bitmanip.d(92): Error: shift by 64 is outside the range `0..63`
        enum ulong maskAllElse = ((~0uL) >> (64 - len)) << offset;
                                                        ^
/usr/include/dlang/ldc/std/bitmanip.d(180): Error: template instance `std.bitmanip.createAccessors!("_a_b_c_d_e_f_g_bf", uint, "f", 32LU, 64LU)` error instantiating
            = createAccessors!(store, Ts[0], Ts[1], Ts[2], offset)
```

This PR fixes this issue by splitting such cases. Producing D code:
```d
union S
{
    struct _Anonymous_0
    {
        import std.bitmanip : bitfields;

        mixin(bitfields!(
            uint, "a", 16,
            uint, "b", 16,
            uint, "c", 8,
            uint, "d", 8,
            uint, "e", 16));

        mixin(bitfields!(
            uint, "f", 32,
            uint, "g", 32));
    }

    _Anonymous_0 Bits;

    struct _Anonymous_1
    {
        ulong Uint64;
        ulong Uint64_1;
    }

    _Anonymous_1 Uint128;
}
```

### DIFF
--- a/dstep/translator/Record.d
+++ b/dstep/translator/Record.d
@@ -49,11 +49,19 @@ void translateBitFields(
         auto type = translateType(context, cursor).makeString();
         auto name = cursor.spelling();
         auto width = cursor.bitFieldWidth();
+        auto typeWidth = cast(uint)(cursor.type.sizeOf * 8);
 
-        if (name.empty)
-            output.singleLine("%s : %d;", type, width);
+        if (typeWidth > 64)
+        {
+            output.singleLine("%s %s; // only %s bits used, struct's layout incompatible with C", type, name, width);
+        }
         else
-            output.singleLine("%s %s : %d;", type, name, width);
+        {
+            if (name.empty)
+                output.singleLine("%s : %d;", type, width);
+            else
+                output.singleLine("%s %s : %d;", type, name, width);
+        }
     }
 }
 
--- a/dstep/translator/Record.d
+++ b/dstep/translator/Record.d
@@ -35,62 +35,6 @@ void translatePackedAttribute(Output output, Context context, Cursor cursor)
         output.singleLine("align (1):");
 }
 
-struct BitField
-{
-    string type;
-    string field;
-    uint width;
-}
-
-BitField[] translateBitFields(
-    Context context,
-    Cursor[] cursors)
-{
-    static void pad(ref BitField[] bitFields, uint totalWidth)
-    {
-        uint padding = 0;
-
-        for (uint size = 8; size <= 64; size *= 2)
-        {
-            if (totalWidth <= size)
-            {
-                padding = size - totalWidth;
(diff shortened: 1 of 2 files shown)


jacob-carlborg:
Perhaps time to move to the built in syntax for bit fields.

jacob-carlborg (changes requested):
There's no test case.

davispuh:
Okay so I completely rewrote this PR like 4 times till ended up at this version because I kept finding issues.
Turns out to get struct layout that would match C using `std.bitmanip.bitfields` is very complex and essentially yeah it's just not worth it so now just use native bitfields.
Only there is case for `__int128` which supporting would also be incredibly complex because we use `Cent` which can't be bitpacked like in C. It is actually possible to implement with unions and sub-structs but it's like 1k code lines of complex logic just for that so I dropped it.

davispuh:
In case someone ever wants to support 128-bitfields here are some edge case examples:
```c

struct Foo {
    int a : 30;
    long b : 34;
    __int128 c : 62;
};

struct Foo {
    __int128 a : 100;
    int b : 20;
};

struct Foo {
    int a : 30;
    long b : 35;
    __int128 c : 87;
    int d : 21;
};

struct Foo {
    __int128 a : 50;
    __int128 b : 50;
};

struct Foo {
    __int128 a : 100;
    int b : 20;
    short c : 4;
};

struct Foo {
    __int128 a : 50;
    int : 14;
    int b : 10;
    __int128 c : 62;
};
```

PR #307, opened 2026-03-19
Support `&=`, `^=`, `|=` for macros

davispuh:
Consider macro like this:
```c
#define FLAG_CLEAR(S, Flag) (S->Flags &= ~(Flag))
```
Currently such macros with `&=`, `^=` and `|=` are dropped.
So this PR implements support for them and will produce D code like:
```d
extern (D) auto FLAG_CLEAR(T0, T1)(auto ref T0 S, auto ref T1 Flag)
{
    return S.Flags &= ~Flag;
}
```


jacob-carlborg (changes requested):
There's no test added.

davispuh:
I fixed operator precedence and added tests.

PR #310, opened 2026-06-30
Fix wide strings with `L` prefix

davispuh:
If you have C code that uses wide strings like
```c
#define H L"hello world"
```
Then currently `dstep` doesn't convert it causing wrong D code to be created.

This PR implements converting such code to valid D.


jacob-carlborg:
Thanks.

PR #311, opened 2026-06-30
Fix <unimplemented> for size_t return type on builtins

davispuh:
Currently if you have code like this:
```c
#include  <stdio.h>

size_t strlen(const char *__s);
size_t fread();
```

It's translated to:
```d
extern (C):

<unimplemented> strlen (const(char)* __s);
<unimplemented> fread ();
```
Which obviously is invalid code.
This PR fixes this to create correct D code.

Note that this issue happens only with compiler's builtin functions that return `size_t` because clang merges it to internal unexposed type.

### DIFF
--- a/dstep/translator/Type.d
+++ b/dstep/translator/Type.d
@@ -300,8 +300,11 @@ do
     if (declaration.isValid)
         return translateType(context, declaration, rewriteIdToObjcObject);
     else
+    {
+        type = type.canonical.isValid ? type.canonical : type;
         return translateType(context, type.kind, rewriteIdToObjcObject)
             .makeSourceNode();
+    }
 }
 
 string translateComplex (Type type)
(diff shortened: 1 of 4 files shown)


jacob-carlborg:
@davispuh I apologize for the delay. I fixed the CI pipeline in master and rebased all your pull requests on top of latest master so the CI pipelines can be run.

davispuh:
Great! But looks like `x86_64-pc-windows-msvc dmd-beta` still fails with:
```
+ ./configure --llvm-path llvm-21.1.8 --statically-link-clang
Failed to invoke the compiler dmd to determine the build platform: '"dmd"' is not recognized as an internal or external command,

operable program or batch file.
```

Other failures are due Windows differences so actual PR issues will need to think how to make it work.

jacob-carlborg:
> Great! But looks like x86_64-pc-windows-msvc dmd-beta still fails with:

Yes, that's expected, but it doesn't block the whole pipeline, see: https://github.com/jacob-carlborg/dstep/actions/runs/30194741518.

